PR #14, opened 2016-07-27
Started NGS-QC section for SOP generation

jfallmann:
Let's get started


bgruening:
This is our first approach to collect some kind of SOPs/RFC/Best-practices/ we would like to teach.

PR #19, opened 2016-08-19
Add a template for slides

bebatut:
Hi, 

I add a template for slides in `reveal.js` inside the `shared` directory, with the shared `css` and `js` scripts.

Bérénice


bgruening:
This is getting so beautiful!

PR #23, opened 2016-08-19
Add initial travis testing

bgruening:
We will start with testing links automatically.

If anyone wants to contribute to this project this should be an easy way to get involved.


bgruening:
@bebatut what do you think about this. If you agree we should get this or something similar in before the hackathon.

bebatut:
Yes. I could work on this in next days

PR #25, opened 2016-08-29
Add commands to automatically build index.html from README.md

bebatut:


bgruening:
Nice one. I think the index.html needs to be pushed back to master without triggering a new travis run.

bgruening:
Let's see how this works. It looks very promising.

bebatut:
It did not change the `index.html` file :( :( 
I will continue to investigate....

PR #110, opened 2016-10-06
Add RNASeq de novo assembly

abretaud:
Some new material for RNASeq de novo assembly.
Currently there is only a dockerfile based on [this new docker flavor](https://github.com/abretaud/docker-galaxy-rnaseq-de-novo) (still building)

### DIFF
--- a/RNA-Seq-de-novo/README.md
+++ b/RNA-Seq-de-novo/README.md
@@ -1,0 +1,50 @@
+RNA-Seq de novo assembly data analysis
+======================================
+
+RNA-sequencing is a method used to reveal the presence and quantity of RNA in a biological sample at a given moment in time.
+When no reference genome is available for the studed species, it is possible to perform an assembly of RNA-Seq data to produce a transcriptome.
+Here, we propose some materials to learn how to analyze RNA-seq data using a de novo assembly strategy.
+
+# Slides
+
+Several deck of slides will be available for this topic...
+
+# Tutorials
+
+A tutorial with hands-on will be available for this topic:
+
+- [RNA-seq de novo assembly](tutorials/de_novo.md)
+
+## Input datasets
+
+The input datasets for the tutorials will be soon available on Zenodo.
+
+## Galaxy instance
+
+For these tutorials, you can use the [dedicated Docker image](docker/README.md):
+
+```
+docker run -d -p 8080:80 bgruening/galaxy-rna-seq-assembly-training
+```
+
+It will launch a flavored Galaxy instance available on
+[http://localhost:8080](http://localhost:8080).
+
+# References
+
+## Papers
+
+## Websites
+
+# Contributors
+
+This material is maintained by:
+
+- Anthony Bretaudeau (@abretaud)
+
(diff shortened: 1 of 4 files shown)


bebatut:
Hi,

Thanks.

It would be better to add this material to the `RNA-Seq` directory.
Do you plan to add a tutorial (markdown and slides)?

Bérénice

lecorguille:
@abretaud A thread about RNASeq denovo path: https://github.com/bgruening/training-material/issues/94

abretaud:
Right, for now I only have the docker stuff, so do you think we should have a single docker for both ref-based and de novo rnaseq? Or keep the 2 separated (but how to store it in RNA-Seq?)
I don't have yet a tutorial, I can work on it, but later :smile:

bebatut:
From my point of view, I would like to have only one docker. But maybe it is not the best solution. What do you think?

abretaud:
Hehe, I hesitate too, that's why I ask :laughing: 
I'll see a bit later if I can switch to the single docker solution.

bgruening:
A single container would lower the maintenance effort and would enable to make a comprehensive workshop about all RNAseq things with one instance.

The only reason I see against one single container is that it get to big at some point ... I'm not so worried about this currently :)

RNA-Seq-de-novo/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+This material is maintained by:
+
+- Anthony Bretaudeau (@abretaud)
+
+For any question related to this topic and the content, you can contact them.
+
+The following individuals have contributed to this training material:
+
+- Name 1
```
bgruening: Your name :)

abretaud:
Yep, and people will have some tools in the left that they won't use depending but it's probably not a problem.
And it's probably better to present the difference between ref-based and de novo in the slides/tutorial.
I'll rework this PR in a few days, I think I won't be able to do much more during this hackathon (sorry for that)

bgruening:
@abretaud no problem. Thanks for your contribution!

RNA-Seq-de-novo/docker/README.md
```diff
@@ -0,0 +1,29 @@
...
+[![Docker Stars](https://img.shields.io/docker/stars/bgruening/galaxy-training-exome-seq.svg?maxAge=2592000)](https://hub.docker.com/r/bgruening/galaxy-training-exome-seq/)
+
+Galaxy Docker repository for Exome sequencing training material :whale: :eyes:
+====
+
+:whale: This Docker Image contains everything what is needed to run this training, including an interactive Galaxy Tour.
+
+# Requirements
+
+- [Docker](https://docs.docker.com/installation/) for Linux / Windows / OSX
```
yhoogstrate: Are there any hardware requirements, e.g. RAM?

bgruening: Depending on the tools and the training data ... so hard to tell. In general no. Every special requirement should be stated in the readme file imho.

abretaud:
Closing this PR, nothing valuable here, we'll rework this later

PR #124, opened 2016-10-07
Makefile to convert slides to PDF

blankclemens:


willdurand:
Fixed by #448?

blankclemens:
Indeed, thanks @willdurand!

PR #133, opened 2016-10-07
Would be possible you include the IFB's logo here

vdda:
logo-ifb.png


bgruening:
Awsome. Will fix https://github.com/bgruening/training-material/issues/121

PR #138, opened 2016-10-11
fix links

shiltemann:
the travis test to validate URLs should now pass 

this test does not test relative URLs, so should we encourage absolute URLs after all so that they are tested? (will be more to change when we move repo but nice to test all links)


shiltemann:
maybe not ;P

ah, it's taking a long time in some reveal.js file, but adding `-not -path "./shared/reveal.js/*"` to find command should remedy that

also, FYI, if you run these in quick succession you start getting 429 errors for too many connection attempts ..but in practice probably not big problem

shiltemann:
lol, and of course broadinstitute website just went down ;P

bgruening:
Adding paths to blacklist folders is the way to go! Thanks!
Is there no way to follow symlinks in URLs? I think in this case it's ok to skip these for now, maybe I find a better library to test this stuff.

Thanks.
I merged the travis PR can you rebase this one here it looks like many things are included.

shiltemann:
sure, will close this and PR to master then :)

PR #139, opened 2016-10-11
fix links and update travis test

shiltemann:


shiltemann:
hmm, why did test pass when there are failed urls? ..does it only take exit code of last checked file?

bgruening:
I guess it is this one: http://stackoverflow.com/questions/14871147/how-to-exit-from-find-exec-if-it-fails-on-one-of-the-files

shiltemann:
ok, changed the travis command, it should now fail if any of the files fail, but still always run the check on all of them :) ..and it's also checking html files now

bgruening:
Awesome!

PR #140, opened 2016-10-12
fixed tense and plural/singular grammar issues

tealiie:
fixed some singular/plural grammar issues
changed words to match the tense used
statements read as questions were rewritten as questions


bebatut:
Thanks for this :)

PR #142, opened 2016-10-17
Fix container name

manabuishii:
I can't pull

```
bgruening/galaxy-exome-seq-training
```

but there is

```
bgruening/galaxy-training-exome-seq
```


bgruening:
Thanks!

Issue #94, opened 2016-10-05
Fill the hands-on tutorial and the slides for Ref based RNA Seq tutorial and add a Galaxy tour


bebatut:
Hi, 

The [tutorial](https://github.com/bgruening/training-material/blob/master/RNA-Seq/tutorials/ref_based.md) and the [associated slides](https://github.com/bgruening/training-material/blob/master/RNA-Seq/tutorials/ref_based.html) on Ref based RNA-seq needs to be filled.

I thought to combine the two previous tutorials:
- [The one used in Freiburg WS](https://github.com/bgruening/training-material/blob/master/RNA-Seq/tmp/RNA-Seq-hands-on.md)
- [Reference-based one](https://github.com/bgruening/training-material/blob/master/RNA-Seq/tmp/Reference-based-RNA-seq-detailed.md)

For the slide deck, we can take inspiration from [slides used in Freiburg WS](https://github.com/bgruening/training-material/blob/master/RNA-Seq/tmp/RNA-Seq-hands-on.md) and also from the first part of [Reference-based tutorial](https://github.com/bgruening/training-material/blob/master/RNA-Seq/tmp/Reference-based-RNA-seq-detailed.md)

We also need to add a Galaxy tour

Thanks,

Bérénice

lecorguille:
What do you suggest for RNASeq denovo for non-model organisms?
We should propose material today for this topic.

For sure, there are sharing the DE analysis but the upstream (assembly) and downstream (annotation) steps are not common.

Must we create a `/master/RNA-Seq-denovo` or a subdirectory `/master/RNA-Seq/RNA-Seq-denovo` ?

bebatut:
For RNASeq de novo, you can add new tutorial *i.e.*:
- A markdown file in `RNA-seq/tutorials/de_novo_rna_seq.md` following the template
- Associated slides in `RNA-seq/slides/de_novo_rna_seq.html` following the template

We tried to detail what you wanted in the [`CONTRIBUTING.md`](https://github.com/bgruening/training-material/blob/master/CONTRIBUTING.md)

Is it clear?

bebatut:
With @blankclemens, we will work on this

bebatut:
We have now the tutorial: http://galaxyproject.github.io/training-material/topics/transcriptomics/tutorials/ref-based/tutorial.html

PR #147, opened 2016-10-20
Formatting the reference based RNA-Seq tutorial

bebatut:
Hi,

As mentioned in #94, I started to reformat the tutorial on reference based RNA-Seq.
I also added more content (questions, ...).

Bérénice


bebatut:
I got some review. I will merge it to integrate it correctly in the new templates. 
We can still discuss about the content here afterwards and I will keep it up to date.

PR #170, opened 2017-01-04
Dev coner new template

lecorguille:
Hop!

I've just apply the new template remark to the Tool Integration slides

I have to propose some modifications within the css:
- Possibility to resize picture 10%, 25%, 50%, 75%: `.image-50` 
- Possibility to use the css class .my-footer outsite div. I can remove that if you think that it putthe mess somewhere else


bgruening:
The failing test was a false positive. Thanks @lecorguille!

PR #174, opened 2017-01-06
Various corrections to dev-corner

abretaud:
I have not finished reviewing, but here's a PR I will update...

### DIFF
--- a/Dev-Corner/slides/tool_integration.html
+++ b/Dev-Corner/slides/tool_integration.html
@@ -490,15 +490,15 @@ $ planemo tool_init --id 'some_short_id' --name 'My super tool'
 Complicated version:
 
 ```bash
-$ planemo tool_init --id 'seqtk_seq' \
-                    --name 'Convert to FASTA (seqtk)' \
-                    --requirement seqtk@1.0-r68 \
-                    --example_command 'seqtk seq -a 2.fastq > 2.fasta' \
-                    --example_input 2.fastq \
-                    --example_output 2.fasta \
-                    --test_case \
-                    --cite_url 'https://github.com/lh3/seqtk' \
-                    --help_from_command 'seqtk seq'
+$ planemo tool_init  --id "samtools_sort" --name "Samtools sort" \
+                     --description "Sort alignments by leftmost coordinates, or by read nam" \
+                     --requirement samtools \
+                     --example_command "samtools sort -o 1.bam null > 1_sorted.bam" \
+                     --example_input 1.bam \
+                     --example_output 1_sorted.bam \
+                     --test_case --version_command "samtools --version | head -1" \
+                     --help_from_command "samtools sort" \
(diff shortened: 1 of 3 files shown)


abretaud:
@lecorguille do you have the sources of the image to fix the mistake in https://github.com/bgruening/training-material/blob/master/Dev-Corner/images/macro.png ?

lecorguille:
The same!

@abretaud It's a pptx ... I can translate it in yED

abretaud:
Or you can play with Paint :)
I think this is ready for review now, can you have a look?
I thought we had some more slides about planemo and its command (tool_init & co), but can't find them... Where they lost in translation from old templates, or is it once again one of my friday hallucinations?

lecorguille:
@abretaud I had some light slides on planemo tool_init in my pptx. I like to explain that they must Read The Funcking Manual. 

But indeed, we can add a hands-on:
```bash
$ planemo tool_init --help
$ planemo serve --help
...
```
 to browse with them the doc and target important paramters
?

abretaud:
I had in mind small descriptions of the main commands, like:
- with tool_init you can create a minimal xml, or even write it completely by passing the right arguments
- with serve you get a little galaxy instance available at http://localhost...
- ...

I fear looking at the whole --help would be a little... boring :D

lecorguille:
As you wish! A little introduction as you propose is ok for me.
About looking at the help, it was just to spot some parameters. But indeed, for `init_tool`and `serve`, there isn't plenty of arguments.

So, as you propose!

abretaud:
Ok, just added a little intro on main planemo commands, what do you think?

Dev-Corner/slides/tool_integration.html
```diff
@@ -477,6 +477,83 @@
...
+```bash
+$ mkdir new_tool
+$ cd new_tool
+$ planemo tool_init --id 'some_short_id' --name 'My super tool'
+```
+
+Complicated version:
+
+```bash
+$ planemo tool_init --id 'seqtk_seq' \
```
lecorguille: Can you use an other example?
This command solves the half of the exercice of this course 🔨 

What do you think about:
```bash
planemo tool_init \
 --id "samtools_sort" --name "Samtools sort" \
 --description "Sort alignments by leftmost coordinates, or by read nam" \
 --requirement samtools \
 --example_command "samtools sort -o 1.bam null > 1_sorted.bam" \
 --example_input 1.bam \
 --example_output 1_sorted.bam \
 --test_case --version_command "samtools --version | head -1" \
 --help_from_command "samtools sort" \
 --doi 10.1093/bioinformatics/btp352
```

abretaud: hehe LAUL
fixed :)

lecorguille: Merci

PR #178, opened 2017-01-13
Visualization tutorial formatting for the new templates

bebatut:
Hi, 

I formatted the new visualization tutorial to fit the new templates.

@shiltemann, @yhoogstrate Can you check that I did not alter the content, please? Is the `file` directory really needed?

Bérénice


shiltemann:
looks good  @bebatut, thank you! :)

we added the folder in the `files` directory because it contains all the code, with the correct directory structure, to reproduce what is in the tutorial, simply by copying this folder to your Galaxy, we thought it might be handy for student who run into problems? :) 

one note (not for now but in future): it would be nice if the syntax highlighting worked in your template/on the website, like it does when you view it directly in github :)

bebatut:
Maybe we can move the files directory at the the root of dev-corner or add them as a Zenodo dataset. We can maybe open an issue about that to discuss how to collect the expected results and intermediate files needed to help the students. 

For the future question, can you open an issue to explain your idea? As that you can keep a track, discuss and trying to fix it.

Thanks

shiltemann:
sure, fine with moving it wherever you think makes the most sense :) ..the other two directories in the `files` folder are not ours, so we should check to see if there are any links to those files in any materials that need to be updated before we move them (we would also need to adjust a link in our tutorial) :)

sure, will open an issue for that :)

thanks again, and see you next week :)

bgruening:
I would vote for keeping code here, this is easier to update. Code github, data zenodo :)

shiltemann:
agreed :) ..also, probably a good idea to start using `git-lfs` for all the images etc? :)

bgruening:
If this repository keeps increasing like this, we probably need lfs at some point. Currently, I'm afraid of this extra barrier.

PR #186, opened 2017-01-16
Update Docker slides

bebatut:
Hi,

I updated Docker slides to fit them to the new templates. I added also new content to introduce Docker and the commands

Bérénice


bgruening:
Do we really need the pngs if we have the SVGs?

PR #190, opened 2017-01-17
small updates to visualizations tutorial

shiltemann:
update some text to fit with new input dataset we're using and remove unrelated objectives from metadata


bgruening:
Thanks!

PR #210, opened 2017-02-05
Dev-Corner Visualizations updates

shiltemann:
I split the visualization dev slides into two (one for viz, one for interactive environments) and made some small adjustments to the latter


metadata/Dev-Corner.yml
```diff
@@ -121,6 +121,35 @@ material:
     - "Help yourself by using the available web plugins and record an interactive tour"
 
 -
+    title: "Galaxy Interactive Environments"
+    type: "tutorial"
+    name: "interactive_environments"
+    zenodo_link: ""
+    galaxy_tour: ""
+    hands_on: "yes"
```
bebatut: Do you plan to add an hands-on specific to the IE? Because here, a link to an hands-on will be generated...

shiltemann:
I would like to, but may be a little while before I have the time, so changed it to no for now :) thanks!

bebatut:
Thanks !

PR #213, opened 2017-02-09
Update Genome-Annotation yml

AnnaSyme:
added in a new section on microbial genome annotation, using the tool Prokka


bgruening:
:) thanks

PR #217, opened 2017-02-10
Update Ref based RNA seq tutorial

bebatut:
Hi,

I changed the ref based RNA seq to add a new toy dataset (for the first part) and make it more coherent. I also changed the mapping tool used to HISAT.

Bérénice


bgruening:
@MoHeydarian can you maybe review this?

MoHeydarian:
These additions look very nice. Thanks, Berenice!

PR #220, opened 2017-02-11
Add former Galaxy Introduction: From peaks to gene

blankclemens:


bgruening:
@blankclemens thanks! Great step forward. I will merge this now as it is clearly better then the paceholder and we probably need it on Monday ;)

ping @bebatut @joachimwolff for an additional post-merge review

PR #221, opened 2017-02-12
Fixes and enhancements to tool_integration slides

nsoranzo:


Dev-Corner/slides/tool_integration.html
```diff
@@ -442,7 +444,7 @@
 graphlan.py
 --format '$format'
 &&
-echo "Yeeha it worked"
+echo "Yeah it worked!"
```
bgruening: https://en.wiktionary.org/wiki/yeeha ;)

nsoranzo: We cowboy now? :stuck_out_tongue_winking_eye:

bgruening: Look at the presidents of the world ... I guess we are. Sad story.
Thanks Nicola!

PR #228, opened 2017-02-13
Update ref_based.md

pavanvidem:
--> extracted reads are only from chr4
--> There was an extra </ol> after the question "How are the samples grouped?" which changed the whole questions formatting in the rest of the document.
--> I think, ENSEMBL_GENE_ID is more intuitive than FLYBASE_GENE_ID because the annotation was from Ensembl.


bebatut:
Thanks!

PR #233, opened 2017-02-13
Fix some slide headers in QC

bebatut:


NGS-QC/slides/index.html
```diff
@@ -72,7 +72,8 @@
 
 ---
```
bgruening: B! is it correct that this file is called html?

bebatut: Yes :)

Issue #197, opened 2017-01-20
De novo transcriptome reconstruction from RNA-seq data (WIP)


malloryfreeberg:
Hi all,

@MoHeydarian and I are working on a tutorial for de novo transcriptome reconstruction (reference gene annotation-free) from RNA-seq data. These data are from the same study as the ChIP-seq tutorial we prepared, so we are hoping in the future to have another tutorial on how to integrate the two analyses. We would like to note that this tutorial differs from the proposed #59 Trinity tutorial, so we feel both tutorials would be good to have.

@bebatut or @bgruening: would it be best to nest this tutorial under the RNA-seq tutorial repo?

bgruening:
@malloryfreeberg I think nesting is completely fine, for example we have multiple introduciton: https://github.com/galaxyproject/training-material/tree/master/Introduction/slides

I would even think that complete sub-repos would make sense to have a tour and a Docker container for every single tutorial.

malloryfreeberg:
Sounds great @bgruening!

PR #241, opened 2017-02-14
New RNA-seq tutorial: de novo transcriptome reconstruction

malloryfreeberg:
Mo and I have put together a tutorial for de novo reconstruction of a transcriptome from RNA-seq data with a reference genome for alignment. The data used in this tutorial are from the same study we used for the ChIP-seq tutorial, so the tutorials are complementary. We are still working on slides and a Dockerfile, but the major tutorial items are ready. Please let us know if we're missing anything from the tutorial or metadata files.


malloryfreeberg:
This satisfies issue #197 so feel free to close that issue if this gets merged.

metadata/RNA-Seq.yml
```diff
@@ -56,6 +56,42 @@ material:
...
+    - "Third key point"
+    - "..."
+
+    requirements:
+    -
+        title: "Mapping"
+        type: "internal"
+        link: "/NGS-mapping"
+    -
+        title: "IGV"
```
bgruening: Trackster is a requirement as well, right? I have the feeling that not many people have a fully functional trackster these days. So maybe we should mention this.

metadata/RNA-Seq.yml
```diff
@@ -56,6 +56,42 @@ material:
...
+    questions:
+    - "What genes are differentially expressed between G1E cells and megakaryocytes?"
+    - "How can we generate a transcriptome de novo from RNA sequencing data?"
+    
+    objectives:
+    - "Analysis of RNA sequencing data using a reference genome"
+    - "Reconstruction of transcripts without reference transcriptome (de novo)"
+    - "Analysis of differentially expressed genes"
+
+    time_estimation: "1d"
```
bgruening: Really a day? :) 6h maybe? Or does a day, really means 24h?

metadata/RNA-Seq.yml
```diff
@@ -56,6 +56,42 @@ material:
...
+
+    requirements:
+    -
+        title: "Mapping"
+        type: "internal"
+        link: "/NGS-mapping"
+    -
+        title: "IGV"
+        type: "internal"
+        link: "/Introduction/IGV"
```
bgruening: Where are you using IGV?

RNA-Seq/tutorials/de_novo.md
```diff
@@ -0,0 +1,377 @@
...
+
+    ![](../images/DESeq2_dispersion1.png)
+
+    This dispersion plot is typical, with the final estimates shrunk from the gene-wise estimates towards the fitted estimates. Some gene-wise estimates are flagged as outliers and not shrunk towards the fitted value. The amount of shrinkage can be more or less than seen here, depending on the sample size, the number of coefficients, the row mean and the variability of the gene-wise estimates.
+
+
+For more information about `DESeq2` and its outputs, you can have a look at [`DESeq2` documentation](https://www.bioconductor.org/packages/release/bioc/manuals/DESeq2/man/DESeq2.pdf).
+
+# Visualization
+Now that we have a list of transcript expression levels and their differential expression levels, it is time to visually inspect our transcript structures and the reads they were predicted from. It is a good practice to visually inspect (and present) loci with transcripts of interest. Fortuantely, there is a built-in genome browser in Galaxy, **Trackster**, that make this task simple (and even fun!).
```
bgruening: Add hint for IGV and a link to the IGV tutorial?

metadata/RNA-Seq.yml
malloryfreeberg: Remnant of copy+paste. I'll update to a more reasonable time :)

metadata/RNA-Seq.yml
malloryfreeberg: This is also a remnant of copy+paste. Will update to Trackster since that is what we actually use in the tutorial.

metadata/RNA-Seq.yml
malloryfreeberg: We aren't. Changing to Trackster.

RNA-Seq/tutorials/de_novo.md
malloryfreeberg: I would argue just to keep everything in relation to Trackster for the tutorials. Is there a reason some tutorials use IGV? Why not UCSC or another browser?

malloryfreeberg:
@bgruening I have double-checked the images and we had replaced 2 of them accidentally for the ref_based tutorial (DESeq2_PCA.png and DESeq2_heatmap.png). I have returned the original images, renamed the images for the de_novo tutorial (DESeq2_PCAplot.png and DESeq2_heatmap_G1E_mega.png), and everything looks good to me. Note that I believe the PCA and heatmap for the ref_based tutorial is no longer embedded in the current ref_bsed.md file, but they are embedded correctly in our older version of ref_based.md.

RNA-Seq/tutorials/de_novo.md
MoHeydarian: I agree with @malloryfreeberg, we should try to keep users inside the Galaxy environment for viewing their data on a genome browser. Moving data to an external genome browser introduces the opportunity to lose track of files and is extremely cumbersome (for larger datasets) IMO. Using Trackster within Galaxy makes viewing data a seamless step (when it works).

You raise a good point @bgruening, Trackster has not always been stable. It has been working more reliably over the past year. It is really a great feature within Galaxy to promote "looking at your data" with a couple of clicks. Maybe we could convince some folks participating in the GCC2017 hackathon to address Trackster issues. I'll be suggesting ways to make Trackster more usable at the GCC2017 datathon (specifically by making reference datasets and repeat tracks available as default).

bgruening: It's just a feeling, but don't know about much Galaxy instances that offer a working trackster experience. IGV can be installed by everyone, very easily. I think we can leave this in, but maybe we should just add a sentence, a link to IGV or others (if we have) to just give pointers.

I totally agree that trackster would be the way to go it is actively maintained.

bgruening:
@malloryfreeberg There are a few broken links to zenodo, can you please check that. Otherwise, I'm fine with merging! Thanks a lot!

bgruening:
Nice! Great work @malloryfreeberg and @MoHeydarian!

PR #243, opened 2017-02-15
Adding new topic and first tutorial for this topic

nekrut:
I added a new topic, which will cover tutorials involving usegalaxy.org. In addition, I have made a number of modification to the site itself:

 - added `_plugins` with a vimeo plugin. It allows embedding vimeo video using `{% vimeo 26521352 %}` shortcode
 - modified `tutorial_hands_on.html` from `_layouts` to render docker and slide menu items (in the navbar) only of they are defined
 - added `topic_docker_image` and `topic_galaxy_instance` variable to metadata *.yaml. See (`Usegalaxy.yml`). This allows defining links to galaxy instance and docker container specific to a given tutorial
- Added `galaxy_instance` variable for topic (see `Usegalaxy.yml`). It is used to render Galaxy instance section within `topic.html` layout.


bgruening:
@nekrut if we have the `galaxy_instance` variable, does it make sense to have it's own usegalaxy.org subfolder or can you use Introduction/ChIP-seq for it?

ping @malloryfreeberg, @MoHeydarian, @bebatut

nekrut:
@bgruening Not sure I understand. But if we will have tags and tag tutorials that run on usegalaxy.org that won't be a problem.

nekrut:
Also, this travis check is very fragile because some links timeout. In the latest the problem was http://onlinelibrary.wiley.com/doi/10.1111/j.1365-294X.2011.05074.x/abstract although now it is perfectly accessible. Not sure what to do about it through. Probably nothing.

_plugins/vimeo.rb
```diff
@@ -0,0 +1,27 @@
+class Vimeo < Liquid::Tag
```
bebatut: Maybe the plugins directory can go into assets. What do you think?

_layouts/tutorial_hands_on.html
```diff
@@ -32,22 +32,30 @@
...
-              <li><a href="{{ site.url }}/{{ topic.name }}/slides"><i class="fa fa-graduation-cap"></i> Introduction slides</a></li>
+              <li><a href="{{ site.url }}" title="Back to front page"><i class="fa fa-home"></i> Home</a></li>
+              <li><a href="{{ site.url }}/{{ topic.name }}" title="Go back to list of tutorials"><i class="fa fa-file-text-o"></i> {{ topic.name }}</a></li>
+
+              {% if tutorial.slides != "" %}
+                <li><a href="{{ site.url }}/{{ topic.name }}/slides" title="Slides for this tutorial"><i class="fa fa-graduation-cap"></i> Introduction slides</a></li>
+              {% endif %}
               <!--<li><a href="{{ site.url }}/{{ topic.name }}/slides/{{ tutorial.name }}/"><i class="fa fa-file-powerpoint-o"></i> Associated slides</a></li>-->
 
-              {% if topic.docker_image != "" %}
```
bebatut: Can you keep the `docker_image` variable for now?

bebatut: Or we can have a nested test there? 
```
{% if tutorial.tutorial_docker_image != "" %}
     <li><a href="{{ site.github_repository }}/tree/master/{{ topic.name }}/docker" title="Docker image for this tutorial"><i class="fa fa-ship"></i> Docker image</a></li>
{% elsif tutorial.tutorial_docker_image != "" %}
     <li><a href="{{ site.github_repository }}/tree/master/{{ topic.name }}/docker"><i class="fa fa-ship"></i> Docker image</a></li>
{% endif %}
```

_plugins/vimeo.rb
nekrut: I don't have particular opinion here

_layouts/tutorial_hands_on.html
nekrut: As @bgruening suggested it probably only makes sense to have topic-level docker images, so I'll rethink that

nekrut:
@bebatut: the idea behind usegalaxy section was to put together tutorial utilizing usegalaxy.org. Are you suggesting simply putting them into corresponding topics (e.g., ChIP-seq) and using `galaxy_instance` instead?

bebatut:
Yes I think it will make more sense for me if we have the `galaxy_instance` variable.

PR #253, opened 2017-02-17
Diploid tutorial fdr usegalaxy section

nekrut:


bgruening:
I would like to see this still under https://github.com/galaxyproject/training-material/tree/master/Exome-Seq and put it together with the other variant calling tutorials.

nekrut:
but Exome-seq is not the only case where diploid variants are called. Perhaps a new section called variation analysis

bgruening:
Sure, we should rename Exome-Seq to the more general Variant Calling.

PR #255, opened 2017-02-17
Change HISAT to HISAT2

nsoranzo:
The correct name is used in the Galaxy wrapper since https://github.com/galaxyproject/tools-iuc/pull/1141 .


bebatut:
Thanks

PR #270, opened 2017-03-29
Added two published GalaxyP workflows

Stortebecker:
Hejhej,

I checked the display of the workflows' Readme.md, but I was not sure, how to link to the workflows in the central Proteomics/Readme.md. Please check, if my current solution suits your needs.

Greetings from Regensburg, Flo

P.S.: Oh, and I just saw that something strange happened with the hisat.png in this PR. I tried to fix it several times, but was not able to. Can you please take care of that?


Proteomics/workflows/ntails/README.md
```diff
@@ -0,0 +1,43 @@
+# GalaxyP Workflow: N-Tails Data Analysis
```
bebatut: Is this document just a description of the workflow? Or do you want to use it as tutorial? In the latter case, it would be better to put it in the tutorials directory, and to follow the requirements for a tutorial. In the first case, can you remove the boxes, please? They are just used for tutorials not for README

Stortebecker:
Hej Berenice.

> Do you plan to have tutorials where each workflow step is done and explained and at the end linking to the full workflow?

No, these are really special techniques, I am not planning to write a full tutorial at this timepoint.

> To what does secretome_prediction/test_data.tabular correspond? We try to avoid data inside the repository

Ah, yes. It is only a list of protein-IDs though, a few bytes all-in-all. I could upload it to zenodo, but it is not available elsewhere online. Do you want me to?

> Do you plan to have a Docker image? It could be great to add the workflow there?

Not at the moment. I would need help to create a docker image. I guess that Björn plans to add a Docker at a later timepoint, please ask him for details.

> In the first case, can you remove the boxes, please?

Do you mean the `> :nut_and_bolt: Comment ` ? If so, I can surely remove them.

Stortebecker:
> In the first case, can you remove the boxes, please?

Would it be sufficient to remove the `>` in the beginning of each line?

Stortebecker:
... and the `{: .comment}` at the end, of course

bgruening:
Hi,

I don't think we should start to add any data to this repo, no matter how big they are.
Regarding the hisat image I would need to fix this on your branch I guess :(

Stortebecker:
@bgruening Shit, can you please fix it? I tried and tried... Even Clemens could not fix it.

and sure, I will upload the data to zenodo.

Stortebecker:
> To what does secretome_prediction/test_data.tabular correspond? We try to avoid data inside the repository

I deleted the file and replaced it by a link to zenodo.

PR #276, opened 2017-04-07
Database handling tutorial update

Stortebecker:
Hejhej,
I updated the database handling tutorial. It should be displayed nicely now, even in the github pages.
Cheers, Flo


bgruening:
Thanks Flo!

PR #277, opened 2017-04-07
Docker - RNA-seq

bgruening:
This PR demonstrates how we can easily pre-populate Docker images with reference indices.

ping @nekrut, @bebatut, @MoHeydarian, @malloryfreeberg

### DIFF
--- a/RNA-Seq/docker/Dockerfile
+++ b/RNA-Seq/docker/Dockerfile
@@ -1,6 +1,6 @@
 # Galaxy - RNA-Seq training material
 
-FROM bgruening/galaxy-rna-seq:17.01
+FROM bgruening/galaxy-rna-seq:dev
 
 MAINTAINER Björn A. Grüning, bjoern.gruening@gmail.com
 
--- a/RNA-Seq/docker/Dockerfile
+++ b/RNA-Seq/docker/Dockerfile
@@ -1,15 +1,14 @@
 # Galaxy - RNA-Seq training material
 
-FROM bgruening/galaxy-rna-seq
+FROM bgruening/galaxy-rna-seq:17.01
 
 MAINTAINER Björn A. Grüning, bjoern.gruening@gmail.com
 
 ENV GALAXY_CONFIG_BRAND RNA-Seq training
-ENV ENABLE_TTS_INSTALL True
 
-# Enable Conda dependency resolution
-ENV GALAXY_CONFIG_CONDA_AUTO_INSTALL=True \
-    GALAXY_CONFIG_CONDA_AUTO_INIT=True
+ADD fetch_and_index_genomes.yaml data_manager_rna_seq.yaml
+
+RUN startup_lite && sleep 60 && run-data-managers -u admin@galaxy.org -p admin --config data_manager_rna_seq.yaml -g http://localhost:8080
 
 # Add a Galaxy tour
 # ADD <URL to dedicated tour in https://github.com/galaxyproject/galaxy-tours> $GALAXY_ROOT/config/plugins/tours/rna-seq.yaml
--- a/RNA-Seq/docker/Dockerfile
+++ b/RNA-Seq/docker/Dockerfile
@@ -6,7 +6,7 @@ MAINTAINER Björn A. Grüning, bjoern.gruening@gmail.com
 
 ENV GALAXY_CONFIG_BRAND RNA-Seq training
 
-ADD fetch_and_index_genomes.yaml data_manager_rna_seq.yaml
+ADD data_manager_rna_seq.yaml data_manager_rna_seq.yaml
 
 RUN startup_lite && sleep 60 && run-data-managers -u admin@galaxy.org -p admin --config data_manager_rna_seq.yaml -g http://localhost:8080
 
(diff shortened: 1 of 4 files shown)


RNA-Seq/docker/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
 
 MAINTAINER Björn A. Grüning, bjoern.gruening@gmail.com
 
 ENV GALAXY_CONFIG_BRAND RNA-Seq training
-ENV ENABLE_TTS_INSTALL True
 
-# Enable Conda dependency resolution
-ENV GALAXY_CONFIG_CONDA_AUTO_INSTALL=True \
-    GALAXY_CONFIG_CONDA_AUTO_INIT=True
+ADD fetch_and_index_genomes.yaml data_manager_rna_seq.yaml
```
nsoranzo: Where does `fetch_and_index_genomes.yaml` come from?

bgruening:
Related PR: https://github.com/galaxyproject/galaxy/pull/3874

RNA-Seq/docker/Dockerfile
bgruening: nowhere, I renamed this in the last minute to be more meaningful.

bgruening:
If everyone agrees I would like to put the images under the `biocontainers` namespace. This one will be build here: biocontainers/galaxy-rna-seq-training

malloryfreeberg:
Just to make sure I'm understanding correctly, a `data_manager_rna_seq.yaml` is made that lists the genome indices that the user wants to have loaded into the Docker image?

bgruening:
@malloryfreeberg yes exactly. You can define the genome of interest and the indices that you want to have installed into the Docker container with this file - as long as there is a data-manager installed.

malloryfreeberg:
@bgruening This is a life-saver. Thanks so much!

bgruening:
@malloryfreeberg it can be that still tools are missing in this container, if this is the case please ping me and I will add them or show you howto add them. tools, data, tours, vis should be all possible now very easily.

malloryfreeberg:
@bgruening If this is supposed to cover all the RNA-seq tutorials, then `data_manager_rna_seq.yaml` also missing mm10 indices, which are needed for the de novo RNA-seq tutorial. I can add that. Where can we add tools that are missing?

RNA-Seq/docker/data_manager_rna_seq.yaml
```diff
@@ -0,0 +1,32 @@
...
+      params:
+        - 'dbkey_source|dbkey': '{{ item }}'
+        - 'reference_source|reference_source_selector': 'ucsc'
+        - 'reference_source|requested_dbkey': '{{ item }}'
+      # Items refere to a list of variables you want to run this data manager. You can use them inside the param field with {{ item }}
+      # In case of genome for example you can run this DM with multiple genomes, or you could give multiple URLs.
+      items:
+        #- ce8
+        - dm3
+        #- mm9
```
malloryfreeberg: Need mm10 for de novo RNA-seq tutorial

RNA-Seq/docker/data_manager_rna_seq.yaml
```diff
@@ -0,0 +1,32 @@
...
+        - all_fasta
+        - __dbkeys__
+
+    - id: toolshed.g2.bx.psu.edu/repos/devteam/data_manager_bowtie2_index_builder/bowtie2_index_builder_data_manager/2.2.6
+      params:
+        - 'all_fasta_source': '{{ item }}'
+      items:
+        #- ce8
+        - dm3
+        #- mm9
```
malloryfreeberg: Need mm10 for de novo RNA-seq tutorial

RNA-Seq/docker/data_manager_rna_seq.yaml
bgruening: @malloryfreeberg do we need dm3 at all? This was just a demonstration ... :)

malloryfreeberg: Yes, keep dm3 because it is used for the ref-based RNA-seq tutorial :)

bgruening:
@malloryfreeberg adding tools might need a little bit of coordination :)

This training image is based on the RNA-seq Docker image (I thought this is a nice idea). So if you think this tool belongs to a general RNA-seq image add it here: https://github.com/bgruening/galaxy-rna-seq/blob/master/rna_seq_tools.yml

The RNA-seq Image in turn is based on the NGS-preprocessing image, which is defined here: https://github.com/bgruening/docker-galaxy-ngs-preprocessing
This image is also the base for the ChIP-Seq image and the Exome-Seq image and so on. So if your tool is so general that it belongs somehow to prepossessing add it there - and it will pop-up in all the other flavors that build on top of https://github.com/bgruening/docker-galaxy-ngs-preprocessing.

If you think your tool is only very useful for the training create a new YAML file like I linked above and push it next to the Dockerfile :)

Not sure if this is to complicated - but with this we can share as many resources as possible.

bgruening:
@malloryfreeberg there is now an image: `docker pull biocontainers/galaxy-rna-seq-training`
It should container mm10 and dm3, let me know what you think.

PR #299, opened 2017-05-11
DevCorner fixes

abretaud:
Hey,
A few changes to the dev-corner slides before a session next week

- reworked the reserved variables with more "real" examples
- reworked the data collection part
- updated a few links


abretaud:
Is it because of the remaining broken urls that it's red? Ifixed some, but for others I don't know how to fix them...

bgruening:
Hi @abretaud! That is great! Thanks. Yes a few more links are broken, I hope we can fix this during our hackathon in 2 weeks.
Thanks!

PR #303, opened 2017-05-17
fix syntax of root readme

martenson:
I am still unclear about who is the target audience of this readme, but at least it renders well now. Also removed link to Planemo docs as it seemed unrelated.


bgruening:
Wow, that is that? Is that a github bug? This was changed 10 month ago last time and it worked for sure! However, thanks for fixing!

PR #318, opened 2017-05-23
Dropped problematic Hisat|hisat.png file; added favicon

nsoranzo:
Both `RNA-Seq/images/Hisat.png` and `RNA-Seq/images/hisat.png` are unused and they are considered the same file under crappy OS X, creating problems in git for training material developers.


dannon:
I think we've fixed this in parallel in multiple branches now, +1 :)

PR #321, opened 2017-05-23
Update tools.yaml

dsobral:
Tools necessary for the tutorial

### DIFF
--- a/Metagenomics/docker/tools.yaml
+++ b/Metagenomics/docker/tools.yaml
@@ -51,10 +51,6 @@ tools:
   owner: devteam
   tool_panel_section_label: "Graph/Display Data"
   
-- name: collapse_collections
-  owner: nml
-  tool_panel_section_label: "Collection Operations"
-  
 - name: newick_display
   owner: dcorreia
   tool_panel_section_label: "Graph/Display Data"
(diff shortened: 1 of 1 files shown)


Metagenomics/docker/tools.yaml
```diff
@@ -39,10 +39,36 @@
...
-- name: ucsc_twobittofa
+- name: suite_mothur
   owner: iuc
-  tool_panel_section_label: "Convert Formats"
+  tool_panel_section_label: "16S-Metagenomics"
   tool_shed_url: https://toolshed.g2.bx.psu.edu/
   install_resolver_dependencies: True
-  install_tool_dependencies: False
+  install_tool_dependencies: True
+- name: xy_plot
```
bgruening: can we include line breaks between every tool? This makes them more readable.

Metagenomics/docker/tools.yaml
```diff
@@ -39,10 +39,36 @@
...
+  owner: devteam
+  tool_panel_section_label: "Graph/Display Data"
+  tool_shed_url: https://toolshed.g2.bx.psu.edu/
```
bgruening: There 3 lines are by default enabled and can be removed.
```
  tool_shed_url: https://toolshed.g2.bx.psu.edu/
   install_resolver_dependencies: True
   install_tool_dependencies: True
```

Metagenomics/docker/tools.yaml
dsobral: Sure! It's a good idea.

Metagenomics/docker/tools.yaml
dsobral: Thanks!

PR #323, opened 2017-05-23
update metadata.yaml

vdda:
edam_ontology : "topic_3169" => chip-seq


bgruening:
Thanks!

topics/chip-seq/metadata.yaml
```diff
@@ -3,6 +3,7 @@ type: "use"
 title: "ChIP-Seq data analysis"
 summary: "ChIP-sequencing is a method used to analyze protein interactions with DNA."
 docker_image: "bgruening/galaxy-chip-seq-training"
+edam_ontology: "topic_3169"
```
bebatut: Can you remove the `topic_`? To make sure that whatever tool we use to parse the EDAM ontology we can find it

bgruening: isn't the topic part of the ontology?

PR #328, opened 2017-05-23
Simon added an introductory Assembly tutorial and slide deck.

Slugger70:
This is an introductory assembly tutorial and slide deck. Still to come is a more in depth slide deck and tutorial


bgruening:
Wonderful! A new topic! I will merge this and try to get it into the build branch, were we can adopt it to the new structure.

PR #331, opened 2017-05-23
Update metadata.yaml

vdda:
Added edam_ontology : "topic_0121"


bgruening:
@vdda we agreed to not do this, correct? We are only annotating the top metadata.yaml file.

PR #338, opened 2017-05-24
New small RNA-seq training material

malloryfreeberg:
First part of a new set of small RNA-seq training materials. The only addition here is the tutorial markdown file (which is incomplete at the moment, but usable). Dockerfile, tools.yaml, library_data.yaml, etc. are to be added after the training material restructuring happens to avoid conflicts with the current versions of these files for the RNA-seq tutorials.


bgruening:
Ok, let's get this in. This is already a valuable resource and I will try to help with the missing stuff as soon as the new infrastructure is in place.

malloryfreeberg:
Thanks @bgruening! I should be able to get most of the additional materials done, I just need will need to update with correct file names/paths once each tutorial is in its own folder.

bgruening:
@malloryfreeberg that's awesome, thanks a lot!

PR #340, opened 2017-05-24
Update tools.yaml

daltis20012009:
A range of tools were added. Mo, we need to check if tool_section_labeling is consistent with slides/tutorial


bebatut:
Thanks a lot

PR #352, opened 2017-06-09
Mothur SOP tutorial updates

shiltemann:
ping @bebatut 

- simplified the `make.contigs` step and minor tweaks to tutorial text
- added workflow .ga file
- updated data library yaml file to use individual datasests
- updated tools.yaml file

*think* everything should now be complete for this tutorial apart from a tour


bgruening:
:100: !!!

PR #354, opened 2017-06-10
Merge the new structure to master

bebatut:
To continue the work of @shiltemann in #353, we could merge this branch to master: the contribution of the hackathon will be online as well as the training content we are developing for GCC, and it will relieve some of the pressure to get the new metalsmith templating up and running. 

ping @shiltemann, @bgruening 

Bérénice

PR #358, opened 2017-06-15
change slide deck type

shiltemann:
ping @nsoranzo, I think this fixes what you meant in your [comment](https://github.com/galaxyproject/training-material/pull/354#issuecomment-308826275)?

### DIFF
--- a/topics/introduction/tutorials/galaxy-intro-get-data/slides.html
+++ b/topics/introduction/tutorials/galaxy-intro-get-data/slides.html
@@ -1,5 +1,5 @@
 ---
-layout: base_slides
+layout: introduction_slides
 topic_name: introduction
 tutorial_name: galaxy-intro-get-data
 logo: GTN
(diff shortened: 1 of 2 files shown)


nsoranzo:
Thanks @shiltemann! Not sure that's enough:
- `data-library.yaml` and `tools.yaml` are not needed because these are slide decks
- there's ``type: "tutorial"`` in `metadata.yaml`
- I think they should be in `topics/introduction/slides/`

shiltemann:
I agree with you, but I think the template currently does not support multiple presentations in the slides directory, think it was originally conceived as being a single introduction to the topic in general, which is something we clearly need to rethink, but I am not sure how useful it is to spend time updating the jekyll template with the new metalsmith just around the corner.. (but anybody is free to do so if they like)

These changes at least remove the empty objectives/goals/keypoints slides from presentation which I think was the immediate concern? The repo is a little bit in limbo right now, halfway prepared for the new templating and halfway reverted back to using jekyll again to at least have something functional for GCC ..what to you say we just merge this and put the rest of your points on a TODO list for the new template?

nsoranzo:
Agreed, thanks @shiltemann!

PR #369, opened 2017-06-21
update config file to jekyll 3.5

bgruening:
I tried jekyll on a fresh VM and since there was a new release (3.5) I got this latest release. Unfortunately, this means I needed to patch the config file as outlined in this PR.

People, that want to try this next week, probably needs this. Old installation will probably not work out of the box and either needs to upgrade or revert this change.

What do you think? Merge or not?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -38,10 +38,11 @@ kramdown:
 
 # Handling Reading
 exclude:
+  - shared/font-awesome/src/
   - Gemfile
   - Gemfile.lock
 
 # Plugins
-gems:
+plugins:
   - jekyll-feed
   - jemoji


nsoranzo:
From https://jekyllrb.com/news/2017/06/15/jekyll-3-5-0-released/ :
The `gems` key in the `_config.yml` is now `plugins`. This is backwards-compatible, as Jekyll will gracefully upgrade `gems` to `plugins` if you use the former.

bgruening:
It does not at least in my case.

bgruening:
@nsoranzo do you think this is ok to merge?

nsoranzo:
I've updated to 3.5 and seems to work for me without changes.

bgruening:
This is what I get:
```
 % make serve                                          
bundle exec jekyll serve
Configuration file: /home/bag/projects/code/training-material/_config.yml
       Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.
            Source: .
       Destination: ./_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
  Liquid Exception: Could not locate the included file 'tests/stacked.html' in any of ["/home/bag/projects/code/training-material/_includes"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source. in shared/font-awesome/src/test/2.3.2.html
jekyll 3.5.0 | Error:  Could not locate the included file 'tests/stacked.html' in any of ["/home/bag/projects/code/training-material/_includes"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source.
Makefile:20: die Regel für Ziel „serve“ scheiterte
make: *** [serve] Fehler 1
```

willdurand:
It looks like Travis has successfully built the website with Jekyll 3.5, thanks to #432 (you can take a look at the [`gh-pages` branch](https://github.com/galaxyproject/training-material/tree/gh-pages)).

bgruening:
@willdurand done!

willdurand:
@bgruening I would have kept your PR actually. It makes sense to exclude FA sources and to use `plugins` instead of `gems` to get rid of the warning.

PR #377, opened 2017-06-22
Metagenomics tutorial for amplicon and shotgun data

bebatut:
Hi,

Here are our tutorial and introduction slides for our workshop (with @shiltemann) at GCC2017.

There is also some css fixes for the tutorial and slides formatting.

Bérénice


bgruening:
Ok, took me a while but I learned a lot :)
Thanks!

PR #378, opened 2017-06-22
Important clarifications/fixes for Conda/Container slides.

jmchilton:
For these smaller commits that just fix things should I opening PRs or committing directly - I can do either no problem - just curious what the best practice is?


bgruening:
We don't have any police in place currently. I would argue if you are the maintainer it is ok to push small changes directly.

PR #384, opened 2017-06-27
remove metadata from train for now

bgruening:


bgruening:
ping @bebatut

bebatut:
Great!
We need also to remove them from the templates and be careful in the tutorial!

PR #403, opened 2017-06-30
sRNA tutorial: small fixes for display

shiltemann:
ping @malloryfreeberg 

made some fixes for the display of the hand-on/agenda boxes, one of them was missing a title, I made one based on what it sounded like to me but you may have a better suggestion :)


bgruening:
Thanks! :)

PR #406, opened 2017-07-07
Add a metadata parameter to enable/disable a tutorial in the topic page

bebatut:
Hi,

I added a metadata parameter to be able to hide a tutorial in its topic page: it can not then be accessed via the topic page (only via its full URL).

Bérénice


bgruening:
Can we make enable true the default option to not being forced to write it?

bebatut:
Yes we can do that. Do I keep it in the metadata.yaml files?

bgruening:
I would remove it. This is very special and people should not worry about it.

bebatut:
@bgruening I removed the `enable` for most tutorial's metadata, except the ones to disable

bgruening:
Thanks looks good.
@bebatut can you rebase.

bebatut:
@bgruening Rebased!

martenson:
@bebatut is this removing all links to that tutorial from the training site? If so we should be careful not to inhibit crawlers' indexing efforts and maybe link from somewhere else to it.

bebatut:
@martenson The tutorial page will be still generated, but not indexed in the topic page (with other tutorials for the topic)

topics/assembly/metadata.yaml
```diff
@@ -24,10 +24,10 @@ material:
 -
     title: "Introduction to Genome Assembly"
     type: "tutorial"
+    enable: "true"
```
bgruening: can be removed

topics/admin/metadata.yaml
```diff
@@ -8,6 +8,7 @@ material:
 -
     title: "Introduction"
     type: "introduction"
+    enable: "true"
```
bgruening: can be removed

bgruening:
Thanks @bebatut !

PR #413, opened 2017-07-11
fix silently failing links

shiltemann:
As noticed by @hrhotz some links were broken but not producing an error in vl 

I changed vl to debug mode so that these errors at least show up in output (for example: [here](https://travis-ci.org/galaxyproject/training-material/builds/252420744#L2823)), even though job will still exit successfully, 

also fixed any other links failing this way (many were due to using wrong scheme http/https)

remaining bad links: 
- bunch of reveal.js links in `shared/slides/project_presentation.html` but maybe we can just remove that now that we've moved to remark? (@bebatut?)
- https://www.genome-cloud.com/ in `topics/introduction/tutorials/options-for-using-galaxy/slides.html` not sure where that should point


bebatut:
@shiltemann I will clean any mention to `reveal.js` (in an other branch...)

bebatut:
@shiltemann I will merge this one and open an issue for the link in `topics/introduction/tutorials/options-for-using-galaxy/slides.html`

bebatut:
@shiltemann Thanks a lot!

PR #416, opened 2017-07-11
Some cleaning of the repo

bebatut:
Hi,

I started to some cleaning:

- I removed the metalsmith start: we decided to not move to Metalsmith as templating system in the next months
- I removed some staging files

Bérénice


dannon:
I missed conversations about not swapping templating for the next few months, can someone catch me up on that?  (maybe from GCC?)

That said, I think it's totally fine to remove these from here and keep the metalsmith implementation in the other branch.

bgruening:
Hi @dannon!
I guess what @bebatut means is that we stick for some more time to the old system until the old system is really ready for prime time and you are recovered.

We needed to take a decision at GCC (slightly before) to stay with the old templates to create our training materials and to introduce new people to it. This state in between was not productive and to much pressure on you.

We will from now on concentrate on content, @nekrut seems to be fine with this style for some time :) maybe a few css changes ;) ... and if you have time again (no pressure) we can migrate this to methalsmith.

dannon:
@bgruening Hey!  Yeah, I talked to @bebatut a bit on gitter about it.  Sounds fine to me for an approach for now.  In retrospect, we should have had more of a build pipeline tutorial or something at the Cambridge hackathon, instead of just a few people focusing on 'just doing it', and done more to involve more people in the actual conversion.

We can do that down the road, though.  I just need to be able to more effectively convince folks that  the hub-style metalsmith build pipeline is more flexible and easier to maintain (for everyone!) than Jekyll :)

bgruening:
:) nicely put!

PR #419, opened 2017-07-12
Support mathematical expressions written in LaTeX

bebatut:
Hi,

I added support for mathematical expressions written in LaTeX: they will be rendered using [MathJax](https://www.mathjax.org/)

Bérénice

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -19,5 +19,5 @@
         </div>
     </body>
     <script src="{{ site.url }}/assets/js/landing_page.js"></script>
-    <script type="text/javascript" src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js"></script>
 </html>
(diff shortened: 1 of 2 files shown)


shiltemann:
cool! :) ..I tried to add an example math expression to the tutorial, but it was not working for me locally, what am I doing wrong?

shiltemann:
hmm, this is what my console says: https://www.mathjax.org/cdn-shutting-down/

shiltemann:
(but even with one of the suggested updated links it does not work for me)

bebatut:
@shiltemann Strange!
I added an example in the content tutorial. Can you check if it correctly rendered, please?
Otherwise, can you send me a snapshot of how it is rendered?

shiltemann:
ok, it works now, may have been a script blocker orso on my end, sorry for the confusion, looks great :)

should we switch to the cloudflare link https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js as suggested in t https://www.mathjax.org/cdn-shutting-down/ ?

bebatut:
I switched!

shiltemann:
great, thanks!

bgruening:
You both just rock! Can't follow with all the good stuff here!

PR #423, opened 2017-07-17
added tour

polkhe:
added tour


bgruening:
Nice, this looks great!

PR #431, opened 2017-07-26
Chip seq

joachimwolff:
Updated chip-seq tutorial. Thanks to @erxleben. 

@malloryfreeberg Can you validate the answers? Thanks a lot!


bgruening:
This is getting better and better!

PR #432, opened 2017-07-27
Site building in Travis

bebatut:
Hi,

I wanted to add several plugins to Jekyll, for example to manage the bibliography or the figure caption (next PR...).
But as the website is currently building by GitHub, we are limited to the plugins approved by GitHub and we can not custom ones.

One way to overpass that is to use Travis to build the website, push it to a `gh-page` branch. and tell to GitHub to serve this branch. 

I then change a bit Travis to do that. I tested it on my fork: http://bebatut.fr/training-material/

And because we are building the site on Travis, we can even use the `make check` command of the (still open) PR #424 in `.travis.yml` to check the correct building of the site and checking the links before pushing it.

Bérénice

PR #433, opened 2017-07-27
Add plugin for figure caption

bebatut:
⚠️  Depends on #432 

---

Hi,

@willdurand developed a plugin to automatically add caption to figure using either the `title` or the `alt` information in the markdown figure (`alt`, with a preference for the `title`).
I added the description to use it in the tutorial for writing tutorials 😛 

Bérénice


shiltemann:
cool beans! this is really nice, thanks both of you :)

_plugins/jekyll-figurify.rb
```diff
@@ -0,0 +1,67 @@
...
+  class Figurify < Jekyll::Generator
+
+    safe true
+
+    def initialize(config)
+      @config = config['figurify'] ||= {}
+    end
+
+    def generate(site)
+      puts "\n      Applying figurify plugin ✨"
```
willdurand: @bebatut you should get rid of this line. It was fun but now it makes me crazy and I cannot see it anymore 😛

bebatut: 😄  I removed it

willdurand:
For the record, the plugin parses the markdown content and transforms markdown images into:

```
<figure id="figure-999">
  <img src="url" alt="alt" title="title">
  <figcation>Figure 999: title</figcaption>
</figure>
```

The plugin does that before the HTML transformation (so after having applied the plugin, you get some HTML into your markdown, which is perfectly valid but still worth mentioning here).

PR #435, opened 2017-07-28
Updated captions, typos and some answers.

joachimwolff:


willdurand:
@joachimwolff `alt` are certainly the most important information on a web page as it is super-useful for people with disability. In general, `alt` should contain the description of the image (alternative content) and `title`, well a title (additional content).

willdurand:
@blankclemens could you open a PR with proper alternative contents please? 🙏

blankclemens:
As far as I can see there are alt tags defined. I'm trusting the author of the tutorial that they contain the intended information, but maybe @joachimwolff wants to check again.

joachimwolff:
I think it would be nice to have the feature if no `alt` information is given that then the caption is used.

bebatut:
We can do that but I am not sure it is a good idea. The alt information and the title information are 2 different types of information. Alt information is more a description of the image in case the image does not appear or for blind peoples using a audio description. To make the training material more accessible, we should then add more information in the alt text than the just title of the image.

joachimwolff:
As I wrote: caption + title if and only if no alt information is provided aka 'better than nothing'.

willdurand:
In #439, it has been decided to not fallback to `alt` if no `title` is supplied for two reasons:

- enforcing `alt` content
- deciding not to add a figure on a specific image by omitting its title

PR #439, opened 2017-07-31
figurify plugin: Remove the alt text as possible caption

bebatut:
Hi,

ping @willdurand 
I removed the alt text as possible caption text to use only the title

Bérénice


willdurand:
Yes, I think it is better that way. It will enforce proper alternative contents and titles only when they make sense.

shiltemann:
makes sense, thanks @bebatut :)

PR #445, opened 2017-07-31
Site proofing

shiltemann:
- Replaces vl check with htmlproofer
- Enables travis testing on the gh-pages branch so that we can set up a daily travis cron job on the served site to continually test links
- Badge on readme now reflects build status of gh-pages branch (this will fail for a while until we go back and add alt-texts for images in existing materials, but maybe this will be good encouragement to do this sooner rather than later :) )

there are also some more broken links reported by htmlproofer, will fix those in a later PR

ping @bebatut

### DIFF
--- a/topics/assembly/tutorials/debruijn-graph-assembly/tutorial.md
+++ b/topics/assembly/tutorials/debruijn-graph-assembly/tutorial.md
@@ -18,7 +18,7 @@ SPAdes is a de novo genome assembler written by Pavel Pevzner's group in St. Pet
 > In this tutorial, we will deal with:
 >
 > 1. [Get the data](#get-the-data)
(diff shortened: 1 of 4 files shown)


topics/assembly/tutorials/debruijn-graph-assembly/tutorial.md
```diff
@@ -18,7 +18,7 @@ SPAdes is a de novo genome assembler written by Pavel Pevzner's group in St. Pet
 > In this tutorial, we will deal with:
 >
 > 1. [Get the data](#get-the-data)
```
bebatut: We can even use: 

```
> 1. TOC
> {:toc}
```

to avoid such issue

shiltemann: ah yes, good point ..will add that to my PR with fixes for the broken links reported by htmlproofer :)

PR #452, opened 2017-08-02
Fix typo

blankclemens:


willdurand:
I know it is only a period that has been contributed in this PR, but I dont think it is wise to merge his own PRs. I guess this should be discussed (and maybe written in the `CONTRIBUTING` file) but you should either wait for someone else to merge your PRs or get a 👍 .

blankclemens:
Thumbs up added. 👍

blankclemens:
For the record: I do agree in general, but taking other peoples time with an added period is not something I personally would do - which would lead to the classic: Too much regulations kill productivity.

PR #458, opened 2017-08-03
HTML proofing

shiltemann:
This fixes more broken links identified by htmlproofer. Aside from missing alt-texts on images and the broken links already discussed earlier, the other failures currently are:

- http://www.france-bioinformatique.fr (but this just looks like a temporary misconfiguration of the website)
- in `ref-based-rad-seq` tutorial: link to Zenodo: http://doi.org/10.5281/zenodo.218574 appears broken, it says the data was removed by the uploader. Has this been re-uploaded under some alternate record maybe? ping @yvanlebras


_layouts/introduction_slides.html
```diff
@@ -18,7 +18,7 @@
         {% elsif material.hands_on == "github" %}
 - [{{ material.title }}]({{ site.github_repository }}/tree/master/topics/{{ topic.name }}/tutorials/{{ material.name }}.md)
         {% elsif material.slides == "yes" %}
-- [{{ material.title }}]({{ site.url }}/topics/{{ topic.name }}/slides/{{ material.name }})
+- [{{ material.title }}]({{ site.url }}/topics/{{ topic.name }}/tutorials/{{ material.name }}/slides.html)
```
nsoranzo: :+1: I was going to open a PR with this same change!

README.md
```diff
@@ -1,4 +1,4 @@
-[![Gitter](https://badges.gitter.im/Galaxy-Training-Network/training-material.svg)](https://gitter.im/Galaxy-Training-Network/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![Build Status](https://travis-ci.org/galaxyproject/training-material.svg?branch=gh-pages)](https://travis-ci.org/galaxyproject/training-material)
+[![Gitter](https://badges.gitter.im/Galaxy-Training-Network/training-material.svg)](https://gitter.im/Galaxy-Training-Network/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![Build Status](https://travis-ci.org/galaxyproject/training-material.svg?branch=master)](https://travis-ci.org/galaxyproject/training-material) html-proofer: [![Deploy Status](https://travis-ci.org/galaxyproject/training-material.svg?branch=gh-pages)](https://travis-ci.org/galaxyproject/training-material)
```
willdurand: Can we have the badges in a table?

_config.yml
```diff
@@ -17,7 +17,7 @@ title: "Galaxy Training!"
 email: your-email@domain.com
 description: A collection of tutorials generated and maintained by Galaxy Community Members across the world
 url: "https://galaxyproject.github.io/training-material/" # the base hostname & protocol for your site, e.g. http://example.com
-github_repository: "https://github.com/galaxyproject/training-material"
+github_repository: "https://github.com/galaxyproject/training-material/"
```
willdurand: By adding the leading slash, we generate `//` URLs, which are not very nice. It would be better to make sure that all links have the following form: `{{ site.url}}/foo`.

willdurand: By adding the leading slash, we generate `//` URLs, which are not very nice. It would be better to make sure that all links have the following form: `{{ site.url}}/foo`.

_layouts/home.html
```diff
@@ -73,17 +73,17 @@
 
     <h1>Acknowledgment and Funding</h1>
 
-    <p>We would like to thank all contributors to our Galaxy courses, especially those mentioned in the <a href="CONTRIBUTORS.md">Contributors list</a>, the Galaxy community for their constant support and our funding partners.</p>
+    <p>We would like to thank all contributors to our Galaxy courses, especially those mentioned in the <a href="{{site.github_repository}}blob/master/CONTRIBUTORS.md">Contributors list</a>, the Galaxy community for their constant support and our funding partners.</p>
```
willdurand: I would add a leading slash before `blob/...`

topics/epigenetics/slides/index.html
```diff
@@ -52,7 +52,7 @@
 
 ---
 
-# [Hands-on](https://galaxyproject.github.io/training-material/topics/epigenetics/tutorials/methylation-seq/tutorial.html)
+# [Hands-on]({{site.url}}topics/epigenetics/tutorials/methylation-seq/tutorial.html)
```
willdurand: Here also I would encourage the use of a leading slash.

willdurand: The reason is `site.url` could be empty (and that would still be valid) but then all your URLs will be relative to the current page, i.e. most of these URLs will be wrong. With a leading slash, you avoid this error as `/...` will always point to the right URL.

shiltemann: sounds good, will change this everywhere :)

bgruening:
Took me a while to go through it! This is a big chunk of great work! Thanks a lot @shiltemann!

PR #462, opened 2017-08-04
FreeBayes was not in Phenotype Association section

willdurand:
Currently attending the workshop on that topic, and this was incorrect.


bgruening:
We should in general avoid the category as this can change from instance to instance.

Issue #304, opened 2017-05-22
Infrastructure: Refining static content generation framework for training-materials


nekrut:
The current jekyll-based infrastructure for deployment of training-materials was a good first step, but significant issues remain. This issue is listing some of them:

- [x] The styling for site is immature. It is not pleasant to the eye (although this is of course subjective). One solution is to adopt the same styling used for galaxy-hub. 
- [ ] The current training materials styling allows for injection of styles using {{ }} syntax. If the styling is redone in galaxy-hub-like way we should keep this functionality.
- [x] Some tutorials will have slides, some won't. Some will have docker images, some won't. Some will cover main site, and some will be generic. Top level menu should be easily configurable to address this. There are tutorials that have docker icons at the top level menu, but these are empty.
- [x] I strongly dislike icons shown within tutorials (such as Pencil etc.). We should stick with a single style. I would prefer using fontsawesome throughout (or any other consistent, well designed library). 
- [ ] If metalsmith is used as the static content generator is should be possible to view changes in real time. The current version for galaxy-hub is waaay too slow. It is very painful to edit current galaxy-hub content because every reload (via `make serve`) takes minutes to finish.
- [x] there has to be a way to "hide" tutorials. Some existing tutorials are incomplete, have no docker images, have no meaningful slides, or otherwise broken. It should be possible to label these at "WIP" and make sure that they do not appear in the production site.

Slugger70:
Hackathon Group 2 - We think this is a good idea and should be available as guidelines and templates for creation of new materials. e.g. Slide examples for different slide types - titles, images, text, video etc.. Then templates/examples/guides for tutorial documents with reveal/hints etc...

bebatut:
@nekrut Do you think we can close this issue?

bebatut:
@nekrut Do you think we can close this issue?

bebatut:
I think this issue is outdated now. I will close it. We can reopen it if needed :smile:

PR #463, opened 2017-08-04
Refactor CSS to more look like the "Hub"

willdurand:
Hi there!

I did a complete CSS redesign (and a bit of HTML too) to match the Galaxy Hub look and feel. It is deployed here: http://williamdurand.fr/training-material/, so you can test it. Emoji (like `:stuff:`) are not rendered as I used `gh-pages` to build the website (neither Travis) but it is working.

I tried to come up with the smallest CSS possible (well, SASS actually but it is built-in in Jekyll). Therefore, all the credits go to Bootstrap. It is now fully responsive and should be slightly more readable. I changed the color scheme for the tutorial boxes with @bebatut and the help of http://paletton.com.

The PR gets pretty big, next chunk will be about replacing the Emoji icons in the "tutorial boxes" to FontAwesome icons, and then maybe redesign other parts. Let me know what you would like to do next!

🔥 Test it at: http://williamdurand.fr/training-material/

---

The logo (top left corner) was updated in https://github.com/galaxyproject/training-material/pull/463/commits/2481e59f46b81d91f320dc3b3ad7136209e58be5.


bgruening:
Thanks @willdurand!

@galaxyproject/training can you please vote here if you think this is good to go and an improvement to the old style.

martenson:
💯

jxtx:
For maintainibility, could we share a single set of base stylesheets between hub and here (and may end even Galaxy itself)?

willdurand:
> For maintainibility, could we share a single set of base stylesheets between hub and here (and may end even Galaxy itself)?

The base stylesheet is Bootstrap I guess. See the [`main.scss`](https://github.com/willdurand/training-material/blob/refactor-css/assets/css/main.scss) file, it is very specific to the tutorials.

martenson:
@jxtx That would probably mean bringing LESS to the build process to hub and here?

jxtx:
There is definitely stuff that is custom, but if we based it off the existing Galaxy styles then the bar sizes would match, fonts, base colors, etc.

jxtx:
(Also, I don't t care about less vs sass, we should just use one thing everywhere).

willdurand:
Yup, this can be done later anyway, especially if it involves Galaxy and Hub. The idea was to get rid of the old CSS and making the website a bit more appealing/responsive.

jxtx:
Fine with merging this now, but let's keep going!

nsoranzo:
Looks sleek! I noticed a couple of issues:
- I found a few "500 Internal Server Error", e.g. http://williamdurand.fr/training-material/topics/assembly/
- the "hat" logo is now used for "Hands-on", while previously a similar "hat" logo was used for "Slides", maybe that can be confusing?

willdurand:
@nsoranzo GitHub is having troubles again (for the 500)

willdurand:
> the "hat" logo is now used for "Hands-on", while previously a similar "hat" logo was used for "Slides", maybe that can be confusing?

Yes. New logo for slides is the one from Slideshare, which seems to make more sense that the hat. Do you think we should switch back to the previous logos?

nsoranzo:
> Yes. New logo for slides is the one from Slideshare, which seems to make more sense that the hat. Do you think we should switch back to the previous logos?

Don't know, mainly raising the point for discussion. Maybe a mouse logo for the tutorial?

willdurand:
This is a good point! Here are all the available icons: http://fontawesome.io/icons/ 😸

bgruening:
What about http://emojipedia.org/female-teacher/

willdurand:
I had that in mind too but read in #304 that FontAwesome should be preferred.

martenson:
For training site we can afford to expand to other icon libraries I believe.

bebatut:
In the navbar, the emoji will look weird. I thought it would be better not having 2 different icons for same things and having icons in the navbar adds a visual hint

nsoranzo:
@willdurand What about http://fontawesome.io/icon/laptop/ for tutorials?

jxtx:
For getting the styles more uniform, all sites could start with this as a base: https://github.com/jxtx/galaxy-bootstrap

That's where the styles in Galaxy proper came from. It needs to be updated and packaged with all the modern stuff, but there it can produce a bootstrap.min.css with Galaxy styles ready to use.

willdurand:
Just replaced the navbar logo with a better one :)

bebatut:
ping @galaxyproject/training: we would like to have more feedbacks 😄

shiltemann:
I like it! thanks a lot @willdurand :)

willdurand:
(rebased)

bgruening:
I think this was open long enough and I haven't heard any negative things - also counting the :+1: here is a clear `merge this` - and here we go!

Thanks @willdurand and all others for feedback. Keep the PRs coming!

PR #468, opened 2017-08-07
Add functional analysis to metaproteomics tutorial

blankclemens:


bgruening:
I discussed with Clemens about the heavy SQL task. We agree that this can be quite hard, but this is how GalaxyP is currently teaching this material, which I think is fair.
Clemens also proposed to have this SQL task replaces by a small workflow which might be a nice idea to offer this in addition.

Thanks @blankclemens!

PR #474, opened 2017-08-12
Minor UI improvements

willdurand:
Hi!

Here are a few UI improvements:

- Replace some more icons (missed the `:book:` ones)
- Add colored background on code snippets
- Wrap text to avoid problems on small devices (long words break the layout)
- Display two columns on bottom homepage for medium/large devices


martenson:
Thanks @willdurand !

PR #475, opened 2017-08-12
Fix broken link on homepage

willdurand:
I noticed a link was broken, the other was not configurable.

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -79,7 +79,7 @@ layout: base
         </div>
 
         <h2>Acknowledgment and Funding</h2>
-        <p>We would like to thank all contributors to our Galaxy courses, especially those mentioned in the <a href="CONTRIBUTORS.md">Contributors list</a>, the Galaxy community for their constant support and our funding partners.</p>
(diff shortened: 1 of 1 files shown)


_layouts/home.html
```diff
@@ -79,7 +79,7 @@
         </div>
 
         <h2>Acknowledgment and Funding</h2>
-        <p>We would like to thank all contributors to our Galaxy courses, especially those mentioned in the <a href="CONTRIBUTORS.md">Contributors list</a>, the Galaxy community for their constant support and our funding partners.</p>
+        <p>We would like to thank all contributors to our Galaxy courses, especially those mentioned in the <a href="{{ site.github_repository }}/blob/master/CONTRIBUTORS.md">Contributors list</a>, the Galaxy community for their constant support and our funding partners.</p>
```
martenson: would it make sense to have the `/blob/master/` in a variable too?

willdurand: This I am not sure since it is unlikely to change.

bgruening:
Mh, I wonder why this was not catched by travis ... :(
Thanks Will.

PR #479, opened 2017-08-28
metaproteomics minor improvements

bernt-matthias:
Added a tools file for the metaproteomics tutorial. 

Missing is still the entry for "Split Tabular Columns" - I did not know which tool the corresponds to. 

I was wondering if there is way to specify visualizations that need to be installed/configured. For this tutorial the required Unipept Tree viewer seems not to work by default.

### DIFF
--- a/topics/proteomics/tutorials/metaproteomics/tools.yaml
+++ b/topics/proteomics/tutorials/metaproteomics/tools.yaml
@@ -1,0 +1,13 @@
+---
+api_key: admin
+galaxy_instance: http://localhost:8080
+tools:
+- name: peptideshaker
+  owner: galaxyp
+  tool_panel_section_label: "Proteomics"
+- name: unipept
+  owner: galaxyp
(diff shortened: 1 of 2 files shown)


bgruening:
@bebatut do you know where the tool "Split Tabular Columns" is living?

bernt-matthias:
Fount the "Split Tabular Columns" and successfully completed the tutorial. 

Two questions from my side remain: 
- is searchgui necessary or is this included in the peptideshaker tool
- There is a tool query_tabular and a suite_query_tabular (same date in the toolshed). Which should be used?

shiltemann:
the tool `query_tabular` is part of the suite `suite_query_tabular` (along with `filter_tabular` and `sqlite_to_tabular`). Seems you need 2/3 of these for the tutorial, so you can either install the suite and get all three, or explicitly install just the two tools you need.

bgruening:
Thanks @bernt-matthias!

PR #478, opened 2017-08-25
Update Galaxy 101 tutorial

nsoranzo:
- Update the "Galaxy 101" tutorial to use the `hg38` genome and the latest dbSNP version from the UCSC Genome Browser. Also update the images to show the UI of Galaxy 17.05.

- A couple of small fixes for the for peaks2genes tutorial.

PR #480, opened 2017-08-29
Alt attributes to images for HTML proofer

polkhe:
Alt attributes to images for HTML proofer


nsoranzo:
@polkhe This is unfortunately going to conflict with https://github.com/galaxyproject/training-material/pull/478 where I've updated the 101 tutorial and also added alt attributes.

bebatut:
@nsoranzo I will merge first your PR and then fix the conflicts here

PR #488, opened 2017-08-31
Make sure the Travis build fails when website cannot be built

willdurand:
Hi,

This PR prevents the website to be broken by ensuring that an error at build time is caught by Travis. I have also added a check to the `publish` script to avoid publishing an empty website. Both changes should avoid yesterday's situation.

Note: the `|| return 0` should be removed as soon as all htmlproofer errors are fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,13 +34,15 @@ install:
 - make install
 
 script:
-- set -e
 # Check website
 - |
+  set -e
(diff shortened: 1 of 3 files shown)


willdurand:
Please see this diff: https://github.com/galaxyproject/training-material/compare/verify-travis-must-fail.

First commit should have failed (change in a YAML file that makes it invalid) but it does not. That's the current `master` behaviour.

Second commit is the one proposed here, which now fails the build.

PR #493, opened 2017-08-31
To fix HTMLProofer

bebatut:
To fix #492

PR #498, opened 2017-09-01
Activate correct Travis check

bebatut:
⚠️  Not merging before YAML linting and #493 
---
I removed the `|| return 0` in Travis, to make sure that Travis tests can fail


willdurand:
🎉

PR #503, opened 2017-09-02
Output of the "Map With BWA" Tool

Ewran:
Output of the "Map With BWA" Tool is a BAM format, but not SAM format.


bgruening:
Thanks @Ewran!

PR #504, opened 2017-09-03
Update tutorial.md

Ewran:
it was written as
"We expect that the replicates of the ChIP-seq experiments should be clustered more closely to each other than the replicates of the input samples.".
but I think replicates of the same samples clustered more closely to each other than replicates of different samples.
So it should be written as
"We expect that the replicates of the same samples (cell types) should be clustered more closely to each other than the replicates of the different samples.".

### DIFF
--- a/topics/chip-seq/tutorials/tal1-binding-site-identification/tutorial.md
+++ b/topics/chip-seq/tutorials/tal1-binding-site-identification/tutorial.md
@@ -200,7 +200,7 @@ Nowadays, there are many read alignment programs for sequenced DNA, `BWA` being 
 
 To assess the similarity between the replicates sequencing datasets, it is a common technique to calculate the correlation of read counts for the different samples.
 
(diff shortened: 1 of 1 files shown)


willdurand:
Hi @malloryfreeberg, what do you think of these changes?

malloryfreeberg:
The phrase "replicates of the different samples" doesn't make sense to me. If two experiments are from different samples, then they aren't replicates. 

A better wording would be: "We expect that replicate samples will cluster most closely to each other than to other samples." This sentence is the most conservative way of saying that we expect, for example, G1E Tal1 IP R1 to most closely cluster with G1E Tal1 IP R2, and not the G1E input replicates or the megakaryocyte Tal1 IP replicates or the megakaryoctye input replicates.

Ewran:
Thank you so much @malloryfreeberg , @willdurand , I revised it again.

willdurand:
thanks @Ewran!

PR #510, opened 2017-09-08
Minor updates in the N-TAILS and Secretome Prediction tutorials

Stortebecker:
Some corrections to the two tutorials on  "N-TAILS" and "Secretome Prediction". I do not quite get why so many commits are included, but its only the two tutorials.

Cheers, Flo


Stortebecker:
Thanks for the review @bebatut ! I fixed the problems. Thanks for the hint about the figure legends.

blankclemens:
Thanks!

PR #522, opened 2017-09-13
Add the possibility to have several introduction slide decks and fix introduction slide title

### LINKED ISSUE #516: Fix introduction slide titles
Since #507, we have the correct title in the intro slides. Yet, all of them have `Introduction` as title, which is not meaningful. Before #507, the title of the topic was used but that was actually a side-effect.

bebatut:
Hi,

This PR adds a feature to enable several introduction slide decks for the topics:

- Add a file in the `slides` directory (`file_name.html` for the example here)
- Reference it in the `metadata.yaml`: 

```
    title: "Introduction to the topic"
    type: "introduction"
    name: "file_name"
    slides: "yes"
    contributors:
      - contributor1
      - contributor2
```
All the introduction slide decks will appear in the topic page and also in the hands-on page with a dropdown menu.

I renamed all existing introduction slide decks into `introduction.html`. I also added a more explicit title than `Introduction` to fix #516.

The tests are failing because the links to the introduction slides changed.

But I have still some issues with the dropdown menu, for example if there is no introduction slide decks. And it is not looking so nice (ping @willdurand)

Bérénice


topics/training/metadata.yaml
```diff
@@ -8,7 +8,7 @@ docker_image:
 tutorial_galaxy_instance:
 material:
   -
-    title: "Introduction"
+    title: "Introduction ot training with Galaxy"
```
nsoranzo: s/ot/to/

bebatut: Thanks for spotting it

topics/admin/metadata.yaml
```diff
@@ -6,7 +6,7 @@ summary: "Administration of Galaxy Servers (setting up, Docker use, databases, .
 docker_image: ""
 material:
   -
-    title: "Introduction"
+    title: "Galaxy from an point of view"
```
shiltemann: forgot the word administrator?

_layouts/tutorial_hands_on.html
```diff
@@ -37,9 +37,21 @@
...
+                        <div class="dropdown">
+                            <a href="#" class="nav-link" data-toggle="dropdown" aria-expanded="false" title="Introduction slides">
+                                {% icon slides %} Introduction slides
+                                <b class="caret"></b>
+                            </a>
+                            <ul class="dropdown-menu">
+                                {% for material in topic.material %}
+                                    {% if material.enable != "false" %}
+                                        {% if material.type == "introduction" %}
+                                            <li><a href="{{ site.url }}/{{ page.url }}/slides/{{ material.name }}.html">{{ material.title }}</a></li>
```
shiltemann: I think `page.url` needs to be `topic.name`?

shiltemann: or rather `/topics/{{ topic.name }}`

topics/admin/metadata.yaml
bebatut: Thanks for spotting it

_layouts/tutorial_hands_on.html
bebatut: Fixed! Thanks

shiltemann: what do you think? should we just remove the `{{ site.url }}` from these types of links? if it starts with a slash it will just assume the root, and htmlproofer would catch these errors as then it's not checking links against the existing online site, but just the newly generated one, might make things easier to spot with restructurings like this?

shiltemann:
we may want to add this option to htmlproofer: `--allow-hash-href  If "true", ignores the "href" "#"`

Makefile
```diff
@@ -26,7 +26,7 @@ build: clean ## build files but do not run a server
 .PHONY: build
 
 check-html: build ## validate HTML
-	bundle exec htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/" ./_site
+	bundle exec htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/","#" --file-ignore "/.*\/files\/.*/" ./_site
```
shiltemann: theres an htmlproofer option for this too btw ` --allow-hash-href  If 'true', ignores the 'href' '#'`

bebatut: thanks! I did not check. Will change that

shiltemann:
looks good, thanks a lot @willdurand and @bebatut !

PR #524, opened 2017-09-13
Methylseq

joachimwolff:
Update of the tutorial to fit the new versions.

### DIFF
--- a/topics/epigenetics/tutorials/methylation-seq/tutorial.md
+++ b/topics/epigenetics/tutorials/methylation-seq/tutorial.md
@@ -181,7 +181,15 @@ The data we use in this tutorial is available at [Zenodo](https://zenodo.org/rec
 >    >    </details>
 >    {: .question}
 >
-> 3. **Replace column** {% icon tool %} Every chromosome is named different, the list to transfer the ensembl notation to ucsc notation is having more than 500 entries. To convert it you can use the tool **Replace column**. Choose for ```File in which you want to replace some values``` the bedGraph file and for ```Replace information file``` [this](https://raw.githubusercontent.com/dpryan79/ChromosomeMappings/master/GRCh38_ensembl2UCSC.txt) conversion file. For ```Which column should be replaced?``` choose ```Column: 1```, for ```Skip this many starting lines``` a ```1``` and for ```Delimited by``` ```Tab```. In case this tool is not available in your galaxy instance, we precomputed the files for you: Please import ```NB1_CpG.meth_ucsc.bedGraph```, ```NB2_CpG.meth_ucsc.bedGraph```, ```BT089_CpG.meth_ucsc.bedGraph```, ```BT126_CpG.meth_ucsc.bedGraph``` and  ```BT198_CpG.meth_ucsc.bedGraph```.
+>    > ### {% icon tip %} Tip: UCSC - Ensembl convert
+>    >
+>    > * Download the ```Replace information file``` for hg38 chromosome: [Download](https://raw.githubusercontent.com/dpryan79/ChromosomeMappings/master/GRCh38_ensembl2UCSC.txt) and import it to Galaxy.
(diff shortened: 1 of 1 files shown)


topics/epigenetics/tutorials/methylation-seq/tutorial.md
```diff
@@ -69,7 +69,7 @@ The data we use in this tutorial is available at [Zenodo](https://zenodo.org/rec
 > We will map now the imported dataset against a reference genome.
 >
 > 1. **Galaxy** {% icon tool %}: Search for the tool ```bwameth```
-> 2. **bwameth** {% icon tool %}: Select for the option ```Select a genome reference from your history or a built-in index?``` ```Use a built-in idex``` and here the human ```hg38``` genome.
+> 2. **bwameth** {% icon tool %}: Select for the option ```Select a genome reference from your history or a built-in index?``` ```Use a built-in index``` and here the human ```hg38``` genome.
```
bebatut: ` is better to use for inline code :)

topics/epigenetics/tutorials/methylation-seq/tutorial.md
```diff
@@ -69,7 +69,7 @@ The data we use in this tutorial is available at [Zenodo](https://zenodo.org/rec
...
 > 3. **bwameth** {% icon tool %}: Choose for the option ```Is this library mate-paired?``` ```Paired-end``` and use the two imported datasets as an input.
```
bebatut: Are all these steps part of the same step to once `bwameth` or do you need to run `bwameth` several times?
In the first case, it may be better (easier to understand) to do something like:

```
> 1. **bwameth** {% icon tool %} with
>     - "Select a genome reference from your history or a built-in index?" to `Use a built-in index`
>     - "Select a reference genome" to `hg38` genome
>     - "Is this library mate-paired?" to `Paired-end`
>     - "First read in pair" and "Second read in pair" to the two imported datasets as an input.
>    
>      Please notice that depending on your system this computation can take some time. If you want to skip this, we provide for you a precomputed alignment: Import ```aligned_subset.bam``` to your history.
>
 {: .hands_on}
```

What do you think?
We used this format in the metagenomics tutorials. I found it better during the workshop to find which parameters changing and with which values.

topics/epigenetics/tutorials/methylation-seq/tutorial.md
```diff
@@ -147,29 +147,18 @@ The data we use in this tutorial is available at [Zenodo](https://zenodo.org/rec
...
-> 8. **computeMatrix** {% icon tool %}: Use the file ```CpGIslands.bed```as ```Regions to plot``` and the in the previous step created bigwig file as the ```score file```.
-> 9. **computeMatrix** {% icon tool %}: Use for the option ```computeMatrix has two main output options``` the value ```reference-point```.
-> 10. **Galaxy** {% icon tool %}: Search for the tool ```plotProfile```.
-> 11. **plotProfile** {% icon tool %}: Choose for ```Matrix file from the computeMatrix tool``` the computed matrix from the tool ```computeMatrix```.
+>    > It can happen that you can not select the correct input file. In this case you have to add meta information about the used genome to the file.
+>    > * Click on the pencil of the correct history item. 
+>    > * Change ```Database/Build:``` to the genome you used.
+>    > * In our case the correct genome is ```Human Dec. 2013 (GRCh38/hg38) (hg38)```.
+>    {: .tip}
+> 3. **Galaxy** {% icon tool %}: Search for the tool ```computeMatrix```.
```
bebatut: Can you add a spare line before? Thanks

bebatut:
Thanks

joachimwolff:
I thought the test cases failed? Have you fixed anything?

bebatut:
Some links were failing because the server hosting FastQC documentation was down... I just relaunched Travis

PR #537, opened 2017-09-19
Skip travis testing on master and move build on publish script

### LINKED ISSUE #535: No website build if Travis is failing on Master
When Travis is failing on master, the website is not built, pushed on `gh-pages` and then not updated even if we want it (merge a failing PR for example). It sounds right. But...

When we create a new tutorial, there is some links to this tutorial (on topic page) that are only created once the website online. So Travis is failing on that and the tutorial can not de deployed... We need then to do it manually (see #534)

One solution would may be:
- Deactivate the `make check` on master branch to be sure that once we merge a PR (even if failing) the site is built, tested and pushed to `gh-pages`
      There is tests inside the build scripts. So if the build is not correct, nothing will be pushed to `gh-pages` and Travis will fail
- Protect the master branch to be not able to push directly on it

What do you think?

bebatut:
To fix #535

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ script:
         make check
       else
         make build
+      fi
     fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,11 @@ script:
     then
       htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/"
     else
-      make check
+      if [ "$TRAVIS_BRANCH" != "master" ]
+      then
+        make check
+      else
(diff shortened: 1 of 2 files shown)


bgruening:
rerun the tests

.travis.yml
```diff
@@ -41,7 +41,11 @@ script:
     then
       htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/"
     else
-      make check
+      if [ "$TRAVIS_BRANCH" != "master" ]
+      then
+        make check
+      else
+        make build
```
shiltemann: closing `fi` is missing for the nested `if` (or use `elif` instead)

PR #541, opened 2017-09-19
increase timeout for link checking

bgruening:
+ fix a bunch of linting issues.


Makefile
```diff
@@ -31,7 +31,7 @@ build: ## build files but do not run a server
 .PHONY: build
 
 check: build ## validate HTML
-	bundle exec htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/" ./_site
+	timeout 120s bundle exec htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/" ./_site
```
bebatut: We will probably need also to add `timeout 120s` in the `.travis.yml` for the `gh-page` branch

bgruening: Why do we run this test on this branch at all?

shiltemann: I set up nightly tests on the branch so that we are warned if links break on the site

bgruening: Ha, nice!

PR #553, opened 2017-09-29
Add upper-case letter for some tutorial title

bebatut:


shiltemann:
I want to merge this but cannot because the tests fail (nature links in unrelated tut) ..I think for link checking that is too strict, maybe we should make that test an [allowed failure](https://improveandrepeat.com/2014/01/travis-ci-and-allowed-failures/) in travis and just require that the build step does not fail?

PR #554, opened 2017-09-29
added tools and tour for proteomics-protein-id-sg-ps

polkhe:
added tools and tour for proteomics-protein-id-sg-ps

### DIFF
--- a/topics/proteomics/tutorials/protein-id-sg-ps/tools.yaml
+++ b/topics/proteomics/tutorials/protein-id-sg-ps/tools.yaml
@@ -2,6 +2,22 @@
 api_key: admin
 galaxy_instance: http://localhost:8080
 tools:
-  - name: add_value
-    owner: devteam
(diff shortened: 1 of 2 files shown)


bebatut:
@blankclemens and @Stortebecker can you check the content of this PR?

blankclemens:
Thanks a lot, looks like a great tour to me! The pre-converted file from Zenodo `qExactive01819_vendor-peakPicking.mzml` only works with `PeakPickerWavelet` for me, but since this tool is mentioned I one can figure that out.. might be still nice if it works with HiRes.

In the FileConverter step it might be helpful to add that the output format has to be selected in the "Advanced Options" (they put them there in OpenMS 2.2).

Again, thanks a lot!

polkhe:
@blankclemens added comment about the output in FileConverter tool

@Stortebecker changed the file name and the link for the input file in the tour :)

PR #556, opened 2017-10-01
yamllint fixes

shiltemann:


topics/usegalaxy/docker/tools.yaml
```diff
@@ -44,5 +44,5 @@ tools:
     owner: iuc
     tool_panel_section_label: "Convert Formats"
     tool_shed_url: https://toolshed.g2.bx.psu.edu/
-    install_resolver_dependencies: True
-    install_tool_dependencies: False
+    install_resolver_dependencies: "True"
```
bgruening: I think it is easier to just write `true` and `false`.
Thanks Saskia!

shiltemann: fair point, will update

bgruening:
Thanks Saskia!

Issue #546, opened 2017-09-22
Pull request returning error for non existing urls of committed files 


Delphine-L:
Hi,

I report an issue after failed test on unicycler training pull request (https://github.com/galaxyproject/training-material/pull/476). Travis returns errors of non existing external urls (https://travis-ci.org/galaxyproject/training-material/builds/277310665?utm_source=github_status&utm_medium=notification).  The urls at fault are linking to files committed in the pull request (eg: https://github.com/galaxyproject/training-material/tree/master/topics/assembly/tutorials/unicycler-assembly/tutorial.md). The errors don't appear in previous or following commits. Not sure if it is a problem with the test or with how I wrote the links ?

Thanks for the help,

martenson:
The errors:

```
Checking 763 external links...
Ran on 98 files!
- ./_site/topics/assembly/index.html
  *  External link https://galaxyproject.github.io/training-material/topics/assembly/tutorials/unicycler-assembly/slides.html failed: 404 No error
  *  External link https://galaxyproject.github.io/training-material/topics/assembly/tutorials/unicycler-assembly/tutorial.html failed: 404 No error
  *  External link https://github.com/%7B failed: 404 No error
- ./_site/topics/assembly/tutorials/unicycler-assembly/tutorial.html
  *  External link http://www.nature.com/nmeth/journal/v9/n4/full/nmeth.1935.html failed: 401 No error
  *  External link https://galaxyproject.github.io/training-material/topics/assembly/tutorials/unicycler-assembly/slides failed: 404 No error
  *  External link https://github.com/galaxyproject/training-material/tree/master/topics/assembly/tutorials/unicycler-assembly/tutorial.md failed: 404 No error
htmlproofer 3.7.2 | Error:  HTML-Proofer found 6 failures!
```

willdurand:
This is fine. Most of the PRs won't add new links so it is better not to try to change the behavior that is correct 90% of the time.

You did nothing wrong, and if everything in the PR is ok, the PR will get merged 💪☺️

martenson:
@willdurand Maybe we shouldn't run the tests on the PR but only on master then. Failing tests for unknown reasons (to contributors) are discouraging.

willdurand:
@martenson I disagree. It is *really* important to have tests before merging code into master.

bgruening:
> before merging **code** into

[...]

PR #557, opened 2017-10-01
Replace site.url by site.baseurl

bebatut:
To limit errors of htmlproofer, particularly when new content is added (as mentioned in #546)


bebatut:
I am not sure it will really work... I need to check a bit more

bebatut:
I made some extra changes:
- Replace all the `site.url` by `site.baseurl`
- Split the URL into `site.url` (with https://galaxyproject.github.io/) and `site.baseurl` (with "/training-material")
- Update the tests, the pdf generation, ... to fit to that

Now, it should not depend on if sth is already on the main website (specially for the tests during addition of new tutorials) and it makes also easy to deploy on a fork for example (not always referring to the main website)

shiltemann:
looks good @bebatut ..if I understand correctly you will now have to navigate to `localhost:4000/training-material` when running this locally right? could you update tutorial `create-new-tutorial-jekyll` to reflect this?

bebatut:
@shiltemann I updated the tutorials too! Thanks for spotting it!

bebatut:
Thanks @shiltemann

PR #559, opened 2017-10-02
Mothur-MiSeq- SOP: add warning about dataset pair names

shiltemann:
If data is imported from Zenodo and the pair names are left to their defaults in the collection creation step, it will cause problems towards the end of the tutorial, so I've added a warning about this.


bgruening:
Thanks @shiltemann!

PR #560, opened 2017-10-04
Travis build matrix

shiltemann:
We have enabled the option that PRs can only be merged if the status checks pass, but I think this may be a bit too strict when it comes to the url checking. Here I have split the travis checking into 3 separate jobs, `yamllint`, `site building`, and `htmlproofer`, the last of which is an allowed_failure. 

Travis now looks something like [this](https://travis-ci.org/shiltemann/training-material/builds/283139312); even though htmlproofer failed it didnt fail the overall job. Maintainers would still have to be careful to examine the failed urls to see if it is something that needs to be fixed or maybe just an external site that is temporarily down orso, but are not prevented from merging PRs. What do you think?

image

ping @nsoranzo thoughts? or would build stages perhaps be a better way to go?


willdurand:
I think @bebatut's PR #557 should fix the problem better than allowing htmlproofer to fail.

shiltemann:
@willdurand that is true for our internal links, but what if a third-party site is down for a day?

shiltemann:
I actually hoped this would have the three checks more prominently displayed in the merge box below, like it is on galaxy's repo ..maybe a different travis setup can get it like that, so that it is more obvious that it did fail.

shiltemann:
hmm, yes, like this it's just hiding the failure, also not what we want ..perhaps undo the allowed_failure, and disable the requirement that we can only merge if tests are passing, and just instruct maintainers to only merge if the yaml + building jobs pass and use their discretion for the htmlproofer job? dunno.. but just feels a bit restrictive to only be able to update our own site if every single external site we link to is up and running..

willdurand:
I also tend to think that some of you should get admin rights to avoid such situations, without having to ping the Galaxy admins who are not always available or aware of what happens to this repo. I am thinking about people like @bgruening, @bebatut and you @shiltemann. Other maintainers (like me) should not get any rights like this and wait for the green "Merge PR" button.

Makefile
```diff
@@ -30,10 +30,17 @@ build: ## build files but do not run a server
...
+check-html: build ## validate HTML
 	timeout 120s bundle exec htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/" ./_site
-.PHONY: check
+.PHONY: check-html
+
+check-html-gh-pages:  ## validate HTML on gh-pages branch (for daily cron job)
+	timeout 120s bundle exec htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/" .
+.PHONY: check-html-gh-pages
+
+check-yaml:
```
bebatut: Can you add a small help there?

Makefile
```diff
@@ -30,10 +30,17 @@ build: ## build files but do not run a server
 	${JEKYLL} build
 .PHONY: build
 
-check: build ## validate HTML
-	yamllint .
+check-html: build ## validate HTML
 	timeout 120s bundle exec htmlproofer --http-status-ignore 405,999 --url-ignore "/.*localhost.*/","/.*vimeo\.com.*/" --file-ignore "/.*\/files\/.*/" ./_site
```
bebatut: Can we use something else than the timeout? It is not working on OSX

shiltemann: sure thing, I will look into alternatives

shiltemann:
removed the `allowed_failure` again. Added a fourth job, for `vl`, because htmlproofer does not check the markdown-formatted links in the slides.

Makefile
shiltemann: I removed the `timeout` again, because by splitting it into separate jobs, nothing needs to run after this anymore so it should be ok if it just hits the travis timeout when it hangs I think? (@bgruening I assume thats why you added this timeout? or was there another reason?)

bgruening: The timeout was needed because the testing generated no output for 10min, which results in a travis error. If this is not the case anymore it can be removed. But travis fails if 10min no output.

topics/dev/tutorials/architecture/slides.html
```diff
@@ -295,7 +295,7 @@
 
 - Python interface for web servers defined by PEP 333 - https://www.python.org/dev/peps/pep-0333/.
 - Galaxy moving from Paster to uwsgi to host the application.
-  - http://pythonpaste.org/
+  - [pythonpaste.org](https://en.wikipedia.org/wiki/Python_Paste)
```
nsoranzo: Or https://bitbucket.org/ianb/paste

shiltemann: ah, better, thanks!

.travis.yml
```diff
@@ -40,19 +47,26 @@ script:
...
         make build
-      else
-        make check
       fi
     fi
 
 after_success:
   - |
-    if [ "$TRAVIS_PULL_REQUEST" == "false" -a "$TRAVIS_BRANCH" == "master" ]
+    if [ "$TRAVIS_PULL_REQUEST" == "false" -a "$TRAVIS_BRANCH" == "master" -a "$CHECK" == "site-build" ]
```
nsoranzo: It seems to me that this means that a commit on `master` or a merged PR will always publish the website even if the some checks failed.

shiltemann: I added this because I thought otherwise each of the 4 jobs would try to do the publishing ..we don't want to skip publishing if links are broken I think (sometimes, like now, a third-party site may be down, and we still merge things and don't want this to stop updating of our own site) ..I figured as long as the site building job succeeded it was safe to update the site?

nsoranzo: If you want to publish only if the 3 check jobs pass, you can use build stages.
If you want to publish even if check jobs fail, then this is fine, but it's risky, you can publish a completely broken site.

bebatut: If it is totally broken, it will failed in `bin/publish` where it is checked that at least `index.html` is there

shiltemann: @nsoranzo but don't we know that the site building itself at least succeeded because I'm having the job that ran `make build` do the publishing with an `after_success` ..or am I misunderstanding?

nsoranzo: Maybe I exaggerated, but you may surely push broken yaml files, no?
And HTML with broken internal and external links and hrefs.

shiltemann: that is true (but if they are very broken it would fail the build too). But I could also add yaml checking to the `make build` command as well, to make sure both are in order before publishing?

shiltemann: yep @ broken links, but for instance the site for quast has been down for the past two days orso now (but probably a temporary problem), and I don't know if we want to prevent ourselves from merging PRs and updating the site at such times (but this can be discussed) 

I do see that htmlproofer has options `--disable-external` and `--check-external` to check only internal or external links respectively, so we could consider adjusting things to at least require internal links to be working perhaps too

PR #562, opened 2017-10-05
added tour for srna

polkhe:


bebatut:
Thanks @polkhe 
@malloryfreeberg can you check it? Thanks

mvdbeek:
I'm a little surprised to find piRNAs in a non-gonadal, somatic cell line. The only cell lines so far that I'm aware of that express a functional piRNA machinery in Drosophila are the OSS and OSC cell lines. I'm assuming that Dmel-2 here means S2 cells ?
 
In any case it is not sufficient to call small RNA molecules piRNAs because their length is between 23 and 29 nucleotides. Subsequently I have frequently seen some minor amounts of small RNAs in this size range, but when aligned to transposons they are most aligned to the sense strand, suggesting that they are degradation products. While the last part is speculation on my side, I think we'd need to have more proof than that to call them piRNAs.

bgruening:
@mvdbeek good points. But this is unrelated to this PR I think and should be directed to the entire small RNA training. ping @malloryfreeberg

mvdbeek:
It is directly related to this PR, which suggests piRNAs can be found in this somatic cell line, when there I no literature on this. I haven't read the remaining tutorials yet, but this PR in it's current form is not scientifically sound. I can correct this next week (a comparison with ovarian data should be pretty self-explanatory).

malloryfreeberg:
Although this PR is for the added sRNA-seq tour, I can address some of the comments here. A few points to keep in mind related to the piRNA discussion:

1. The downstream focus of the tutorial (as it is now) is on identifying differentially expressed **siRNAs** in the samples (yes, Dmel-2 are embryonic, cultured S2 cells), not piRNAs. This is the first sRNA-seq tutorial in the GTN, but by no means needs to be the only one. 

2. I mention piRNAs in the tutorial to keep it more general and applicable to systems that **do** express piRNAs and to illustrate to users that small RNA subclasses in a variety of systems can be explored with the given tools.

3. These sRNA-seq datasets were intentionally downsampled to have both 20-22nt and 23-29nt reads, and thus the tutorial data should not be used to draw conclusions about the study it references. In fact, 23-29nt length reads made up a small proportion of the entire dataset for these samples. The theme of whether/how much to downsample as downsampled data can give vastly different results than analyzing the whole dataset has always been an issue. I'm happy to discuss approaches for downsampling that avoid any confusions or may be misleading.

4. I completely agree that size-based classification alone is not sufficient to 100% correctly categorize small RNAs. The better way is to classify small RNAs by their protein binding partners which involves analyzing a different type of sequencing (CLIP-seq/HITS-CLIP/etc.) for which I hope tutorials will be added. For sRNA-seq datasets, however, I am happy to discuss alternative workflows that could be suggested, such as aligning to transposons first and then analyzing length distribution. Whichever workflow we choose, classifications based on the length of the small RNA **is** a very common approach, especially in e.g. *C. elegans* which have specific classes of small RNAs that are 21nt, 22nt, and 26nt. Illustrating to users how they can use Galaxy to segregate their reads by length is a vital feature that is missing from all other tutorials, and is thus a key feature of the sRNA-seq tutorial(s).

Looking forward to continuing this discussion as it will benefit the usefulness of the tutorial in the end :)

mvdbeek:
> Whichever workflow we choose, classifications based on the length of the small RNA is a very common approach, especially in e.g. C. elegans which have specific classes of small RNAs that are 21nt, 22nt, and 26nt. Illustrating to users how they can use Galaxy to segregate their reads by length is a vital feature that is missing from all other tutorials, and is thus a key feature of the sRNA-seq tutorial(s).

I totally agree, but then why not use a dataset that actually expresses piRNAs ? There's plenty of ovarian or OSC data out there, if it has to be Drosophila. If c.elegans has a more distinct size profile that may also be an alternative. Yet another alternative is to specifically mention that the profile looks nothing like a real piRNA profile.

topics/transcriptomics/tutorials/srna/tours/tour.yaml
```diff
@@ -0,0 +1,771 @@
...
+      the Illumina Universal Adaptors, we will align our trimmed reads to
+      reference Drosophila rRNA and miRNA sequences (dm3) to remove these
+      artifacts. Interestingly, Drosophila have a short 2S rRNA sequence that is
+      30nt long and typically co-migrates with sRNA populations during gel
+      electrophoresis. rRNAs make up a very large proportion of all non-coding
+      RNAs, and thus need to be removed. Oftentimes, experimental approaches can
+      be utilized to deplete or avoid capture of rRNAs, but these methods are
+      not always 100% efficient. We also want to remove any miRNA sequences, as
+      these are not relevant to our analysis. After removing rRNA and miRNA
+      reads, we will analyze the remaining reads as siRNA and piRNA sequences.
```
mvdbeek: If you don't want to change the datasets you could change this to
"we will analyze the remaining reads, which may be siRNA or piRNA sequences."

topics/transcriptomics/tutorials/srna/tours/tour.yaml
```diff
@@ -0,0 +1,771 @@
...
+      Rename each resulting dataset collection something meaningful (i.e. “blank
+      RNAi - siRNA reads (20-22nt)”)
+    backdrop: true
+  - title: Extract subclasses
+    content: >-
+      The regular expression in the <b>Match by</b> parameter tells the tool to
+      identify sequences that are length 12-19 or 23-50 (inclusive), and the
+      <b>Miscellaneous Manipulation Type</b> parameter tells the tool to remove
+      these sequences. What remains are sequences of length 20-22nt. We will now
+      repeat these steps to identify 23-29nt piRNA sequences.
```
mvdbeek: "to identify sequences in the size-range of piRNAs (23-29nt)."

topics/transcriptomics/tutorials/srna/tours/tour.yaml
```diff
@@ -0,0 +1,771 @@
...
+    content: Click on the "Manipulate FASTQ" tool to open it
+    placement: right
+    postclick:
+      - >-
+        a[href$="/tool_runner?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fdevteam%2Ffastq_manipulation%2Ffastq_manipulation%2F1.0.1"]
+  - title: Extract subclasses
+    element: '#tool-search'
+    content: >-
+      Execute the tool on each collection of non-r/miRNA reads to identify
+      piRNAs (23-29nt) using the following parameters <ul> <li><b>FASTQ
```
mvdbeek: "to identify sequences in the size-range of piRNAs (23-29nt) using ..."

topics/transcriptomics/tutorials/srna/tours/tour.yaml
```diff
@@ -0,0 +1,771 @@
...
+    position: left
+  - title: Extract subclasses
+    element: '.history-right-panel .list-items > *:first'
+    content: >-
+      Inspect the outputs. You should notice that we have ~32k reads that are
+      20-22nt long in our Reads_20-22_nt_length_siRNAs file and ~16k reads that
+      are 23-29nt long in our Reads_23-29_nt_length_piRNAs file, as expected.
+      Given that we had ~268k trimmed reads in the Blank RNAi replicate 3 file,
+      siRNAs represent ~12.1% of the reads and piRNAs represent ~6.2% of the
+      reads in this experiment.
```
mvdbeek: Here we could mention that there is no distinct peak in the piRNA range and refer to a publication where the reader can compare this to what a real piRNA profile looks like.

bebatut:
@polkhe Can you make the changes suggested by @mvdbeek ?
You will also probably need to rebase and add the tour metadata.

topics/transcriptomics/tutorials/srna/tours/tour.yaml
bebatut: @mvdbeek do you have an example of such publication?

mvdbeek: http://dx.doi.org/10.1016/j.cell.2009.03.040
Fig S1 is nice in that regard. Surprisingly few publications show that actually full size-distribution.

shiltemann:
Thanks for the valuable discussion @mvdbeek and @malloryfreeberg. I have added the same edits suggested by @mvdbeek also to the tutorial text. If any further changes are needed, please open an issue and we can further update both the tutorial and tour.

shiltemann:
thanks for the tour @polkhe!

PR #565, opened 2017-10-09
added tour for De novo transcriptome reconstruction with RNA-seq

polkhe:


bebatut:
Thanks @polkhe 
@malloryfreeberg can you check the tour? Thanks a lot !

bebatut:
I forgot to update the metadata

Issue #552, opened 2017-09-27
update tutorials using HISAT2 to new version


mblue9:
Creating this issue to flag that, as HISAT2 has been updated, the current screenshots and text in some GTN tutorials no longer match the newest version of the tool, so will need to be updated at some stage e.g. these tuts

- https://galaxyproject.github.io/training-material/topics/transcriptomics/tutorials/ref-based/tutorial.html
- https://galaxyproject.github.io/training-material/topics/transcriptomics/tutorials/de-novo/tutorial.html
- https://galaxyproject.github.io/training-material/topics/transcriptomics/tutorials/srna/tutorial.html

Happy to help update if you like when I get a chance.

bgruening:
I guess this is a good example why we should avoid including screenshots of tools. Any opinions about this?
Thanks @mblue9 for tracking the issue.

bernt-matthias:
I just stumbled over the same issue. Removing the screenshots would not be sufficient, since also parameter names or the available parameters change. I would suggest to keep the screenshots, but to include the version of the used tools in the tolls.yml file (and maybe in the tutorial text?).

bgruening:
The revision will be included as soon as we auto generate the tools.yml file from the workflow files. But I still think the parameter names are way more stable then images.

mblue9:
It's easier for maintainability to remove the images but I think it's easier for users if there are images in the tutorials (I've had a user asking if the material could have _more_ screenshots). I have no idea if this is possible but the ideal solution imho would be if the tutorials could be automatically run (Selenium-style?) to test that they work and auto-generate the screenshots.

bernt-matthias:
on the other hand there are the tours

bebatut:
I have the feeling that if the parameters are clearly described (full name of the parameters, clear value, parameters in the same order as they appear in the Galaxy interface) the screenshot are not really needed (some time bad quality and more difficult to update):

![screen shot 2017-09-28 at 09 50 34](https://user-images.githubusercontent.com/1842467/30955024-865881ee-a432-11e7-9db7-1f306358eb60.png)

At least, we can also add comment about the parameters directly and it is easier to update and less issue also with the name of the input 😄 
We could think about a good way to describe the parameters (other issue?) that we try to have in all the tutorials, except the first steps in the introductions (where I can understand screenshots are better to familiarize yourself with the interface)

shiltemann:
[...]

Issue #544, opened 2017-09-20
DEXSeq step error in Ref-based RNA-seq


mfernandes61:
Hi,
I'm getting an error with the procedure towards the end that uses DEXSeq.
It has occurred on my GVL instance and on the usegalaxy public server
It is possible that I have made a mistake when entering factors but I have tried several times and got
the same result (Output below)

Any insights/answers?
Thanks,
Mark

Dataset generation errors

Dataset 9: DEXSeq result file on data 8, data 5, and others

Tool execution generated the following error message:
Fatal error: An undefined error occured, please check your input carefully and contact your administrator.
Loading required package: BiocParallel
Loading required package: methods
Loading required package: Biobase
Loading required package: BiocGenerics
Loading required package: parallel

Attaching package: ‘BiocGenerics’

The following objects are masked from ‘package:parallel’:

    clusterApply, clusterApplyLB, clusterCall, clusterEvalQ,
    clusterExport, clusterMap, parApply, parCapply, parLapply,
    parLapplyLB, parRapply, parSapply, parSapplyLB

The following objects are masked from ‘package:stats’:

    IQR, mad, xtabs

The following objects are masked from ‘package:base’:

    anyDuplicated, append, as.data.frame, cbind, colnames, do.call,
    duplicated, eval, evalq, Filter, Find, get, grep, grepl, intersect,
    is.unsorted, lapply, lengths, Map, mapply, match, mget, order,
    paste, pmax, pmax.int, pmin, pmin.int, Position, rank, rbind,
    Reduce, rownames, sapply, setdiff, sort, table, tapply, union,
    unique, unsplit, which, which.max, which.min

Welcome to Bioconductor

    Vignettes contain introductory material; view with
    'browseVignettes()'. To cite Bioconductor, see
    'citation("Biobase")', and for packages 'citation("pkgname")'.

Loading required package: SummarizedExperiment
Loading required package: GenomicRanges
Loading required package: stats4
Loading required package: S4Vectors

Attaching package: ‘S4Vectors’

The following objects are masked from ‘package:base’:

    colMeans, colSums, expand.grid, rowMeans, rowSums

Loading required package: IRanges
Loading required package: GenomeInfoDb

Attaching package: ‘SummarizedExperiment’

The following objects are masked from ‘package:IRanges’:

    acbind, arbind

Loading required package: DESeq2
Loading required package: AnnotationDbi
Loading required package: RColorBrewer
Warning messages:
[...]

Issue #531, opened 2017-09-18
replace DAVID tool with an alternative


mfernandes61:
Hi,
Trying to run the ref-based RNA-seq in a GVL instance.
Can't find DAVID tool (Mentioned on page 21) in tool-shed - what should I be looking for?

TIA,
Mark

shiltemann:
The tools seems to refer to a non-toolshed tool installed on usegalaxy.org: https://usegalaxy.org/?tool_id=hgv_david&version=1.0.1

Perhaps somebody could share the tool xml and we can see if we can turn it into a toolshed tool? (ping @nekrut)

yvanlebras:
Hi,
After exchanging with @bgruening and @nsoranzo because I was searching for a kind of "DAVID" tool, they mention others already Galaxy integrated tools like gprofiler : https://toolshed.g2.bx.psu.edu/view/tambeta/gprofiler/dfda180ac951  / Kobas https://toolshed.g2.bx.psu.edu/view/iuc/kobas/83b998fa34ea / Goseq https://toolshed.g2.bx.psu.edu/view/iuc/goseq/9d1256d9ef0b . I think, once we have tested these 3 tools, we have to modify the DAVID mention by one (or more ?) of these tools on the tutorial...

mvdbeek:
In any case DAVID or any other enrichment test should **absolutely not** be used with RNA-seq data without some sort of length correction. The [goseq vignette](https://www.bioconductor.org/packages/devel/bioc/vignettes/goseq/inst/doc/goseq.pdf) and paper
```
Young MD, Wakefield MJ, Smyth GK and Oshlack A (2010). “Gene ontology analysis for RNA-seq: accounting for selection bias.” Genome Biology, 11, pp. R14.
```
go into details about this (I have read other papers that reach similar conclusions, but I don't have them at hand now). Basically DAVID was developed for microarray data, where there appears to be no length bias.

Issue #519, opened 2017-09-11
ref-based tutorial for RNA-seq what is the tool "Infer Experiment"


yvanlebras:
In the RNA-seq ref-based tuto.md https://github.com/galaxyproject/training-material/blob/master/topics/transcriptomics/tutorials/ref-based/tutorial.md  , I saw the mention of a tool, named "Infer Experiment". What is this tool ? I don't find it on the main TS....

bebatut:
It is part of rseqc tool: https://toolshed.g2.bx.psu.edu/view/nilesh/rseqc

bebatut:
Did you find it? Can I close the issue?

yvanlebras:
yes thanks! Maybe this can be of interest to update the tuto...

bebatut:
What do you mean? The tool is called "Infer Experiment" when searching it on Galaxy. On the ToolShed the rseqc tool install many different tools

PR #571, opened 2017-10-13
Improving the reference based Transcriptomics tutorial

bernt-matthias:
I'm currently working on this tutorial. Lets see if we an make some progress here:

https://github.com/galaxyproject/training-material/issues/519 should be solved by my first commit. 

I discovered the following issues where comments would be welcome: 

https://github.com/galaxyproject/training-material/issues/552
https://github.com/galaxyproject/training-material/issues/544
https://github.com/galaxyproject/training-material/issues/531

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ pdf: install-pdf detached-serve ## generate the PDF of the tutorials and slides
             --print-to-pdf="$$name" \
             "$(SITE_URL)/$$t" \
             2> /dev/null ; \
-    done
+	done
 	@for s in $(SLIDES); do \
 		name="$(PDF_DIR)/$$(echo $$s | tr '/' '-' | sed -e 's/html/pdf/' -e 's/topics-//' -e 's/tutorials-//')"; \
 		$(DECKTAPE_DIR)/phantomjs \
--- a/topics/transcriptomics/tutorials/ref-based/tools.yaml
+++ b/topics/transcriptomics/tutorials/ref-based/tools.yaml
@@ -14,9 +14,9 @@ tools:
     owner: iuc
     tool_panel_section_label: 'NGS: RNA Analysis'
 
-  - name: stringtie
-    owner: iuc
-    tool_panel_section_label: 'NGS: RNA Analysis'
+#   - name: stringtie
+#     owner: iuc
+#     tool_panel_section_label: 'NGS: RNA Analysis'
 
   - name: featurecounts
     owner: iuc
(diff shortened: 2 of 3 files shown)


nsoranzo:
It was also proposed to use featureCounts instead of htseq-count.

bernt-matthias:
I see: as in the de-novo tutorial. What would be the pro/cons? 

Somehow I like the idea to cover more than one tool in the different tutorials (having in mind that the tutorials serve also as human operated Galaxy/Tool tests).

bernt-matthias:
I also was wondering what the `Filter` tool is?

nsoranzo:
> What would be the pro/cons?
> Somehow I like the idea to cover more than one tool in the different tutorials (having in mind that the tutorials serve also as human operated Galaxy/Tool tests).

featureCounts was much faster with equivalent results, don't know with the latest versions. A comparison is at http://bioinformatics.cvr.ac.uk/blog/featurecounts-or-htseq-count/

> I also was wondering what the Filter tool is?

I suppose the one shipped with Galaxy with tool id `Filter1` and full name/description "Filter data on any column using simple expressions".

bgruening:
My preferred tutorial would be. Star for mapping, featureCount for counting, replacing David with potentially goseq.

bernt-matthias:
One question regarding data collections. I tried to process all paired samples as list of pairs and the others as list, which works fine until DeSeq2. The problem is that it seems impossible to select the data sets corresponding to the factors within such a list. Also breaking the lists (btw is there a better way than unhiding the actual data sets) seems not helpful, since the automatic naming does not allow to identify the factors anymore. Also tags seem to be no option since they are lost when trim galore is run on such a list -- and factor selection by tag in a list seems not possible .. (I guess tags are still in development).

nsoranzo:
Maybe @mvdbeek can help with the tag question here?

mvdbeek:
> The problem is that it seems impossible to select the data sets corresponding to the factors within such a list.

Right, what I do is creating a collection per factor (you can do this after the fact with "Filter List from contents of a file"). This will keep the element identifiers.

> factor selection by tag in a list seems not possible .. (I guess tags are still in development).

It is in some input boxes, but not in the multi-input list at the moment.

bernt-matthias:
I finally got goseq running (maybe you have some comments): 
- I obtained go data with `BioMart Ensembl server` 
- Gene length data was obtained with Gene `length and gc content` (using BioMart did not work since the Transcript length (including UTRs and CDS) results in a table having duplicate gene entries)
- Differentially expressed gene file was obtained with `Compute an expression on every row`  and `cut`

mvdbeek:
@bernt-matthias that looks right to me. @mblue9 did some work on goseq in https://github.com/galaxyproject/tools-iuc/pull/1523 to be able to fetch go categories in goseq, so you could skip the first step if the annotations for your species are among human, mouse, fruitfly or zebrafish.

bernt-matthias:
The PR looks great. I will check if the new version is in the TS when I start writing.

bebatut:
@bernt-matthias I will close this PR. We updated the tutorial with #691 and changed the `tour.yaml` file
Let me know if I miss a change that should be added to the tutorial

PR #572, opened 2017-10-13
docker: switch to galaxy user for install

shiltemann:
run the installation of tutorials as user `galaxy` to prevent problems when some files in conda are owned by root (as mentioned by @nekrut on Gitter)

### DIFF
--- a/bin/docker-install-tutorials.sh
+++ b/bin/docker-install-tutorials.sh
@@ -10,6 +10,9 @@ startup_lite
 # wait until galaxy has started
 galaxy-wait -g $galaxy_instance
 
+# run tutorial install as user galaxy
+su - galaxy
+
 # install other tutorial materials
 for dir in /tutorials/*
 do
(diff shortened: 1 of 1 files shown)


bin/docker-install-tutorials.sh
```diff
@@ -10,6 +10,9 @@ startup_lite
 # wait until galaxy has started
 galaxy-wait -g $galaxy_instance
 
+# run tutorial install as user galaxy
+su - galaxy
```
nsoranzo: Can we reuse the `$GALAXY_USER` environment variable from https://github.com/bgruening/docker-galaxy-stable/blob/master/galaxy/Dockerfile#L29 ?

shiltemann: good call, definitely.

bgruening:
This is what we do in flavors: https://github.com/galaxyproject/ansible-galaxy-extras/blob/master/templates/install_repo_wrapper.sh.j2#L16
So its similar to this patch.

Thanks @shiltemann!

PR #573, opened 2017-10-14
chip-seq: fix link in tutorial

shiltemann:
thanks for reporting @AisanjiangW


bebatut:
Thanks @shiltemann for fixing the link

PR #574, opened 2017-10-14
contributors: update changed github username

shiltemann:
@AisanjiangW changed their github username (used to be ewran)


bebatut:
Thanks @shiltemann

PR #576, opened 2017-10-16
Update tutorial template

shiltemann:
with an example of a tool configuration description, to encourage a uniform look and feel of these hands-on parts

ping @bebatut


bebatut:
It looks good for me! Thanks @shiltemann

PR #578, opened 2017-10-17
workflow for seq-an-mapping

polkhe:
added  workflow for the topic of mapping


bebatut:
@polkhe Looks good! Thanks!

PR #587, opened 2017-10-18
Annotation of Galaxy instances able to run a tutorial

bebatut:
Hi,

This PR should enable the annotation of which Galaxy instance can be used for a tutorial (meaning that the tools in the `tools.yaml` file are installed on this instance).
How is it working:
- Generation of a `instances.yaml` (in `metadata` dir):
   - Take the list of public server from Martin (https://github.com/martenson/public-galaxy-servers)
   - Extract the tools installed there
   - Parse all the tutorials and check on which instances are available the tools in `tools.yaml`
   - Extract this information to `instances.yaml`
- Use this information and add it to the website
   - Add on the top of the tutorial with links to the instances and with Docker
       screen shot 2017-10-18 at 17 29 48
   - Add also this information in the topic page
       screen shot 2017-10-18 at 17 29 03

It is not looking really nice (@willdurand I need your magic there 😄 )

Bérénice


bgruening:
@bebatut have you got my mail last week with a similar implementation?

bebatut:
@bgruening I saw them. Thanks: I used some parts in this implementation to improve my fist implementation. I was just not sure how and where to use the badges and if they were really useful there (or I missed the idea). We can still integrate them. I wanted a first Proof-of-Concept of my first idea and find how to implement it 😄

bgruening:
I would like to avoid the complexity in this repository. 
My idea was more like a very small web service that serves the badges, maybe in combination with https://grafana.denbi.uni-freiburg.de/dashboard/db/public-galaxy-servers. This way we keep complexity out here and serve badges that every university/contributor can use on there own website. We can then simply include a link here to the badges, list of supported servers etc... A lot of freedom without messing with the actual training material.

bebatut:
Then it is complementary to that. Do not you think? We have the `instances.yaml` that we could update regularly and based on that we could have outside of this repo a service such as the one you described

bebatut:
I think having the links to the Galaxy instances directly in the website (on the tutorial and on the topic level) are useful (for someone who want to try the tutorial)

bgruening:
The badges are the links imho and the instances.yaml is not needed if we simply rely on https://github.com/martenson/public-galaxy-servers and point people to add the servers there. I would like to avoid further duplication.

bebatut:
The `instances.yaml` is based on https://github.com/martenson/public-galaxy-servers. I used this list to check if the tools required for a tutorial are on each of the public galaxy servers on the list. It is not an other list of servers but for each tutorial which Galaxy instance can be used. The one redundant is the `galaxy_instances.yaml` that should be removed.
Where do you want to put the badges? 
I am not sure to really understand your idea

erasche:
before anyone spends too much time writing "very small web service that serves the badges", our badge-serving code in public-galaxy-servers is wayyyy simpler than a small web service.

- cron job triggers execution
- I generate all the pngs ([download + cache copies from shields.io](https://github.com/martenson/public-galaxy-servers/blob/master/badges.py#L85))
- and [copy them](https://github.com/martenson/public-galaxy-servers/blob/master/badges.py#L95) into a directory
- apache (which is already serving grafana) serves them.

willdurand:
Dropdown fixed.

willdurand:
(and rebased)

shiltemann:
thanks @bebatut @erasche @willdurand !

PR #589, opened 2017-10-24
Badge support for training-compatible instances

erasche:


erasche:
@bebatut in order to now have to parse external data sources I needed to annotate *all* instances, including unsupported ones. This makes the metadata file much larger, but means that 100% of the necessary information is in one place.

erasche:
Ok @bebatut finally this should be "ok".

1. no duplication with your work
2. (thus) no network requests needed other than fetching badges.
3. can be run separately from your script. I keep going back on whether or not this is the "best" way, but this way at least we have the option to run it on a different machine / in a different deployment way.

erasche:
auswahl_059

erasche:
ping @bebatut ready.

erasche:
rebased as well https://github.com/galaxyproject/training-material/pull/587#issuecomment-348539842

willdurand:
😎

PR #591, opened 2017-11-06
tour for protein id with oms

polkhe:
Added tour for proteomics-protein id with oms

### DIFF
--- a/topics/proteomics/tutorials/protein-id-oms/tours/tour.yaml
+++ b/topics/proteomics/tutorials/protein-id-oms/tours/tour.yaml
@@ -135,7 +135,9 @@ steps:
     position: right
   - title: Uploading the protein database
     element: '#tool-panel-upload-button .fa.fa-upload'
-    content: For spectrum-to-sequence matching, we need to upload a protein database. Open the Galaxy Upload Manager.
+    content: >-
+      For spectrum-to-sequence matching, we need to upload a protein database.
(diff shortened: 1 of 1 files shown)


bebatut:
@Stortebecker and @blankclemens, could you check this tour? Thanks

topics/proteomics/tutorials/protein-id-oms/tours/tour.yaml
```diff
@@ -0,0 +1,590 @@
...
+    postclick:
+      - '#tool-panel-upload-button .fa.fa-upload'
+      - '#btn-reset'
+  - title: Uploading the input data
+    element: '#btn-new'
+    content: Click on Paste/Fetch Data
+    placement: right
+    postclick:
+      - '#btn-new'
+  - title: Uploading the input data
```
blankclemens: Would it be possible to provide an already converted file? `msconvert` needs to run on Windows to support raw formats - which is probably not set up on most Galaxy instances (and especially not supported in docker flavors which could benefit from including this tutorial). But we should mention this hiccup to make users aware on how to handle their own data.

Stortebecker: A converted file is provided alongside the `raw` in Zenodo. We should definitely use the `mzML` file and skip the conversion! I can do the necessary changes tomorrow.

PR #597, opened 2017-11-14
Updated tutorial for protein ID with OpenMS

Stortebecker:
- Added a section for using multiple search engines
- updated workflows
- minor fixes


bgruening:
Thanks Flo!

PR #604, opened 2017-11-27
Update RNA-based tutorial tool

afgane:
`trimgalore` (particularly from `iuc`) does not exist in the Tool Shed so update to an existing tool.


shiltemann:
thanks @afgane :)

PR #607, opened 2017-12-01
Update decktape installation

bebatut:
- Add npm as dependency
- Use npm to install decktape
- Update decktape command to create PDF for the slides


shiltemann:
thanks @bebatut

willdurand:
Why not using a docker image?

bebatut:
What do you mean?

willdurand:
Why not using a docker image for decktape?

PR #610, opened 2017-12-01
workflow for ref-based-rad-seq tutorial

polkhe:
added workflow for ref-based-rad-seq tutorial


bebatut:
@yvanlebras Can you check this workflow?

Issue #280, opened 2017-04-17
RNAseq Tutorial_Some Broken linkages


afalaki:
Hi,
I am very new to this repository. I was running the RNA-Seq training by @Bérénice Batut and @MoHeydarian when I noticed a couple of broken linkages. 
- Galaxy instance: when I clicked to see about a flavored Galaxy instance available on http://localhost:8080
- The link to Adam Roberts et al: Improving RNA-Seq expression estimates by correcting for fragment bias

Also, there were some minor issues regarding the Reference-based RNA-seq data analysis (https://galaxyproject.github.io/training-material//RNA-Seq/tutorials/ref_based):
-Is it possible to please to add a comment or hyperlink on how to run the infer experiment in Galaxy? Shall I use a local Galaxy to be able to install this package?

-Requirements link to the IGV is broken
-In data upload, there is a misspelling for GSM461181 (in the hyperlink, the dataset is linked as GSM4611810, so it is not possible to find the data by just clicking on it)
-The link to jump to the Analysis of the differential expression in the Agenda does not work
-In the Analysis of the differential gene expression, you asked for p-value equal to or below 0.05, but later in the answer p-value less than 0.05 is used. This will change the number of significant genes.
- Although I followed your tutorial, the number of significant up or downregulated genes that I got is slightly different from the tutorial that I did not understand the reason.

Thanks for your consideration,
Ali

bgruening:
@afalaki this is awesome feedback, thanks a bunch!
We will fix the broken links as soon as possible. Regarding the link to localhost:8080, this will only work if you are running a Galaxy Docker instance on your computer. We are aiming in providing a Galaxy instance for every training for everyone. So you need to install this locally (which is hopefully very easy.)

bebatut:
@afalaki Thanks

1. The link to Adam Roberts et al has been fixed
2. Requirements link to the IGV has been removed
3. We are fixing the link for GSM461181
4. The link to jump to the Analysis of the differential expression has been fixed
5. The inconsistency for the p-values has been fixed

For the different results, I do not how to explain that 😟

PR #611, opened 2017-12-01
Fix link to GEO for one sample in RNA-Seq Ref based tutorial

bebatut:


shiltemann:
nice catch

bebatut:
It was catched by @afalaki in #280

PR #618, opened 2017-12-01
add anchors to the topic page

bgruening:


willdurand:
Hmm the anchors are already there actually, with the `id` attribute set on the titles.

PR #623, opened 2017-12-02
Quote URLs to avoid formatting (e.g. italics).

jxtx:


bebatut:
This formatting has broken the boxes:
screen shot 2017-12-03 at 11 50 21
I open a PR to fix that #625

PR #627, opened 2017-12-03
enable docker builds

bgruening:
This PR will enable Docker builds for all topics.

The big elephant in the room is the size of the Docker images. A few tutorials are simply to big to bundle everything in one image.

### data managers

With CVMFS support inside the images we don't need to strictly run the data_managers anymore, which helps a lot, so the current PR will not run data_managers anymore. For the time being I think we should clean-up data_managers.yaml and remove all reference data that is in the CVMFS. Then only bundle the data_managers into the container that are not yet in CVMFS.

### tools

We are working on a way to distribute also tool dependencies via CMVFS,  just to be prepared for the future currently it seems to work.
I needed to remove a few tools (GATK based) that are not updated since years and had no conda package.

The following should now be the default for all tool.yaml files. The templates were updated accordingly.
```
install_resolver_dependencies: true
install_tool_dependencies: false
```

### data libraries

Most of your data_libraries.yaml files are just paceholders, which is odd and should be fixed in a follow-up PR. However, not all topics do need data, so we might want to remove this from the tests.
I hacked up a small script to merged all data_libraries.yaml files from one topic, so that we avoid redundancy and deliver a clean data library to our users.

Images that at least compile atm:

* https://quay.io/repository/galaxy/epigenetics-training
* https://quay.io/repository/galaxy/introduction-training
* https://quay.io/repository/galaxy/metagenomics-training
* https://quay.io/repository/galaxy/sequence-analysis-training
* [ ] https://quay.io/repository/galaxy/proteomics-training (has a deactivated broken workflow)

To debug:

* [ ] https://quay.io/repository/galaxy/transcriptomics-training (workflow broken?)
* [ ] https://quay.io/repository/galaxy/variant-analysis-training
* [ ] https://quay.io/repository/galaxy/chip-seq-training

No image needed:

* dev
* train the trainer


bgruening:
@bebatut @shiltemann its at least a start. 3 containers does not build yet for several reasons. e.g. broken workflows. However, it will add a bunch of containers to the other topics so its better than master I think.

I will try to work on this over Xmas.

PR #628, opened 2017-12-03
Fix path to tours in the topic page and update tour metadata

bebatut:
I linked the tours to the tours on GitHub

### DIFF
--- a/topics/assembly/metadata.yaml
+++ b/topics/assembly/metadata.yaml
@@ -27,7 +27,7 @@ material:
     type: "tutorial"
     name: "general-introduction"
     zenodo_link: "https://doi.org/10.5281/zenodo.582600"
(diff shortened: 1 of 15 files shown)


topics/assembly/metadata.yaml
```diff
@@ -27,7 +27,7 @@ material:
     type: "tutorial"
     name: "general-introduction"
     zenodo_link: "https://doi.org/10.5281/zenodo.582600"
-    galaxy_tour: ""
+    galaxy_tour: "yes"
```
willdurand: If you use `yes` (or `no`) without the double quotes, it becomes a boolean. This allows:

- to be slightly more standard/interoperable
- to use conditions in the templates (`{% if tutorial.galaxy_tour %}`)

PR #629, opened 2017-12-04
Use real booleans for "hands-on" and "slides" in metadata

bebatut:
I replaced the quoted yes/no by non quoted yes/no, such that they can be used as booleans


bebatut:
I checked it 😄

PR #630, opened 2017-12-04
whitelist vimeo in awesomebot

shiltemann:
like we do in htmlproofer. these links are failing the slide checking jobs currently. ping @bebatut


shiltemann:
@willdurand yes, but downside to that was that then it will always look green on page here even when there are truly broken links :/ ..not sure what is better

..but it was a temporary problem with a couple site, and it's green again now after a restart of the job :)

bebatut:
We can maybe see how it is going in the next days with at least this patch

PR #632, opened 2017-12-05
Add workflows for ref based RNA Seq tutorial

bebatut:


martenson:
trying to trigger things

PR #635, opened 2017-12-05
Add an emoji to the homepage

willdurand:
We are testing Heroku Preview Apps :)

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -20,7 +20,7 @@ layout: base
     <section>
         <h1>Welcome to {{ site.title }}</h1>
         <p class="lead">
-            Collection of tutorials developed and maintained by the worldwide Galaxy community.
+            Collection of tutorials developed and maintained by the worldwide Galaxy community. 😍
         </p>
 
         <div class="row">


willdurand:
closing as this was an example.

afgane:
Does a PR need to be open for this preview page to work? I just stumbled here and don't quite see what's supposed to be happening here :/

bgruening:
The preview website will be removed after a day. You are unfortunately too late :)

afgane:
Thanks for clarifying that. I wasn't clear if I was missing something.

It's like a trigger - there's now a deadline for reviewing PRs :)

bgruening:
Yes, review fast - spreading knowledge can not wait ;)
I don't think this isn't a big problem, this feature is more intended for people who want to check the changes without building it locally. Core contributor can probably build it locally at any time for the review.

So more for a preview than a review imho ;)

willdurand:
"after a day" once the PR is merged or closed. otherwise, the site is up as long as the PR is open.


The only new thing I noticed is that PRs from forks don't create new websites automatically (for security/billing reasons apparently). @martenson can create them on the Heroku dashboard though.

martenson:
We can also change the 1-day to something different...

PR #637, opened 2017-12-06
workflow for rna-seq-viz-with-cummerbund

polkhe:
added workflow and metadata


bebatut:
@bagnacan can you have a look at this worklow? Thanks a lot

bagnacan:
@bebatut, once the input dataset is loaded, and the workflow is imported, all steps run smoothly, and the result is the desired one. Thanks @polkhe!

bebatut:
Thanks @bagnacan for checking! I merge it then

PR #639, opened 2017-12-08
small correction in ref-based tutorial

polkhe:
'condition' should be in lower case as stated in next paragraph


bebatut:
Thanks @polkhe

PR #644, opened 2017-12-20
Typo fix and clarified an input as a collection

jennaj:
- Section "Hands-on: Subsampling": count.groups input clarified as a collection in final "check" step
 - Section "Hands-on: Correlation": Fix tool name from Get.communitype to Get.communitytype in last hands-on step


bgruening:
Thanks @jennaj!

Issue #599, opened 2017-11-17
best location for a mapping-by-sequencing tutorial


wm75:
Hi,
I just started work on some new training material covering mapping-by-sequencing type of analyses with our MiModD tool suite as a replacement for the old CloudMap tools on Galaxy main and I would like to have opinions where to put this.

MiModD, like the old CloudMap, is for identifying causative mutations behind phenotypes of interest in model organisms. It is for people who have isolated phenotypic mutants and have crossed these mutants to a mapping strain containing lots of marker SNPs. MiModD reidentifies these markers in WGS data from recombinant progeny, performs linkage analyses that show which markers are associated or not with the original phenotype and identifies candidate mutations in the corresponding genomic regions.
With that I think it would fit in the "Variant Analysis" topic quite well. OTOH, the tutorial will be specifically about using the tool on Galaxy main (we already have quite an extensive general tutorial section in the MiModD documentation at http://mimodd.readthedocs.io, which we do not want to duplicate) so you could argue that it also fits in the "Tutorials using public Galaxy at usegalaxy.org" section.

Thanks for your feedback,
Wolfgang

bgruening:
Hi @wm75! Nice to see you here! "Tutorials using public Galaxy at usegalaxy.org" will go away, probably this month and we will group the tutorials into other sections. Tutorials should not depend on a specific Galaxy instance to I would encourage you to use "Variant Analysis".

shiltemann:
Hi @wm75 that sounds like a great tutorial! we are trying to move tutorials out of this `usegalaxy.org` topic, and instead just annotate all trainings with a list of public instances supporting, so please use `variant analysis` topic, this sounds like a good fit :)

wm75:
Thanks for your feedback about where to put the tutorial. I'm about half-way done with it by now. Most importantly, I've created training datasets to work with.
I've read that preferably you would make these available via zenodo links, but I've no experience with zenodo at all. Is material there organized in categories/topics/anything and, if so, is there a recommended place to put training material, or do I just upload the data, get a link and that's it?

shiltemann:
Hi @wm75 awesome! :) Yes,it's basically just upload and get a link. :) Zenodo has communities, and there is one for the [Galaxy training network](https://zenodo.org/communities/galaxy-training/?page=1&size=20). When you upload your data they will ask you if you want to add it to any communities.

bebatut:
This tutorial has been added so I will close this issue

PR #657, opened 2018-01-17
add mapping-by-sequencing tutorial

wm75:
I'm finally done with the new tutorial on Mapping-by-Sequencing ( see #599 ).
Before this gets merged, however, I think I still have a few open questions concerning my tools yaml. Does it look correct and specifically:

- do I need to specify install_resolver_dependencies and install_tool_dependencies for my tools (I couldn't find out much about these parameters just that some tools.yaml files seem to set them, but others don't
    
- for the snpeff repo I want to require the revision indicated in the tools.yaml specifically.
  Is this the right way to do it?
  Could it break other tutorials that do not say which version of the tool they want?
    
- one of the tools in the mimodd_main repo that the tutorial is using produces
  html output and needs to be added to the display whitelist of Galaxy for that output
  to be styled correctly. Is there a yaml or something way to state this?


bgruening:
>  - do I need to specify install_resolver_dependencies and install_tool_dependencies for my tools (I couldn't find out much about these parameters just that some tools.yaml files seem to set them, but others don't

 install_resolver_dependencies is more or less for conda, the other one the classical tool_dependencies of Galaxy, I assume you are using conda for all your tools so setting  install_resolver_dependencies=True and the other to False should be good.

> - for the snpeff repo I want to require the revision indicated in the tools.yaml specifically.
  Is this the right way to do it?
  Could it break other tutorials that do not say which version of the tool they want?

I could yes, in theory we pin all tools with there versions, this would probably be the right thing to do.
Soon we will have testing for workflows and we know when things break.
 
> - one of the tools in the mimodd_main repo that the tutorial is using produces
  html output and needs to be added to the display whitelist of Galaxy for that output
  to be styled correctly. Is there a yaml or something way to state this?

No, but you can set this in the Dockerfile via an env called `GALAXY_CONFIG_....`.

Hope that helps,
Bjoern

wm75:
Ok, so I'm going to update this PR and add the install dependencies options to the tools.yaml.

Do I understand you correctly that the Dockerfile env you mention is something that happens **after** a merge or is there anything I need to do about this right now?

bgruening:
You can just edit this file: https://github.com/wm75/training-material/blob/mapping_by_sequencing/topics/variant-analysis/docker/Dockerfile

wm75:
I see. However, my problem is that I currently cannot get any gtn docker containers to run since the contained Galaxy always fails with:

galaxy.tours INFO 2018-01-18 10:47:06,239 Loaded tour 'core.scratchbook'
Traceback (most recent call last):
  File "lib/galaxy/webapps/galaxy/buildapp.py", line 58, in paste_app_factory
    app = galaxy.app.UniverseApplication(global_conf=global_conf, **kwargs)
  File "lib/galaxy/app.py", line 139, in __init__
    self.tour_registry = ToursRegistry(self.config.tour_config_dir)
  File "lib/galaxy/tours/__init__.py", line 33, in __init__
    self.load_tours()
  File "lib/galaxy/tours/__init__.py", line 58, in load_tours
    self._load_tour_from_path(os.path.join(tour_dir, filename))
  File "lib/galaxy/tours/__init__.py", line 72, in _load_tour_from_path
    tour = tour_loader(conf)
  File "lib/galaxy/tours/__init__.py", line 16, in tour_loader
    title_default = contents_dict.get('title_default', None)
AttributeError: 'NoneType' object has no attribute 'get'

Is that a known issue? Anyway it prevents me from testing anything docker-related.

bgruening:
Oh, no this is not known :(
Which file are you using?

wm75:
So, I guess my previous error report was somewhat inaccurate.
I'm trying to run this:

`docker build -t variant-analysis -f topics/variant-analysis/docker/Dockerfile .`

Note that at this step already I had to cheat a bit and remove all data-library.yaml files from the tutorials in the folder because if I didn't I would end up in a connection retry loop of bioblend sending libraries api requests forever. Without the data-library.yaml files the build completes however, but then when I run:

`docker run -p "8080:80" -t variant-analysis`

the terminal gets flooded with recurring Galaxy startup output that usually pauses briefly at the
```
File "lib/galaxy/tours/init.py", line 16, in tour_loader
title_default = contents_dict.get('title_default', None)
AttributeError: 'NoneType' object has no attribute 'get'
```
I was mentioning before, but I guess it isn't actually failing at this point, but trying to do something repeatedly. When I'm trying to access localhost:8080, while Galaxy is "running" this way, all I'm getting is a 403 Forbidden from nginx.

In the meantime, I tried with a Dockerfile from the chip-seq topic and this one gave the same problem for the data library requests (so, again, I had to remove the corresponding yaml files to continue), but running the container gave a working Galaxy.

Any ideas on either the data library or the Galaxy startup problem?

bgruening:
@wm75 are you doing this on the latest master? We merged my recent Docker work a few days ago with a lot of fixes.

wm75:
@bgruening, yes, I got these errors initially on my older branch, but then I moved to the latest master to see if that helps, but it didn't :(

bgruening:
ok, on it!

bgruening:
Can you please try to uncomment the data_library.yaml file. One of the file can not be uploaded and its ending in a endless loop.

wm75:
yes, that is the first bug I encountered. I don't think it is a specific data_library.yaml file, but any. Happens for me at least with every topic I tried (e.g., chip-seq, too) and also with the latest master unmodified. What I tried already is renaming all the data_library.yaml files inside variant-analysis. If I do that I can build the container, but then when I try to run it, that's when I end up in the Galaxy startup loop.
=> The data-library.yaml problem seems to be topic-independent, the startup problem is variant-analysis-specific

topics/variant-analysis/tutorials/mapping-by-sequencing/tutorial.md
```diff
@@ -0,0 +1,505 @@
...
+> 3. Import the reference genome
+>
+>    > ### {% icon tip %} Tip
+>    > This step is only necessary if the TAIR10 version of the *A. thaliana*
+>    > genome is not available as a built-in genome on your Galaxy server.
+>    {: .tip}
+>
+>    1. Upload the reference genome using the URL:
+>       `https://www.arabidopsis.org/download_files/Genes/TAIR10_genome_release/TAIR10_chromosome_files/TAIR10_chr_all.fas`
+>    2. Truncate the title lines in the fasta dataset to just the bare
```
bebatut: A cosmetic comment...
We are trying to make the name of tool more obvious (like in the first few words) and also having first the name of the parameters and then the value. 
Do you think something we could change the next steps to something like?

```
>    2. Run **Replace parts of text** {% icon tool %} to truncate the title lines in the fasta dataset to just the bare chromosome names with the:
>       - "Find pattern" to `^>(\w+).*`
>            
>            This will find entire fasta title lines, or more specifically, any lines starting with `>` followed by at least one letter, digit or underscore)
>
>       - "Replace with" to `>chr$1`
>           
>            `$1` in it refers to the parenthesized part of the find pattern, *i.e.*, just the substring of letters/digits/underscores directly following the `>`
>
>       - "Find-Pattern is a regular expression" to `Yes`
```
It will render something like:

screen shot 2018-01-23 at 16 56 42

What do you think? And do you think we could have something similar for the other hands-on?

topics/variant-analysis/metadata.yaml
```diff
@@ -78,6 +78,27 @@ material:
       - torhou
       - erxleben
       - bgruening
+  -
+    title: "Mapping-by-sequencing"
+    type: "tutorial"
+    name: "mapping-by-sequencing"
+    zenodo_link: "https://doi.org/10.5281/zenodo.1098033"
+    galaxy_tour: ""
+    hands_on: "yes"
```
bebatut: For `hands_on` and `slides`, can you remove the quotes around `yes` and `no`? They are booleans 😄

wm75: @bebatut Thank you for these comments. I changed all the hands-on sections according to the preferred style.

Regarding the quotes around boolean values in the yaml file: once you're ok with the actual content of the tutorial, it may be easier to reopen this PR against the current tip of master because lots of things have changed in the yaml and dockerfiles since I created my branch.
Does that sound ok for you?

bebatut: The best would be rebase your branch against the master branch. If you want I can do it

wm75: Well, if you have the time that would be great! Just go ahead :)

bebatut:
@wm75 I rebased and fix the metadata and some linting issues. Is it ok for you?

wm75:
So @bgruening told me to ignore the docker issues for now and just focus on getting the tutorial itself ready.
You'd want to check my latest changes, but I think it is ready to get merged.

wm75:
Thanks @bebatut for all your help!

PR #665, opened 2018-01-23
Using latest syntax for icons in template tutorial.md

claresloggett:
The template `tutorial.md` file uses two different syntaxes for icons at the top of marked-up sections: one is `:bulb:` and the other is `{% icon tip %}`. The latter is configurable at the repo level with `_config.yml` and I'm told is the recommended way to do things. I'm not sure if including both syntaxes is deliberate? This PR updates the template to use only the `{% icon name %}` syntax, and adds an explanatory comment.

### DIFF
--- a/templates/tutorials/tutorial1/tutorial.md
+++ b/templates/tutorials/tutorial1/tutorial.md
@@ -7,6 +7,8 @@ tutorial_name: your_tutorial_name
 # Introduction
 {:.no_toc}
 
+<!-- This is a comment. -->
+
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks @claresloggett!

bebatut:
Thanks again @claresloggett for your contribution.
Do not hesitate to add your information in the [contributor file](https://github.com/galaxyproject/training-material/blob/master/CONTRIBUTORS.yaml)

PR #670, opened 2018-01-23
one more test of heroku auto deploy

martenson:


martenson:
@bgruening this works well, our heroku settings is ok, I suspect heroku does not want to auto-deploy PRs that are not mergeable with our setup on training-materials that is all of them (forced reviews, failing travis - which I both disabled for the purpose of this test)

ping @bebatut @shiltemann

Issue #349, opened 2017-06-05
Add category tags to tutorials


Slugger70:
Torsten Seemann asked if we could add tags to tutorials for things like
Prokaryotic vs Eukaryotic
Virus
Minimum Galaxy version (Björn's idea)
etc

shiltemann:
tags are now possible thanks to @bebatut

PR #677, opened 2018-01-30
Add category tags for tutorials

bebatut:
As asked in #349, this PR should add the possibility to have tags for the tutorials in the metadata files that will be display below the tutorial name on the topic page: 

screen shot 2018-01-30 at 15 15 10

Several tags can be added for a tutorial as a list:

screen shot 2018-01-30 at 15 19 57

PS: Currently the labels are not links. But we could think (in the future) to extend this feature to list all the tutorials with the same tag. But I am not sure if it will be easy...

### DIFF
--- a/topics/transcriptomics/metadata.yaml
+++ b/topics/transcriptomics/metadata.yaml
@@ -57,8 +57,6 @@ material:
     hands_on: yes
     slides: no
     workflows: yes
-    tags:
-      - eukaryote
     questions:
       - "What are the effects of Pasilla (PS) gene depletion on splicing events?"
       - "How to analyze RNA sequencing data using a reference genome?"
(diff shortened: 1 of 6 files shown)


topics/transcriptomics/metadata.yaml
```diff
@@ -57,6 +57,8 @@ material:
     hands_on: yes
     slides: no
     workflows: yes
+    tags:
```
bgruening: Most of our tutorials are for eukaryotes and I'm not sure its a good idea to add this particular label to all our tutorials.
I added one for prokaryotes, which imho makes more sense.

bebatut:
@bgruening I added more "prokaryote" and removed the "eukaryote" tag
I see also the issue with table

screen shot 2018-04-27 at 12 13 14

@willdurand any idea how we could fix that?

willdurand:
> @willdurand any idea how we could fix that?

we can center the icons (vertically) but that's it. You would have the same output if the titles were longer, I don't think it is an issue.

bgruening:
Thanks B!

PR #678, opened 2018-01-30
[WIP] Reformate the tutorial on the technical infra

bebatut:
I reformate the tutorial to:
- Add how to pre-generate the `tools.yaml` file from the workflow file
- Add a section on how to populate and prepare an existing Galaxy instance for a tutorial

Btw, @shiltemann do we have the scripts to help for the second point? We should then document them here. What do you think?


shiltemann:
@bebatut I did work on some scripts for that at the Cambridge hackathon, yes. Will have a look to see if it is working/still up-to-date etc and add a section here :)

shiltemann:
(contents merged into `training_refactor` branch)

PR #713, opened 2018-03-06
galaxy tour for hicexplorer

polkhe:
@joachimwolff

### DIFF
--- a/topics/epigenetics/tutorials/hicexplorer/tours/tour.yaml
+++ b/topics/epigenetics/tutorials/hicexplorer/tours/tour.yaml
@@ -43,9 +43,7 @@ steps:
   - title: Uploading the input data
     element: .upload-text-column .upload-text .upload-text-content.form-control
     content: >-
-      Insert the links here.<br> Load the example dataset into your history from
-      <a target="_blank" 
-      href="https://zenodo.org/record/1183661">Zenodo</a>
+      Insert the links here.<br> Load the example dataset into your history from <a target="_blank" href="https://zenodo.org/record/1183661">Zenodo</a>
     placement: right
     textinsert: |-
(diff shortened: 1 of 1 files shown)


bgruening:
ping @joachimwolff

joachimwolff:
Thanks for creating the tour, it works in general. 

I have one major issue:
-  "Computing A/B compartments": Wrong tool is used. It is not hicFindTADs but hicPCA.

And a few minor: 
1. Introduction: It would be nice to have the list as a list and not as concatenated text. 
image
2. "Uploading the new data": Make the link to zenodo open in a new tab and not in the same window.
3. "Uploading the input data": The guide says I have to click on "Past/Fetch Data". If I do so and click next, the same action is performed by the tour too. 
image
And then it looks like this 
image and I get four input files. 
4. On smaller screens you cannot edit the attributes because the tour is overlapping the "save" button.
image. Concerning the text: It is "Datatype_s_" with the 's' in the end.
5. BWA-MEM: Maybe add a reminder that they have to use "ctrl"-key to select multiple files. 
6. To click on: "Execute" can not be automatized? 
7. "Creation of a Hi-C matrix": `The extension of output matrix files is .h5`. Please add the cool-format.
8. "Rename the output": Change text to "Rename the output 'hicBuildMatrix MATRIX on data 6 and data5' to 10kb contact matrix'
9. The boolean options are "Yes" and "No" and not "True" and "False" in the Galaxy interface. Please correct the text. 
10. Plotting TADs 1/3: There is no 'color track' for the bedgraph matrix. Please remove this point.
11. On the last card there is one empty bullet point.

bgruening:
1) can be done with HTML elements like `<ul><li>` I think.
3) we can simply remove this sentence *or* the action as you like.
4) it not possible I guess ...
6) yes can be - the Galaxy introduction tour does this imho

PR #724, opened 2018-03-21
Fix typo

nsoranzo:


willdurand:
@martenson could you take a look at the Heroku logs? I wonder why there is a problem deploying this change. cheers!

martenson:
@willdurand 
```
ownloading nokogiri-1.8.1 revealed dependencies not in the API or the lockfile
       (mini_portile2 (~> 2.3.0)).
       Either installing with `--full-index` or running `bundle update nokogiri` should
       fix the problem.
 !
 !     Failed to install gems via Bundler.
 !
 !     Push rejected, failed to compile Ruby app.
 !     Push failed
```

martenson:
also 
```
       Warning: the running version of Bundler (1.15.2) is older than the version that created the lockfile (1.15.3). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
```

PR #734, opened 2018-04-08
Updates to the location of the hg19

ReginaEsinamAbotsi:
Hi there!

I have suggested an update to the location of the hg19 file.
I will appreciate it if you could please check it out.

Thanks.
Regina


ReginaEsinamAbotsi:
I have also noticed that the location instruction is only valid for usegalaxy.eu not on usegalaxy.org. Is there a way we can make this clear in the tutorial?

ReginaEsinamAbotsi:
Hi there,

At Line 232
1. Create a pedigree file (PED) like this

Please pardon my ignorance, but there is no clear instruction on how to create the PED file.

Could you please clarify ?

Thanks a lot.

Regina

PR #742, opened 2018-04-20
"fork" is probably unknown to our users

bgruening:


_includes/default-header.html
```diff
@@ -12,8 +12,8 @@
             <div class="collapse navbar-collapse" id="top-navbar">
                 <ul class="navbar-nav">
                     <li class="nav-item">
-                        <a class="nav-link" href="{{ site.github_repository }}" title="Fork on GitHub">
-                            {% icon github %} Fork me on GitHub
+                        <a class="nav-link" href="{{ site.github_repository }}" title="Edit on GitHub">
+                            {% icon github %} Edit me on GitHub
```
willdurand: That's not an "edit" link though. Since "Fork" is a GitHub concept, I'd argue that it is unknown to anyone who never went to GitHub before. If you want to change it, maybe "Browse" or "View" would make more sense.

bgruening: I was following readthedocs and there it is also `Edit`. See for example here: https://deeptools.readthedocs.io/en/develop/

PR #757, opened 2018-05-16
Fix for display of input datasets in de novo rna-seq

blankenberg:
https://zenodo.org/record/583140/files/G1E_rep1_forward_read_%28SRR549355_1%29 was not previously being displayed


bgruening:
Thanks Dan!

Issue #756, opened 2018-05-16
Structure of the training data on the shared data library


bebatut:
It would be nice if the training data associated to the tutorials are found in the same place in the different Galaxy instances ("Shared data"), following the structure:

```
Data libraries   
  |_ GTN - Training
        |- topic1
        |   |_ tutorial1
        |   |   |_ DOI from Zenodo 
        |   |       |_ each data file
        |   |_ tutorial2
        |       |_ DOI from Zenodo 
        |_ topic2
            |_ tutorial1
                |_ DOI from Zenodo
```

This structure would be in the `data_library.yml` files with the tutorials. This file could be used to populate the data libraries in any Galaxy instance and also the shared CVMFS (used by the Docker images soon and by the usegalaxy.*)

bgruening:
```
Data libraries   
  |_ GTN - Training
        |- topic1
        |   |_ tutorial1
        |   |   |_ DOI from Zenodo 
        |   |       |_ each data file
        |   |   |_ DOI-2 from Zenodo 
        |   |       |_ each data file
        |   |_ tutorial2
        |       |_ DOI from Zenodo 
        |_ topic2
            |_ tutorial1
                |_ DOI from Zenodo
```

bebatut:
This structure will be implemented in the training command of planemo: https://github.com/galaxyproject/planemo/pull/861

erasche:
Done!

PR #759, opened 2018-05-22
example of the new data library format

bgruening:
xref: https://github.com/galaxyproject/training-material/issues/756


bebatut:
All the data library files will be updated automatically soon....

PR #770, opened 2018-05-22
testing methylation tutorial

bgruening:
This workflow is not working as it seems.

Problems:
  - reference data can be missing. This is a general problem which we can fix by using CVMFS in Docker. But I think this will be not a problem on public Galaxy instances.
  - wigToBigWig: command not found for some tool
  - the  file GRCh38_ensembl2UCSC.txt is not used afaik

ping @joachimwolff


jmchilton:
Freshly minted planemo 0.53.0 should be ready to go with the new syntax for specifying inputs as URIs. Thanks a ton for pushing this @bgruening !

PR #771, opened 2018-05-22
testing unicycler assembly tutorial

bgruening:
This is WIP, as unicycler is running for 5h now on my computer.
Other than this there is a strange error somewhere, that one tool can not find its inputs.

Pushing the current state


nekrut:
we also need to see if bioconda recipe for unicycler needs updating

PR #780, opened 2018-05-23
Added schema.org annotations for training materials

njall:
I've formatted each training material's metadata into a JSON-LD representation of schema.org's [CreativeWork](http://schema.org/CreativeWork) vocabulary. 

There is a small issue. There's several empty {} items (in hasPart, mentions, and contributor). This is due to Liquid templating engine not being able to handle arrays very well. An empty item at the end of array objects ensures there's no trailing comma. Metadata harvesters and Google will strip this empty content out I think so it's just stylistically unfortunate.


_layouts/topic.html
```diff
@@ -50,6 +50,91 @@ <h2 id="material">Material</h2>
...
+                                    "@type": "Thing",
+                                    "url": "{{material.zenodo_link}}",
+                                    "name": "Training data for {{material.title}} tutorial"
+                                },
+                            {% endif %}
+                            {% if material.workflows %}
+                                {
+                                    "@type": "Thing",
+                                    "url": "{{ site.github_repository  }}/tree/master/topics/{{ topic.name }}/tutorials/{{ material.name }}/workflows/",
+                                    "name": "Workflow for {{material.title}} tutorial"
```
bgruening: `Workflow(s)`

willdurand:
> There is a small issue. There's several empty {} items (in hasPart, mentions, and contributor). This is due to Liquid templating engine not being able to handle arrays very well. An empty item at the end of array objects ensures there's no trailing comma. Metadata harvesters and Google will strip this empty content out I think so it's just stylistically unfortunate.

What you probably want is to use the attributes of the `forloop` object, like `.last` for example: https://help.shopify.com/themes/liquid/objects/for-loops#forloop-last.

willdurand:
Also, it would probably make sense to add a new file with the JSON content to ease readability: http://talk.jekyllrb.com/t/how-to-add-a-schema-for-each-blog-post/1149/4

njall:
Agreed about separating json for readability. 

The forloop last thing is trickier though as you cannot easily create array variables within the template to take advantage of the loop _last_ feature. The only way would be to use the split function which seems very messy 
https://heliumdev.com/blog/create-an-array-in-shopifys-liquid

willdurand:
> The forloop last thing is trickier though. You cannot easily create array variables within the template to take advantage of the loop last feature.

I am not sure to understand, for the `contributor`, you always put a trailing comma for each item, but it could be done differently no? Like, only adding the comma if it is not the last iteration? Or am I missing something?

njall:
Sorry… you can for contributor, yes. But not for hasPart or mentions where they are not already in array format

_includes/material.jsonld
```diff
@@ -0,0 +1,69 @@
...
+                        "name": "Slides for {{material.title}}"
+                      },            
+            {% endif %}
+            {% if material.galaxy_tour %}
+                    {
+                        "@type": "CreativeWork",
+                        "url": "{{ site.github_repository }}/tree/master/topics/{{ topic.name }}/tutorials/{{ material.name }}/tours/",
+                        "name": "Galaxy Tour for {{material.title}}"
+                      },                
+            {% endif %}
```
willdurand: There are only two items here, so if there are slides, we can maybe add the comma only if there are also a galaxy tour no?

I have something like this in mind but maybe it does not work at all...

```
{% if material.slides %}
  {
    ...
  }
{% endif %}
{% if material.galaxy_tour %}
{% if material.slides %}
   ,
{% endif %}
  {
    ...
  }
{% endif %}
```

njall: Yeah, good point. I've added that work-around

bebatut:
Thanks @njall

bgruening:
:100:

PR #793, opened 2018-05-30
Hyperlink DOIs against preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). Yes, a bit ironic, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org) and the old links with the `dx` subdomain continue to work.

So, there is no urgent need to do anything, but since you also refer to Zenodo, which started using `https://doi.org` for its badges as well, I'd like to suggest to hereby update all static DOI links and any code that generates new ones.

Cheers!


bgruening:
@katrinleinweber thanks a lot! Also for all the other PRs today :)

katrinleinweber:
You're welcome :-)

bebatut:
Thanks a lot @katrinleinweber

katrinleinweber:
You're welcome! I thank you for merging :-)

PR #796, opened 2018-05-31
Update CONTRIBUTORS.yaml

hrhotz:
add extra information

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -127,7 +127,7 @@ hrhotz:
     twitter: hrhotz
     gitter: hrhotz
     orcid: 0000-0002-2799-424X
-    linkedin: Hans-Rudolf Hotz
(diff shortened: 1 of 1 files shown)


CONTRIBUTORS.yaml
```diff
@@ -123,6 +123,11 @@ hmenager:
 
 hrhotz:
     name: Hans-Rudolf Hotz
+    email: hrhotz@googlemail.com
+    twitter: hrhotz
+    gitter: hrhotz
+    orcid: 0000-0002-2799-424X
+    linkedin: Hans-Rudolf Hotz
```
bgruening: mh, should this be: `hans-rudolf-hotz-542b31`

hrhotz: where ?

hrhotz: ...now I understand...will fix it

PR #809, opened 2018-06-06
Remove mention of the external de novo workflow inside the tutorial

bebatut:
To fix the dead link


nsoranzo:
The link points to a Galaxy Main shared workflow which generates an Internal Server Error, is this tracked on https://github.com/galaxyproject/galaxy ?

nsoranzo:
Reported at https://github.com/galaxyproject/galaxy/issues/6296 .

PR #750, opened 2018-05-03
Some makefile fixes

bebatut:
- Add make target to help conda env installation 
- Use only gem for installation
- Remove Gemfiles
- Update environment file
- Update package files
- Fix installation command
- Remove bundle usage

PR #813, opened 2018-06-06
attempt to fix heroku review apps deployment

martenson:


martenson:
The build log seems fine, this might actually be us hitting unverified account heroku limit. I will see what I can do with it.

martenson:
someone with better Ruby knowledge might need to have a look here

willdurand:
Well, I think it makes sense since https://github.com/galaxyproject/training-material/pull/750 removed `Gemfile` and `Gemfile.lock` and bundler support.

Heroku does not know which stack to use or tools to install. I think the `Gemfile`/`Gemfile.lock` as well as Bundler support should be re-added because there is no reason not to have them.

martenson:
I was curious why gemfile is not in the tree.

willdurand:
> I was curious why gemfile is not in the tree.

Yeah, hopefully it was a mistake and not the cause of some install errors so that we can re-add these files soon.

martenson:
this has been remedied via https://github.com/galaxyproject/training-material/pull/1485

PR #814, opened 2018-06-07
Add contact email in config file

bebatut:


_config.yml
```diff
@@ -15,7 +15,7 @@
 # in the templates via {{ site.myvariable }}.
 ---
 title: "Galaxy Training!"
-email: your-email@domain.com
+email: galaxytrainingnetwork@gmail.com
```
erasche: off topic: would you like `training@usegalaxy.eu` to go anywhere special?

bgruening: GTN mailing list?

nsoranzo: Aren't the `gitter_url` and `biostars_url` enough?

bebatut: @nsoranzo Some people may prefer to send emails, for example to ask for help/information for new tutorial/topic if they are not sure. It may feel less "impressive" (maybe I am wrong)
@bgruening GTN mailing list there or for the `training@usegalaxy.eu` ?

bgruening: Its this one isn't it? http://galaxy-training-mailing-list-archive.35427.n7.nabble.com/

shiltemann:
merging this now because it is used in the FAQ, if we decide on a different email address we can always change it later :)

PR #828, opened 2018-06-13
improve grammar

martenson:
imo this whole section could be excluded from the admin training topic, since it does not make much sense

### DIFF
--- a/_layouts/topic.html
+++ b/_layouts/topic.html
@@ -155,7 +155,7 @@ layout: base
         <h2 id="docker_image">Galaxy instances</h2>
         <p>You can use a public Galaxy instance which has been tested for the availability of the used tools. They are listed along with the tutorials above.</p>
         {% if topic.docker_image and topic.docker_image != "" %}
-            <p>You can also use the following Docker image for these tutorials.</p>
+            <p>You can also use the following Docker image for these tutorials:</p>
             <pre><code>docker run -d -p 8080:80 {{ topic.docker_image }}</code></pre>
             <p>It will launch a flavored Galaxy instance available on <a href="http://localhost:8080">http://localhost:8080</a>.</p>
         {% endif %}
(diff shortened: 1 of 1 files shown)


bebatut:
Looks like there is a Docker image for the admin topic. If you think it is not useful, you can remove the Docker image name in the `metadata.yaml` (`docker_image: `) of the admin topic. Then the possible Docker image will not appear on the topic page :smile:

_layouts/topic.html
```diff
@@ -153,9 +153,9 @@ <h2 id="material">Material</h2>
         </table>
 
         <h2 id="docker_image">Galaxy instances</h2>
-        <p>You can also a public Galaxy instance which has been tested for the availability of the used tools. They are listed along with the tutorials above.</p>
+        <p>You can use a public Galaxy instance which has been tested for the availability of the used tools. They are listed along with the tutorials above.</p>
         {% if topic.docker_image and topic.docker_image != "" %}
-            <p>You can also use for these tutorials the dedicated Docker image:</p>
+            <p>You can also use the following Docker image for these tutorials.</p>
```
nsoranzo: I'd keep the colon instead of the period, FWIW.

nsoranzo:
A Docker image still makes sense, but a public Galaxy server does not.

martenson:
@bebatut but the whole section will still render no matter if the image is there I think: https://github.com/galaxyproject/training-material/blob/master/_layouts/topic.html#L155

bebatut:
@martenson True: I was checking something else and missed what you meant 😟

martenson:
@nsoranzo even the docker training does not use this docker image, I do not think it is of much use here.

martenson:
@nsoranzo that sounds awesome, but if there is no image yet, I would refrain from advertising it on the page (similar as linking to [incomplete tutorials](https://github.com/galaxyproject/training-material/blob/master/topics/admin/tutorials/dev-to-production/tutorial.md), and providing placeholder files in repo)

nsoranzo:
Right now we have a Docker image per topic, in fact the template uses `topic.docker_image`, so we would need to extend that to tutorial-specific images.

PR #834, opened 2018-06-14
fix yamllint warnings

shiltemann:
..they were making it hard to find the errors in output log


shiltemann:
oh, nope, don't merge this yet, our template doesnt like it, will fix later

shiltemann:
that's better (tho I guess yes/no is maybe a tad more intuitive for some people, hmm)

shiltemann:
much better this way, thanks @erasche

PR #839, opened 2018-06-18
Fix Jen's Twitter username

bebatut:
@jennaj Can you confirm it is correct? `jenhjackson` was failing on the tests


jennaj:
Oh, I was messing with my account. I'll fix that. Had no idea it would impact all the way here :)

bebatut:
Ok. I close the PR then. Thanks!

PR #849, opened 2018-06-25
Modified admin/connect-to-compute-cluster tutorial to change .ini to …

Slugger70:
….yml refs.


natefoo:
Might want to make those, e.g.:

```yml
galaxy:
    tool_config_file: /srv/galaxy/config/tool_conf.xml,/srv/galaxy/config/shed_tool_conf.xml
```

So it's clear they need to be in the `galaxy` dict (at least that's what I did in other exercises)

PR #858, opened 2018-06-26
Modified conclusion section

MoHeydarian:
Modified conclusion section to summarize de novo transcriptome reconstruction RNA-seq pipeline. Previous version had been copied and pasted from the ref-based RNA-seq tutorial.


bebatut:
@MoHeydarian all the changes here are already in the current version of the tutorial: https://github.com/galaxyproject/training-material/blob/master/topics/transcriptomics/tutorials/de-novo/tutorial.md
I close the PR. Thanks for having thought about it :smile:

PR #859, opened 2018-06-26
[WIP] initial transfer of uwsgi training from dagobah

martenson:
needs

- [ ] missing picture
- [ ] better adding to the slides using `--` so it does not actually reshuffle _everything_
- [ ] fix the text alignment to be left instead of center


erasche:
Hi @martenson, we [changed the way metadata works](https://github.com/galaxyproject/training-material/pull/966) in the repository, so I updated your PR accordingly, metadata is now at the top of the slides.html file.

erasche:
Doing this in a new PR. Will address review comments.

PR #862, opened 2018-06-26
Japanese Translation for 3 tutorials

A-Asai:
We have translated 3 tutorials and completed reviewing by Japanese community.
- peak2genes
- galaxy-intro-strands
- mothur-miseq-sop


shiltemann:
@A-Asai cool! thanks a lot :) 

I cannot comment on the content, but it looks good to me. I will see if we can get some mechanism in place to switch languages in the website interface some time :)

bebatut:
We finally use automatic translation on the fly of all tutorials using Galaxy Translate. It avoid having to update the translation every time the tutorial changes. It means less effort for the community. So I will close this PR.

On the other hand, could you check if the Japanase translation of some tutorials are correct? 
We are working on the tutorials themselves to improve the english and then have better automatic translation. We also avoid translation of some words. But we need more evaluations and feedback of the automatic translation

PR #864, opened 2018-06-26
Attempt to fix slide render problem with Tool Deps and Conda slides

natefoo:
Right now it looks like this: 

slide

And a bunch of the following slides are unviewable. But this fix is just a guess as to what's wrong (I'm using the PR to build the slides to test).


natefoo:
Actually I guess I'm assuming it's possible to view the Travis build but that's probably not the case?

natefoo:
Yay, I found the training material on how to run the training material site 😄

So, this looks good to me locally.

PR #868, opened 2018-06-27
Allow usernames without names

erasche:
I noticed that this page https://galaxyproject.github.io/training-material/topics/dev/ had an empty item at the bottom so added support for usernames without a `name` field attached to it. Probably not necessary to do that since everyone has a `name:` but it made tracking down the underlying metadata issue a bit easier.


bebatut:
Thanks @erasche

PR #895, opened 2018-06-30
Style and grammar tweaks

mccalluc:
I was confused by this section:
> The largest contig is 4,576,290 bp (for comparison E. coli K12 MG1655 strain genome length is 4,656,144 bp) and the smallest is 4,581,676 (total length) - 4,576,290 (length of the largest) = 5,386 bp. 

"Largest" and "smallest" seem to be reversed? and the fact that the difference between them is small is just an accident, right? Maybe just remove it if it's not adding anything.

### DIFF
--- a/topics/assembly/tutorials/unicycler-assembly/tutorial.md
+++ b/topics/assembly/tutorials/unicycler-assembly/tutorial.md
@@ -6,9 +6,9 @@ tutorial_name: unicycler-assembly
 
 # The goal: *E. coli* C-1 assembly
 
-In this tutorial we assemble and annotate genome of *E. coli* strain [C-1](http://cgsc2.biology.yale.edu/Strain.php?ID=8232). This strain is routinely used in experimental evolution studies involving bacteriophages. For instance, now classical works by Holly Wichman and Jim Bull ([Bull et al. 1997](https://www.ncbi.nlm.nih.gov/pubmed/9409816), [Bull & Wichman 1998](https://www.ncbi.nlm.nih.gov/pubmed/9767038), [Wichman et al. 1999](https://www.ncbi.nlm.nih.gov/pubmed/10411508)) have been performed using this strain and bacteriophage phiX174.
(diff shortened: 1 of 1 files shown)


bebatut:
ping @Delphine-L

Delphine-L:
@mccalluc Sorry for the confusion, I mean here that the smallest is 5,386 bp, that I get by subtracting the length of the other contig from  the total length of the genome , but we can remove the substraction if it's confusing

mccalluc:
Oh, I get it now: I won't add any edits around that section in this PR, so it can be reviewed as-is.

If other folks have thoughts about how to rephrase that, great, or it can just be left in its present state.

topics/assembly/tutorials/unicycler-assembly/tutorial.md
```diff
@@ -61,38 +61,38 @@ In this analysis we will perform two tasks: (1) assembly and (2) annotation. Bel
...
 
-![Unicycler process](../../images/unicycler.png "Simplified view of the Unicycler assembly process (From Wick:2017). In short, Unicycler uses SPAdes (see below) to produce an assembly graph, which is then bridged (simplified) using long reads to produce longest possible set of contigs. These are then polished by aligning original short reads against produced contigs and feeding these alignment to Pilon - an assembly improvement tool.")
+![Unicycler process](../../images/unicycler.png "Simplified view of the Unicycler assembly process (From Wick:2017). In short, Unicycler uses SPAdes (see below) to produce an assembly graph, which is then bridged (simplified) using long reads to produce the longest possible set of contigs. These are then polished by aligning the original short-reads against long-read contigs and feeding these alignments to Pilon - an assembly improvement tool.")
```
Delphine-L: The alignment is short reads against the contigs that were previously assembled from both short and long reads,  not against long reads

mccalluc:
@Delphine-L : Sorry about that. See if that section seems right now, or if it needs more work.

PR #897, opened 2018-06-30
Move RAD-seq tutorials to a new topic "Reduced representation sequencing"

bebatut:


topics/reduced-representation-seq/docker/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+#    docker build -t <your_tag> -f topics/<your_topic>/docker/Dockerfile .
+#
+# to run image:
+#    docker run -p "8080:80" -t <your_tag>
+
+FROM bgruening/galaxy-stable
+
+MAINTAINER Galaxy Training Material
+
+ENV GALAXY_CONFIG_BRAND "GTN: <your_topic>"
```
shiltemann: still some `<your_topic>` placeholders in this file

topics/reduced-representation-seq/metadata.yaml
```diff
@@ -0,0 +1,122 @@
+---
+name: reduced-representation-seq
+type: use
+title: Reduced-representation sequencing
+summary: "Plant, and particularly cereal genomes, are challenging to sequence due\
```
shiltemann: ok, so I know next to nothing about this topic, but it sounds a bit narrow, should we broaden the topic to something like "plant genomics" orso? what do you think?

topics/reduced-representation-seq/slides/introduction.html
```diff
@@ -0,0 +1,53 @@
+---
```
shiltemann: remove this file?

topics/reduced-representation-seq/README.md
```diff
@@ -0,0 +1,4 @@
+Topic name
```
shiltemann: update

topics/reduced-representation-seq/metadata.yaml
bebatut: @yvanlebras what do you think?

yvanlebras: Not sure to fully understand, sorry, but for me reduced representation sequencing is ok. If we want to give it a topic name less technologic and more related to biological applications, this will be more related to non-model species population genomics and/or quantitative genetics. But true, this is a kind of approach used for plant genomics even if rad-seq is more used by marine population studies...

yvanlebras: One thing, and we have discuss a little about that with Bérénice in Portland, is that actually, training topics are more related to "technologies" or Bioinformatics approach as assembly / chi-seq / epigenetics / annotation / metagenomics / proteomics / transcriptomics .... so I really think that maybe it's better to keep this kind of approach for the RAD-seq one, and Reduced Representation Library / Reduced Representation Sequencing is something relevant IMO.... And there is others methods than RAA-seq to do so + communities interested "only" on these approaches like plant scientists and fish scientists and for molecular ecology as selection and livestock production. To go further, I think it will be interesting to have another entry point to all training materials, using "biological topics"...

bgruening: I don't think its "technologies". It was very in the beginning and ChIP-seq is a left-over of this. But for example epigentics is a broader topic, not a technique. We have indeed HiCseq and MethylCseq in it. To distinct techniques. Same for Proteomics and we probably will have Metabolomics with different techniques soon.

yvanlebras: Agree, and chip-seq is really IMO same level than RRL/ Reduced Representation Sequencing.... But not related to an approach in Biology (physiology/quantitative genetics/population genetics/molecular ecology/macro-ecology/..) so more related to technical things IMO

shiltemann: I agree with @bgruening, we currently divide topics into the thing that is studied (genome, transcriptome, proteome, epigenome) each of which has multiple techniques within it (I agree Chip-seq is an exception to this and I think that should also be moved/renamed) 

But, we did add tutorial tags recently, so we can add a tag for the technique used (and some day users will be able to search on these tags across topics).  

Technologies come and go, but the biological areas of research will remain, and I am just a little afraid of getting a long list of technology topics that will only ever have a small number of tutorials in them before they become outdated, and that is why I would personally prefer the topic to be named something that is a bit broader/biological. 

If your tutorials are really about the technology itself and not just using it to answer a biological question, then we could consider a topic like "Sequencing Technologies" 

That said, this is not a simple problem and our current topic list also isn't perfect, I agree, so thanks for the discussion here :)

yvanlebras:
Thank you all for discussing on this, I think, quite important subject! I'm agree with a classification as genomics/transcriptomics/proteomics/metabolomics/epigenetics, and if I follow you, chip-seq have to move (so the RAD-seq too) into something like a broader genomics topic. ITM what about assembly (need to be under genomics?) and metagenomics too (if we anticipate the next metaproteomics training metrial for example)? Or a meta-omics topic is more adequat.... It seems to me that this is not an easy question this classification under general topics....

PR #909, opened 2018-07-02
Update metadata.yaml

NicHerndon:
Added **Key points** to "Understanding Galaxy history system".


topics/introduction/metadata.yaml
```diff
@@ -248,6 +248,13 @@ material:
       - "Gain understanding on navigating and manipulating histories"
     time_estimation: "30 min"
     key_points:
+      - In Galaxy, data is stored in datasets and organized in histories.
+      - The right pane shows the current history. It displays the name of the history, the tags and annotations assigned to it, and the size of the data in it.
+      - The datasets are shown differently depending on their status: queued, running, ok, error, or paused.
```
shiltemann: could you put quotes around this line? (since it contains a colon)

topics/introduction/metadata.yaml
```diff
@@ -248,6 +248,13 @@ material:
...
+        - The summary view for each dataset can be expanded to show additional information.
```
shiltemann: there are a few too many spaces on beginning of this line and next

shiltemann:
thanks a lot @NicHerndon! :)

topics/introduction/metadata.yaml
NicHerndon: Thank you for pointing this out. I removed the extra spaces.

topics/introduction/metadata.yaml
NicHerndon: Sure. Hope it looks better now.

shiltemann: sorry, I meant quote the entire line (so like the objectives and questions above it) this is because since the colon has a special meaning in yaml, the parser will not understand it without the quotes.

NicHerndon: I see. I changed it.

PR #924, opened 2018-07-05
Replacement of zenodo link 

Delphine-L:
Replacement of zenodo link by a zenodo repo that contains only the dataset necessary for the training. The previous Zenodo repo contains also the raw data, too big for the import.


bgruening:
Thanks!

PR #928, opened 2018-07-11
Ansible: add note about choosing a role

erasche:
Figured this might be useful to provide some heuristics for choosing a role. E.g. trying to find the right one from [these](https://galaxy.ansible.com/search?vendor=false&keywords=telegraf&order_by=-relevance&page_size=10) or [these](https://galaxy.ansible.com/search?vendor=false&keywords=rabbitmq&order_by=-relevance&page_size=10) is not trivial.


shiltemann:
thanks!

PR #929, opened 2018-07-11
tour for rna-seq-viz-with-cummerbund

polkhe:
@bebatut added tour


bebatut:
@bagnacan could you test this tour? Thanks :smile:

shiltemann:
Thanks @polkhe!

PR #942, opened 2018-07-19
Update tutorial.md

subinamehta:


subinamehta:
I just put in a video that we created for metaproteomics as well as the website we usually use for metaproteomics training.

bgruening:
Thanks!

PR #961, opened 2018-08-17
yamllint: don't fail on trailling whitespace

shiltemann:
no longer fail the travis test due to trailing spaces in yaml files


bgruening:
Thanks!

PR #966, opened 2018-08-17
Move metadata to file headers

### LINKED ISSUE #947: Metadata of a tutorial in the same Markdown file as the tutorial
As discussed during BOF at GCC/BOSC 2018 (see #941), it would be nice to have metadata of a tutorial in the same Markdown file as the tutorial to simplify the contribution.

erasche:
Moves all metadata to headers of the slides/tutorial documents. Fixes #947 

pros:

- It's pretty great
- No more `materials` item in the metadata.yaml
- new files added to directories will be automatically picked up
- drop your file with metadata in and you're ready to go!
- slides/hands-ons can now have separate contributors list and track those/give credit more accurately
- aimed for and mostly achieved line-for-line compatibility with a build from master branch at the same point. The only differences are orderings of the tutorials (in this PR tutorials + many other lists are rendered is alphabetized to make the build reproducible/easier to diff for the next person)
- can now move tutorials between folders without editing the tutorial to note the folder name

cons:

- slight duplication between tutorial/slides, but these will likely diverge over time and is probably less of a concern
- the plugin runs like a million times, crickey. It could be memoized and the data stuck in site or something. Low hanging fruit.

todo:

- [x] update tutorial on metadata
- [x] check external hands-ons
- [x] requirements pages in slides
~~- [ ] fix TeSS scraping~~ looks like we will break it temporarily while they migrate to json-ld


bgruening:
@njall if we merge this could you update the TeSS integration?

shiltemann:
this looks great, I think it makes a lot of sense to keep metadata in same file as tutorials. I can't see any difference in the served site, so that's perfect ;)

And thanks for also updating the template files and tutorial on how to fill the metadata btw :) 

..guess the only small drawback is that things like key points and objectives are now duplicated in two places for modules that have both a hands-on tutorial and slides, but not sure if there is any good solution for that, or if it even really matters (and maybe sometimes even desirable to define different keypoints/objectives etc for the slides vs the hands-on tutorial?)

thanks a lot @erasche :+1:

njall:
Ah great work guys! Looks good. @bgruening - I didn't finish the new JSON-LD scraper, we still use the YAML one. I'll write the JSON-LD scraper this week and switch it over asap.

PR #968, opened 2018-08-21
Update to Makefile for more flexible hosting params and incremental builds

mtekman:


Makefile
```diff
@@ -46,11 +49,11 @@ install: clean ## install dependencies
 .PHONY: install
 
 serve: ## run a local server}
-	${JEKYLL} serve -d _site/training-material
+	${JEKYLL} serve -d _site/training-material -P ${PORT} -H ${HOST} `[ "${INCREMENTAL}" != "" ] && echo '--incremental'`
```
bebatut: Why not using just `${INCREMENTAL}` there and define `INCREMENTAL?=` or `INCREMENTAL?=--incremental`?

mtekman: I thought of that initially, but then the command line would look a bit weird:

 `make serve HOST=something PORT=1234 INCREMENTAL=--incremental`

I thought instead that something like:

 `make serve HOST=something PORT=1234 INCREMENTAL=yes`

would look better on the cli, but with the flexibility that the `yes` value could be anything

 `make serve HOST=something PORT=1234 INCREMENTAL=true`
 `make serve HOST=something PORT=1234 INCREMENTAL=T`

bebatut: even `make serve HOST=something PORT=1234 INCREMENTAL=false` currently :smile:

mtekman: true --- okay, I've left it up to the user to set the INCREMENTAL var, although this could be used as a hook to pass any kind of flag to jekyll. Not sure if this is the best way, but I do not Make well enough to be able to override shell ENV variables

`make serve HOST=something PORT=1234 INCREMENTAL=--incremental`

erasche:
Thanks @mtekman!

PR #985, opened 2018-08-27
Add stricted check of front matter by jekyll

bebatut:
To check the YAML on the top of the Markdown files (now that most of the metadata are there)


erasche:
Looks like it's [just checking for valid yaml](https://github.com/jmhooper/jekyll_strict_front_matter/blob/master/lib/jekyll_strict_front_matter.rb) which is definitely useful. I'll see if I can write a plugin for doing stricter validation at some point maybe?

Issue #981, opened 2018-08-27
Feedback forms: Small box at the bottom with :) / :( and a short feedback form.


bebatut:
Most likely to receive feedback than any other method, as discussed during the last GTN meetings #954 #941

bebatut:
Feedback form is now embed at the bottom of the tutorial and the result sent to: https://github.com/galaxyproject/training-material/issues/989

PR #988, opened 2018-08-27
Embed a short feedback form at the bottom of the tutorial

bebatut:
To solve #981, a feedback form can be embedded at the bottom of the tutorial (with the tutorial name already filled):

screen shot 2018-08-28 at 14 00 13

There is 2 issues with the current implementation:

- I can not get rid of the header of the form (previous screenshot)

   I tried to play with it both on Google form side and CSS. But no way to get rid of it. I think we will keep it this way

- When leaving the page, a pop up box open with "This page is asking you to confirm that you want to leave - data you have entered may not be saved" (on Firefox)

@willdurand any idea?

### DIFF
--- a/_layouts/tutorial_hands_on.html
+++ b/_layouts/tutorial_hands_on.html
@@ -217,9 +217,9 @@ layout: base
         </blockquote>
         {% endif %}
 
-        <br>
         <h3>{% icon congratulations %} Congratulations on successfully completing this tutorial!</h3>
-
+        <hr>
+        <br>
 
         <iframe class="google-form" src="https://docs.google.com/forms/d/e/1FAIpQLSd4VZptFTQ03kHkMz0JyW9b6_S8geU5KjNE_tLM0dixT3ZQmA/viewform?embedded=true&entry.1235803833={{ page.title }} ({{ topic.title }})">Loading...
         </iframe>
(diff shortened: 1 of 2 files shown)


shiltemann:
We have also added a script to the google form to automatically post the responses to a [github issue](https://github.com/galaxyproject/training-material/issues/989). This way feedback is transparent to everybody. I have added a disclaimer to the form telling people their feedback will be anonymous but public

_layouts/tutorial_hands_on.html
```diff
@@ -221,12 +221,16 @@ <h3>{% icon congratulations %} Congratulations on successfully completing this t
 
         <hr>
 
+        <p>Please let us know what you thought of this tutorial below! This feedback should be about the training materials themselves only. If tools were not working or missing, please contact the administrators of the Galaxy instance you used. </p>
+
+        <p>Please note that your feedback will be anonymous, but <strong>publicly visible</strong> (on this <a href="https://github.com/galaxyproject/training-material/issues/989">GitHub issue</a>).</p>
```
nsoranzo: Can we make this optional, maybe?

shiltemann: I have added a checkbox to the forms to make feedback confidential, in which case it will not be posted to the github issue

nsoranzo: Fantastic, thanks @shiltemann !

mblue9:
Automatically posting feedback to Github!..very cool!! (It never even occurred to me that that was possible 😄)

shiltemann:
thanks a lot @bebatut!

PR #997, opened 2018-09-13
Add custom 404 page

### LINKED ISSUE #996: implement custom 404
Links keep changing as content is being moved and matures.  They should stay the same, or be redirected but this is the first step to actually give the user a chance to find what they want instead of showing empty 404 github page with no links.

it seems to be possible with github pages: https://help.github.com/articles/creating-a-custom-404-page-for-your-github-pages-site/

bebatut:
To fix #996

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,0 +1,4 @@
+---
+layout: 404
+permalink: /404.html
+---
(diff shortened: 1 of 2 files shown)


martenson:
perfect, thanks!

PR #922, opened 2018-07-03
[WIP] limma-voom tutorial

mblue9:
I've a bit more work to do on this but aiming to finish it soon (in next couple of weeks). Submitting it now in case anyone has a chance to look in the meantime. If there are things I should change please let me know. Thanks in advance!

PR #1004, opened 2018-09-17
limma-voom tutorial (fastq to counts)

mblue9:
I'm not quite finished this but aiming to get it finished asap as I would like to use this on Oct 8th if possible. This is an accompanying tutorial to the limma-voom tutorial here https://github.com/galaxyproject/training-material/pull/922. I split it into 2 parts as otherwise it's long and am teaching it in 2 parts anyway, but am not quite sure how best to name the 2 parts.

This part shows how to convert the mouse fastqs to counts, along with some QC steps. It's similar to the ref-based RNA-seq tut, which I like very much and am trying not to duplicate here, but it's not quite the same as it:

- uses Collections rather than datasets in the analysis. All our RNA-seq users have multiple samples so I think it's better for them to use collections and so I'd prefer to teach them how to work with collections from the beginning.
- uses the Rule Uploader to show how it can be used to easily import data from URLs
- has some extra QC steps
- doesn't have a mix of paired-end and single-end reads, it's all single-end. Personally I think the ref-based drosophila dataset is good for showing how known confounders like different sequencing types can be incorporated, however mixing paired-end and single-end samples is not a typical thing our users do. I'd prefer to teach them how to do a more "typical" analysis first before adding complications.
- the differential expression part of this tutorial has a complementary R tutorial. I think that can be helpful for users who want to see how this type of analysis can be done in both Galaxy and R.

I still need to do more work on this (add in some screenshots, solutions, text, fixes etc) but any comments are very welcome!


mblue9:
>@mblue9 a general comment/question, would it be good to split this into a universal fastq-to-count tutorial, which then can be used in all different flavors?

@bgruening yes I'd be happy to do that when I get some time. If possible it would be great to have this tutorial available as a "hidden" tutorial first while I test it out. For that, do I just need to set `enable: false` in the top of the tutorial.md metadata?

bgruening:
> For that, do I just need to set enable: false in the top of the tutorial.md metadata?

Yes, that should work. See here: https://github.com/galaxyproject/training-material/blob/master/topics/epigenetics/tutorials/ewas-suite/tutorial.md

bgruening:
Good luck with your workshop!

mblue9:
Thanks a lot!!

PR #1019, opened 2018-09-25
fixed links to section requirements within all slides

bagnacan:
Links to internal training material, pointed to by entries in the "Requirements" section of many training's slides, were broken. This fixes the issue.

Example:
In the Quality control slides, [page #2](https://galaxyproject.github.io/training-material/topics/sequence-analysis/tutorials/quality-control/slides.html#2) links to the required introductory material using the following link
`https://galaxyproject.github.io/training-material/introduction/`
instead of
`https://galaxyproject.github.io/training-material/topics/introduction/`

PR #1027, opened 2018-09-26
Fix a couple broken links

erasche:
Thanks @nsoranzo, found in https://github.com/galaxyproject/training-material/pull/1019.

Also fixes favicon in slides.

As part of testing I noticed that the magic metadata stuff introduced in https://github.com/galaxyproject/training-material/pull/966 is broken with `--incremental`. Neat. It looks like `page.topic_name` is not set on subsequent incremental runs.


nsoranzo:
Thanks @erasche!

PR #1028, opened 2018-09-27
Add some og: tags for our pages

erasche:
Should allow for slightly nicer link previews when pasting into social media sites (e.g. twitter, whatsapp, telegram)


erasche:
The change in UX for people pasting links in chat:

grafik

The descriptions aren't great, but it wasn't possible to pull from the body of the tutorial since not everyone's first words are the introduction. Not sure how to solve that.

bebatut:
Thanks!

PR #1033, opened 2018-09-30
limma-voom: Add goseq and some edits to limma voom (diff exp)

mblue9:


bgruening:
That is super cool. We should probably have this as a standalone tutorial as well.

Thanks a lo!

PR #1035, opened 2018-10-01
Updated for new IDmapper option 'reduce collections'

foellmelanie:
IDmapper got a new option 'reduce collections' that has to be set differently from default for this tutorial


shiltemann:
thanks @foellmelanie! do you think we should update the workflow as well?

foellmelanie:
it was actually IDmerger not IDmapper. I will update the workflow as well.

PR #1039, opened 2018-10-02
Hands-on Analysis of Contaminants was made less confusing

Matthias313:
Point 3 and 4 were redundant, therefore Point 3 was removed.
Select tool was changed to the whole name to avoid misunderstandings, Point 4 was described more clear.
Questions are still valid and do not need further adaptions.


foellmelanie:
LGTM...

PR #1050, opened 2018-10-04
Mothur MiSeq SOP: add extra background information about mock communities

shiltemann:
Based on [this feedback](https://github.com/galaxyproject/training-material/issues/989#issuecomment-421968851) we received, I added some more background information about the rationale behind sequencing a mock community.

Thanks anonymous user for your constructive feedback!


bgruening:
Cool, thanks.

PR #1053, opened 2018-10-05
Support for post mortems (aka "Trainer experiences")

erasche:
Chose to name them something friendlier than we use in operations, the only place 'post mortem' remains is in the branch.

Added to topic pages as "For trainers"

grafik

The trainer experience list:

grafik

An example trainer experience:

grafik


erasche:
@tklingstrom how do you feel about this as well? The sort of thing you imagined?

TKlingstrom:
Yes, I really like it.

_layouts/training_experience.html
```diff
@@ -0,0 +1,107 @@
...
+                    </li>
+                </ul>
+            </div>
+        </div>
+    </nav>
+</header>
+
+<div class="container main-content">
+    <section class="tutorial">
+        <h1>{{ page.title }}</h1>
```
bebatut: Maybe adding the date also?

_layouts/training_experience_list.html
```diff
@@ -0,0 +1,110 @@
...
+                    </li>
+                </ul>
+            </div>
+        </div>
+    </nav>
+</header>
+
+<div class="container main-content">
+    <section class="tutorial">
+        <h1>{{ page.title }}</h1>
```
bebatut: There was no title displayed for me locally. Maybe we could have a default title there

erasche:
Probably in instructors is a good place. I'll add some documentation for that.

But @shiltemann has some really good points that we should discuss (maybe in person). The main one is: many people give many trainings and it would be hard to ask them to split it into 6 different documents (for each training), and it leaves them with no place for "overall stuff" unless they're integrating that into the "training todo list".

(Any offline discussion I will summarize discussion here for future reference.)

erasche:
This really feels sufficiently implemented by just reporting in the workshop  feedback issue #1452.

Closing this, we don't need it more structured or having a higher threshold to contribute feedback.

PR #1062, opened 2018-10-08
test fixing a couple urls

erasche:
Noticed that slides were failing due to `toolshed.g2.bx.psu.edu/&quot;`, trying with a `#` after it to see if I can get the slides checker to pass

Please do not merge until after slide check has run.

### DIFF
--- a/topics/dev/tutorials/toolshed/slides.html
+++ b/topics/dev/tutorials/toolshed/slides.html
@@ -82,12 +82,7 @@ contributors:
 ### Connect your Galaxy to a Tool Shed
 
 - Only for third party toolsheds (Main and Test by default)
-- In `config/tool_sheds_conf.xml`:
-```xml
-  <tool_sheds>
-    <tool_shed name="Galaxy main tool shed" url="https://toolshed.g2.bx.psu.edu/#" />
-  </tool_sheds>
-```
+- Add `<tool_shed />` block in `config/tool_sheds_conf.xml`
 - Restart Galaxy
 
 ---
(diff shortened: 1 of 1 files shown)


topics/dev/tutorials/toolshed/slides.html
```diff
@@ -83,9 +83,9 @@
 
 - Only for third party toolsheds (Main and Test by default)
 - In `config/tool_sheds_conf.xml`:
-```bash
+```xml
   <tool_sheds>
-    <tool_shed name="Galaxy main tool shed" url="https://toolshed.g2.bx.psu.edu/" />
+    <tool_shed name="Galaxy main tool shed" url="https://toolshed.g2.bx.psu.edu/#" />
   </tool_sheds>
 ```
```
nsoranzo: We can probably just drop this example, since it's just (part of) the default config file.

erasche: That works too. If you're fine with that I'll kill that

nsoranzo: Go ahead!

erasche:
The only other failing case in the slides test is the galaxy.pasteur.fr but that's *usually* online according to the data EU has.

nsoranzo:
> Please do not merge until after slide check has run.

It's green now, are you planning to remove it from `allow_failures:` ?

erasche:
We could? @shiltemann how do you feel? Is it stable enough?

shiltemann:
umm, sure, let's try it? 

all PRs will fail every time an external site goes temporarily down, but on the other hand right now it is easy to miss links that are really broken. And at least the slides link checking seems to be a lot more stable that the tutorial checker

@bebatut thoughts?

PR #1068, opened 2018-10-09
Test requiring internal links to pass

erasche:
Adds a new target for `check-html-internal` which is required to pass.

PR #978, opened 2018-08-24
Add some more JSON-LD Terms

erasche:
for accessibility stuff. Also standardise timeRequired in all tutorials and expose it in the json-ld.

PR #1071, opened 2018-10-09
Revert "Add some more JSON-LD Terms"

erasche:
Reverts galaxyproject/training-material#978

Due to https://github.com/galaxyproject/training-material/pull/1068#issuecomment-428251723


erasche:
@nsoranzo could you approve this so I can re-try it?

PR #1076, opened 2018-10-11
Limma-voom: some edits

mblue9:


bgruening:
Tutorial is still hidden, merging right away.

PR #1080, opened 2018-10-11
Fix workflow testing YAML

erasche:
Started trying to run the workflow tests against usegalaxy.eu only to find many contained yaml issues that caused immediate test failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,7 @@ check-slides: build  ## check the markdown-formatted links in slides
 check-yaml: ## lint yaml files
 	( $(ACTIVATE_ENV) && \
 	  find . -path "**/*.yaml" | xargs -L 1 -I '{}' sh -c "yamllint {}" \
+	  find . -path "**/*.yml" | xargs -L 1 -I '{}' sh -c "yamllint {}" \
 	)
 .PHONY: check-yaml
 
--- a/Makefile
+++ b/Makefile
@@ -115,8 +115,8 @@ check-slides: build  ## check the markdown-formatted links in slides
 
 check-yaml: ## lint yaml files
 	( $(ACTIVATE_ENV) && \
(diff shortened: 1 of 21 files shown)


shiltemann:
we could add the following to `.yamllint` to change the missing document start from warning to error?

```
document-start:
    level: error
    present: yes
```

bgruening:
@shiltemann I can be mistaken, but I think someone ( @bebatut ? ) removed the `---` intentionally a few weeks ago from the linter.

erasche:
@bgruening I don't see anything like that in the history https://github.com/galaxyproject/training-material/commits/master/.yamllint

Makefile
```diff
@@ -115,7 +115,8 @@ check-slides: build  ## check the markdown-formatted links in slides
 
 check-yaml: ## lint yaml files
 	( $(ACTIVATE_ENV) && \
-	  find . -path "**/*.yaml" | xargs -L 1 -I '{}' sh -c "yamllint {}" \
+	  find . -name "*.yaml" | xargs -L 1 -I '{}' sh -c "yamllint {}" \
```
nsoranzo: I thinkg we should also use single quotes for `*.yaml` to prevent shell expansion, i.e. `find . -name '*.yaml' | xargs ...`

erasche: I don't believe expansion happens when double or single quoted but I could be wrong? I was only under the impression that variables are expanded, not file paths.

shiltemann:
@bgruening yes, I relaxed some of the yaml rules a while ago (tho not specifically this one, it was always a warning) but if it gets in the way of your automatic workflow testing I thought it would be nice to require it? instead of you having to keep manually fixing it?

Makefile
shiltemann: ok, will do @nsoranzo :)

nsoranzo: Ah, @erasche is right, sometimes Bash is so confusing!

PR #1084, opened 2018-10-12
Fix for step 5 in exercise 2

Chris-Lowe-Integrity:


Chris-Lowe-Integrity:
Currently this causes fleeqtk_uv_ instead of version 1.3

topics/dev/tutorials/conda/slides.html
```diff
@@ -818,7 +818,7 @@
   * fleeqtk can be built using `make` and installed with `make install`
 3. Use `conda build` to build the recipe
 4. Add a `requirement` for this new package in the example tool.
-5. Run `planemo conda_install --conda_use_local fleeqtk_seq.xml` to install the package for Galaxy
+5. Run `planemo conda_install --conda_use_local fleeqtk=1.3` to install the package for Galaxy
```
bgruening: This should not happen if you give the requirement a version as seen here: https://github.com/galaxyproject/planemo/blob/master/project_templates/conda_answers/exercise_2/fleeqtk_seq.xml#L3

Chris-Lowe-Integrity: Hmmm something odd must of occurred then.  The version was specified in the xml.

nsoranzo:
All the hands-on details should be removed from the slides, it's covered on https://planemo.readthedocs.io/en/latest/writing_advanced.html#exercise-build-a-recipe , which I'm actually updating it to fix some issues:
- the example should have the requirement already included
- `conda skeleton` cannot be used because fleeqtk is a C program
- `planemo conda_install` does not work if planemo is installed through conda
- I'm adding support for `--conda_use_local` to `planemo test`

PR #1087, opened 2018-10-12
reword masthead help

martenson:
in order to avoid using possibly unknown terms

### DIFF
--- a/_includes/help.html
+++ b/_includes/help.html
@@ -7,10 +7,10 @@
             FAQ
         </a>
         <a class="dropdown-item" href="{{ site.gitter_url }}" title="Chat on Gitter">
-            Chat
+            Chat on Gitter
         </a>
         <a class="dropdown-item" href="{{ site.biostars_url }}" title="Ask on Biostars">
-            Ask questions
+            Ask questions on Biostars
         </a>
     </div>
 </li>
--- a/_includes/help.html
+++ b/_includes/help.html
@@ -7,10 +7,10 @@
             FAQ
         </a>
         <a class="dropdown-item" href="{{ site.gitter_url }}" title="Chat on Gitter">
-            Gitter
+            Chat
         </a>
         <a class="dropdown-item" href="{{ site.biostars_url }}" title="Ask on Biostars">
-            Biostars
+            Ask questions
         </a>
     </div>
 </li>


_includes/help.html
```diff
@@ -7,10 +7,10 @@
             FAQ
         </a>
         <a class="dropdown-item" href="{{ site.gitter_url }}" title="Chat on Gitter">
-            Gitter
+            Chat
         </a>
         <a class="dropdown-item" href="{{ site.biostars_url }}" title="Ask on Biostars">
-            Biostars
+            Ask questions
```
bgruening: What about `Ask questions (Biostars)` or like the title `Ask on Biostars`?

martenson: Why do you want 'biostars' there? The term won't help anybody who does not know it already. And people that know it do not need a link to it.

shiltemann: Gitter is also in the list and also might not be known to everybody, but both can be used to ask questions, so maybe we could do something like `Ask on Gitter (chat)` and `Ask on Biostars (forum)` to make clear what each of those options are?

martenson: Changing 'gitter' to 'chat' is a part of this PR. I think there is no need to push brand names here, but feel free to modify as you wish, you can write to my branch.

shiltemann: @martenson sorry, didn't look closely enough, thanks for changing that to chat, I was just worried we were somehow implying biostars was only place to ask questions, I don't mind removing the brand names

bebatut: I like the idea of knowing where I will be redirected. But also agree that we should make it clearer what is what

_includes/help.html
```diff
@@ -7,10 +7,10 @@
             FAQ
         </a>
         <a class="dropdown-item" href="{{ site.gitter_url }}" title="Chat on Gitter">
-            Gitter
+            Chat on Gitter
         </a>
         <a class="dropdown-item" href="{{ site.biostars_url }}" title="Ask on Biostars">
-            Biostars
+            Ask questions on Gitter
```
martenson: this is not correct

bebatut: lol

bebatut: it should be fixed

shiltemann:
thanks @martenson!

PR #1094, opened 2018-10-16
Contributing tutorials: some small tweaks and fixes

shiltemann:
updated some links and instructions, added instruction to use `planemo training_init` with a locally downloaded workflow `.ga` file


topics/contributing/files/example-workflow.ga
```diff
@@ -0,0 +1,368 @@
+{
```
bebatut: Why not putting it in `topics/contributing/tutorials/create-new-tutorial/workflows`?

bebatut:
Looks good to me :smile:

topics/contributing/files/example-workflow.ga
shiltemann: because that would have made way too much sense ;) thanks, moved it :)

Issue #656, opened 2018-01-10
implement google analytics?


martenson:
Galaxy uses google analytics to gather stats and insights from some of the websites. Is the training community interested in these stats for training-materials?

bebatut:
What are the statistics? What should we do to get them?

martenson:
@bebatut the stats are currently not being gathered for training pages. We would need to add a tracking code to the site to enable it.

The stats cover many things like user numbers, sessions, browsing flow, country/city of visitors, site loading speeds etc.

bebatut:
@martenson any idea how to do that?

martenson:
@bebatut The short and fast way is to just include the javascript tracking code for google analytics on every page. The long and slow way is to use something self-hosted like https://matomo.org/ (used to be piwik).

PR #1100, opened 2018-10-18
implement ga

martenson:
implements https://github.com/galaxyproject/training-material/issues/656 using Google Analytics


martenson:
Is there an instance specific config different from the standard _config for jekyl? I am thinking about how to make the tracking/account number configurable so people don't spin up their dev instances with our tracking code.

martenson:
yeah, ready to test

martenson:
data flowing in 🎉

Issue #1056, opened 2018-10-05
Training stories/philosophies for training handbooks


bebatut:
As discussed during the Instructor meeting #1009, it would be nice to develop some handbooks for instructors and workshop organizers with pedagogical/technical recommendations/checklist.
I created the instructor topic #1051 to store these handbooks. 

What would be the targets for you? 
I had in mind targets: workshop organizers and instructors (they can overlap for sure). Do you think it makes sense?
We could have one handbook for each target.

We thought about taking inspiration from [the Carpentries handbook](https://docs.carpentries.org/topic_folders/hosts_instructors/index.html), linking to it for general recommendations and extending to Galaxy specific recommendation.
Do you have some recommendations that we should absolutely have?

bebatut:
We started the handbook and we realized something essential is missing: training stories/philosophies

We are all giving training in a different way (own pace or step by step, using only slides/training or alternating between them, showing only Galaxy, etc), depending on many factors and how we learn it. It would be nice collect all these stories/philosophies here and we will add them in the handbooks later on. It could help instructors to identify what could fit the best for them.

bebatut:
I will then start with my training story/philosophy.

When I am giving a training using Galaxy (e.g. ref-based RNA-seq), I usually give a 10-20 min introduction to RNA, RNA-seq using the slides that in the transcriptomics topic.

Then I switch to the tutorial. I show the participants where to find the tutorial on the training material website. I introduce the tutorial story: explain that we want to reproduce an analysis from a paper, introduce this paper, talk about the datasets/samples we will use.
 
After that, I open Galaxy and I go step by step through the analysis with them: 
- introduce the step: give an idea why we need this step
- talk a bit about the "theory" of this step
   - using the schemas in the tutorial to explain for example RNA-seq mapping
   - using legos to explain mapping (and let them try to map lego reads on a reference genome)
- open the tools in Galaxy and going through the parameters with them, explaining the main parameters and their potential effect
- run the tool
- inspect the generated output with them, insist on each new format
- ask the questions that are in the question boxes and/or to do an extra step (identify a new tool, identify and use different parameters, etc)

At some point, I introduce the step and explain the purpose of the step, but let them do it by themselves.

At the end of the tutorial, I spent some time to go through again each step, asking them again why this step, what were the inputs/outputs and in which format, repeating again :smile: 

Here is the way I usually give a data analysis training using Galaxy.

shiltemann:
[...]

PR #1101, opened 2018-10-19
Training handbook

bebatut:
Hi,

As we discussed in #1056, we started to collect recommendations/checklists for organizing and running a workshop. 

It is mostly based on our workshop. So please add/comment what is different in your workshops.

Also, we are all giving training in a different way (own pace or step by step, using only slides/training or alternating between them, showing only Galaxy, etc), depending on many factors and how we learn it. We will collect them in the issue #1056 (I added mine :smile:) and then we will add them in the handbooks there afterwards. It could help instructors to identify what could fit the best for them. 

Thanks!


erasche:
todo(@erasche): add note that food should be labelled for any allergens / contents.

PR #1103, opened 2018-10-19
remove screenshots from genome-annot/annot-w-prokka

polkhe:
- genome-annot/annotation-with-prokka
@bebatut


bebatut:
@AnnaSyme, @Slugger70 do you want to have a look at the changes?

PR #1107, opened 2018-10-19
Correct naming of dataset to join with DESeq2 output

pvanheus:
Addresses the error reported in #989


nsoranzo:
Thanks @pvanheus !

PR #1111, opened 2018-10-23
Ref-based RNA-seq: fix leading spaces in URLs

shiltemann:
the zenodo urls had some leading spaces causing upload to fail if copy-pasted as-is


bgruening:
Merci!

PR #1112, opened 2018-10-24
Fixes #1086

### LINKED ISSUE #1086: materials are linked using unsecure protocol
Whatever topic I click on from (secured) page https://galaxyproject.github.io/training-material/ will be accessed only with `http` including any links to subsequent content. This seems unfortunate given all the pages support https. We should promote only https links. 

We should also redirect automatically if possible.

erasche:
It looks like #1086 was because of the lack of a trailing slash. I'm not sure why that caused the insecure redirect, but it's an easy fix.

First few requests are normal, last request is after going back to index and manually editing link.

grafik


martenson:
works great, thanks mate @erasche !

PR #1122, opened 2018-11-01
[martenson-patch-3] Google analytics ID via travis variable

shiltemann:
@martenson I think this should work to prevent people who spin up a dev instance from using our ga ID, we set an environment variable in travis and use the [jekyll-environment-variables plugin](https://github.com/ajwhite/jekyll-environment-variables)

xref: https://github.com/galaxyproject/training-material/pull/1100


martenson:
thank you @shiltemann !

PR #1125, opened 2018-11-05
Add an icon for repeat parameters

shiltemann:
While reviewing a PR that included a tool with a lot of repeat parameters, I thought it would be a lot clearer if these are prefixed with a `plus` icon, as they are in Galaxy:

image

so I added such an icon


erasche:
Can't wait to use this in Circos!

bgruening:
Thanks @shiltemann!

bebatut:
I update Planemo to use it now in workflow generation

Issue #456, opened 2017-08-03
Concernings about the ChIP-Seq tutorial 'Identification of the binding sites of the T-cell acute lymphocytic leukemia protein 1 (TAL1)'


joachimwolff:
We have taught the tutorial about ChIP-seq: [Identification of the binding sites of the T-cell acute lymphocytic leukemia protein 1 (TAL1)](https://galaxyproject.github.io/training-material//topics/chip-seq/tutorials/tal1-binding-site-identification/tutorial.html) as it is published. During the preparation of the session some questions appeared that we can not answer with the given material. 

1. If you compare the file sizes of all G1E-cell types and all megakaryocytes cells you see that all G1E samples have a file size between 30 - 60 MB, except of G1E_input_R2 which is having around 400 MB. For megakaryocytes it looks the same: all samples have a file size of around 400 MB except of megakaryocytes_input_R1 which is having a file size of around 60 MB. Are these two files swapped around by mistake?

2. Step 5: plotFingerprint: 
a ) G1E_input_R1 and G1E-TAl1_R1 are having the same 'enrichment'. (blue and green lines). We think this is not a good example for teaching. The control and the treatment should differ more.
![image](https://user-images.githubusercontent.com/13435136/28916678-5153d7b6-7843-11e7-8a67-8a57c3389c17.png)
b) The megakaryoctyes do not show a distinction between control and samples.
![image](https://user-images.githubusercontent.com/13435136/28916791-b47db212-7843-11e7-8c3c-1177165392eb.png)

3. Compute Matrix and Heatmap plotting
a) For Step 10: ComputeMatrix for heatmap plotting:
Step 10.3. "Under “Score file”, select the bigWigs (log2 ratios from bamCompare)." indicates not exactly which files (samples) we should select for 'Score file'. Given the image below we assumed to choose all samples for the score file i.e. the results of bamCompare for `G1E_input_R1+G1E_TAL1_R1`;  `G1E_input_R2+G1E_TAL1_R2`;  `mega_input_R1+mega_TAL1_R1`;  `mega_input_R1+mega_TAL1_R1`;. But for us it does not make sense to use one region for one cell type and to use the score files for all cell types. We think it would make more sense to use the same cell types for the region and score file.
b) Assuming the setup is correct, we would expect that for the region file for cell type G1E that the enrichment would be seen for this cell type too. But in the given heatmap plot there is no enrichment for G1E compared to the megakaryocytes. 

[...]

PR #1139, opened 2018-11-15
Disable tutorials without Zenodo link and some chip-seq

bebatut:
And also add a missing Zenodo link

### DIFF
--- a/bin/validate-frontmatter.rb
+++ b/bin/validate-frontmatter.rb
@@ -24,7 +24,7 @@ errs = []
 def skip_disabled(data, fn)
   # If there's an 'enable' key and it is one flavor of 'false', then, exit
   # immediately without testing.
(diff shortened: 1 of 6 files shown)


MoHeydarian:
Are there plans to re-enable the disabled chip-seq tutorials? I am happy to make updates to the Tal1 tutorial, I was planning on teaching from it this week.

bebatut:
It was confusing for people to know which of the 3 tutorial using. We kept the new ChiP-seq tutorial is available, which has the same steps as the other 2, with more explanation.
We disabled the Tal1 tutorial because there were several issues that were never solved: https://github.com/galaxyproject/training-material/issues/456, but it should still be accessible online (https://galaxyproject.github.io/training-material/topics/chip-seq/tutorials/tal1-binding-site-identification/tutorial.html) if you need it.

MoHeydarian:
Thanks for the clarification and the disabled link.

PR #1153, opened 2018-11-19
Add Maria Doyle as maintainer for transcriptomics topic

shiltemann:
Thanks a lot @mblue9!


mblue9:
Yep go for it @erasche ! And please add me here https://github.com/galaxyproject/training-material/pull/1154 (great idea! I'll need that at the moment)

PR #539, opened 2017-09-19
Add CODEOWNERS file for automatic PR review requests

bebatut:
As suggested by @erasche in #489, this PR add a `CODEOWNERS` to have automatic PR review requests of the maintainers of each topic

Can you check if it is ok with you?

PR #1154, opened 2018-11-19
Add CODEOWNERS file

erasche:
I would like to be automatically pinged things I've volunteered to be a maintainer of. Maybe I miss something if it is a week where I'm not watching too closely and I wouldn't want anyone's contributions to go unnoticed.

This is an opt-in system, right now I just do it for me since I worry I miss something.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,7 +2,7 @@
 # https://blog.github.com/2017-07-06-introducing-code-owners/
 
 
-topics/admin/*       @erasche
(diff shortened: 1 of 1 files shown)


erasche:
I've added some more names from https://github.com/galaxyproject/training-material/pull/539 since that was not successful then. If you are interested in joining, please add yourself. (You can use the "suggest" mode if you cannot just add to my branch.)

Slugger70:
I'm happy to be added to assembly

PR #1156, opened 2018-11-20
Add internal link check in make check command

bebatut:
to make it possible to locally check....


erasche:
(Merging without tests as nothing tests the makefile)

PR #1158, opened 2018-11-20
Proteogenomics: fix workflow folder names

shiltemann:
workflow folder was missing an `s` at the end


subinamehta:
Thanks @shiltemann for fixing this!

PR #1169, opened 2018-11-30
peak2genes tiny update

erasche:
- use the full tag since only clarified in the note which someone could skip that you need the `#` sign
- the bar chart doesn't seem to include the icon for me?


erasche:
An additional question is that "bar diagram" which is mentioned in the training has quite some results? Maybe it should be changed to be something more explicit?

grafik

bgruening:
@erasche for the plotting, this was "free-work" I normally just say ... go plotting and play around for 5-10 min.

PR #1173, opened 2018-12-05
final peaks2genes changes

erasche:
ok, think I'm finally happy :)

### DIFF
--- a/snippets/import_via_link.md
+++ b/snippets/import_via_link.md
@@ -22,5 +22,6 @@
 >    > {% else %}
 >    > * **Close** the window
 >    > {% endif %}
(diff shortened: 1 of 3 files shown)


snippets/import_via_link.md
```diff
@@ -22,5 +22,6 @@
 >    > {% else %}
 >    > * **Close** the window
 >    > {% endif %}
+>    > By default, Galaxy uses the URL as then name, so rename the files with a more useful name.
```
bgruening: s/then/the/

erasche:
```suggestion
>    > By default, Galaxy uses the URL as the name, so rename the files with a more useful name.
```

PR #1179, opened 2018-12-12
Cummerbund workflow update

bagnacan:
With the present update, the workflow reflects the contents of the latest edits from the hands-on material (see [PR 1083](https://github.com/galaxyproject/training-material/pull/1083)).

PR #1180, opened 2018-12-12
Reformat workflows + Catch future formatting issues

erasche:
~~1. Reformat all workflows nicely so that future diffs are more readable (until we switch to yaml).~~
2. Add a validator to check that workflows are 1) valid json, ~~2) keys sorted.~~

The second check is specifically because an invalid workflow was merged in #1179 which I only caught by accident.

Maybe we should also remove workflows like [this one](https://github.com/galaxyproject/training-material/pull/1180/files#diff-4a3639255cafc0f2eb372a39585fcfcd) which don't actually have any steps?

Edit: This PR will fail until #1179 is corrected.


erasche:
@bebatut this fails due to some travis issue it looks like? Any ideas?

PR #1182, opened 2018-12-13
Allow checking specific servers for inability to support a training

erasche:
@shiltemann and I were wondering why EU was marked as not supporting her mothur miseq tutorial. I've updated the annotation generation script to support checking a single server:

```console
$ python bin/add_galaxy_instance_annotations.py \
  --server https://usegalaxy.eu --verbose
...
DEBUG:root:MISSING:none:metagenomics/mothur-miseq-sop: toolshed.g2.bx.psu.edu/repos/dcorreia/newick_display
...
```

And can easily see that this tool should have its ID updated to the new IUC tool. (Or we can find out other things we're missing to support different trainings.)

### DIFF
--- a/bin/add_galaxy_instance_annotations.py
+++ b/bin/add_galaxy_instance_annotations.py
@@ -5,7 +5,7 @@ import argparse
 import yaml
 import requests
 import io
-import pandas as pd
+import csv
 import multiprocessing.pool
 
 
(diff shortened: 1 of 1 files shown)


shiltemann:
thanks a lot @erasche, very useful :)

PR #1205, opened 2019-01-27
Ansible for installing + using CVMFS

erasche:
ping @Slugger70, imported your work in the first commit and made it more training-materials compatible in the second.


erasche:
Ready for review @Slugger70

PR #1210, opened 2019-01-27
minor updates to upgrading slides

martenson:


topics/admin/tutorials/upgrading/slides.html
```diff
@@ -135,16 +142,14 @@
...
-Since January 2018, the `dev` and pre-release (frozen) Galaxy branches do not
-contain client build artifacts (e.g. JavaScript bundles in `static/`).  For
-`dev`, Galaxy will automatically build the client locally, as needed, though if
-you need to do so manually (e.g. to include local changes), the following steps
-should do it:
+* Since 18.09 not even the `release_*` branches include the client build. Instead Galaxy will check for local JavaScript changes and perform a build if neccessary (including the build dependency installation).
 
-1. install `yarn`
-2. run `make client-production`
+* You can always run `make client-production` to force the client build (but this assumes you manage the dependencies).
```
erasche: should it be `sh common_startup.sh` instead? Since make doesn't install nodenv/yarn.

martenson: > but this assumes you manage the dependencies

If they want to force a build they should manage it I think. a.k.a. you only force it when something goes wrong, right?

erasche: yeah ok

PR #1211, opened 2019-01-27
Import all other admin slides

erasche:


topics/admin/tutorials/toolshed/slides.html
```diff
@@ -0,0 +1,224 @@
+---
+layout: base_slides
+logo: "GTN"
+
+title: "The Galaxy Tool Shed"
```
nsoranzo: We already have https://galaxyproject.github.io/training-material/topics/dev/tutorials/toolshed/slides.html , not sure if these slides are needed.

erasche: @martenson what do you want to do?

martenson: I plan to merge them in for now, we can reconvene and merge the materials after the training is over. That 80 slide deck is not something easy to grasp or follow, I'd rather have smaller one-purpose decks in the future.

PR #1251, opened 2019-02-07
Automatic archival builds of repository tags

erasche:
This adds a script which will check out all git tags in the repo and automatically build them if they haven't previously been built.

The results are stored http://galaxy-training.s3-website.us-east-1.amazonaws.com/ which will be proxied in the near future, so it will become something like:

```
training.galaxyproject.org/<normal website>
training.galaxyproject.org/archive/2019-01-01/<archived website>
```

Tags are fairly lightweight and easy to add as needed per-training or per-month. I propose that it is done automatically per-month, but this will be done in a separate PR. I have added an experimental tag that will show up when this is merged.

The versions currently on the S3 bucket are built from the tags in my fork and will disappear when we merge this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,4 +87,5 @@ after_success:
       make annotate
       make build
       bin/publish
+      bin/publish-archive
     fi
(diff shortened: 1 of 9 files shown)


nsoranzo:
Thanks for working on this, nice to see implemented the ideas discussed at the Biohackathon near Paris!

shiltemann:
cool! :tada:

PR #1256, opened 2019-02-08
Update for proxying

erasche:
Training material will now be accessible from https://training.galaxyproject.org/ with /topics directly from the root. This is currently partially live, in an attempt to cause the minimum amount of disruption.

I've coded the assets to use the full site url so the old github pages will still render correctly. If this wasn't done, the assets would attempt loading from `/assets` instead of `/training-material/assets` and be unavailable and break.

I've also fixed some other incorrect uses of site.url instead of site.baseurl, one from an admin tuto that would break if you ran the command.

Lastly moves archive index page into /archive so that will load properly too.

With this PR merged, everything should be done, https://training.galaxyproject.org/ should have all content and browsing around on there should stay on that site. Browsing from bookmarked github.io links will, on-click, redirect to the galaxyproject.org version. Old versions will be available under https://training.galaxyproject.org/archive/

### DIFF
--- a/topics/admin/tutorials/tool-management/tutorial.md
+++ b/topics/admin/tutorials/tool-management/tutorial.md
@@ -35,7 +35,7 @@ This tutorial will introduce you to one of Galaxy's associated projects - [Ephem
 
 # Background
 
-You are an administrator of your lab's Galaxy. A colleague has approached you with a request to run a specific [Galaxy workflow]({{ site.url }}/topics/sequence-analysis/tutorials/mapping/workflows/mapping.ga) on the lab's data. In order to run this workflow you have to accomplish several substeps first. You will need to:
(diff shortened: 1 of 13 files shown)


topics/admin/tutorials/tool-management/tutorial.md
```diff
@@ -77,7 +77,7 @@ Let's try running this on a real worfklow.
 >
 > 1. Download the mapping workflow:
 >    ```console
->    $ wget {{ site.url }}/topics/sequence-analysis/tutorials/mapping/workflows/mapping.ga
+>    $ wget {{ site.url }}{{ site.baseurl }}/topics/sequence-analysis/tutorials/mapping/workflows/mapping.ga
```
bgruening: That looks suspicious ...

erasche: It is correct. URL is just protocol + fqdn. Baseurl is only the /... part.

erasche:
Need to double check having trailing slash in URL but not baseurl

PR #1262, opened 2019-02-12
Debruijn wf tour

polkhe:
@bebatut 
- workflow
- tour


bebatut:
@slugger70 do you want to review this workflow and this tour?

topics/assembly/tutorials/debruijn-graph-assembly/tours/tour.yaml
```diff
@@ -0,0 +1,241 @@
+﻿id: debruijn-graph-assembly
```
bebatut: `---` is missing at the beginning (Travis fails)

PR #1265, opened 2019-02-13
Training clipseq new version

Florian-H-Lab:
- I changed the training workflow and update the genome version from hg19 to hg38 and the version of RCAS to 1.5.4


bgruening:
Do you mind updating the workflow testing repo?

PR #1267, opened 2019-02-13
Goenrichment

dsobral:
Flattened and unzipped the zenodo dataset.


bedroesb:
Thank you for this fast response @dsobral !

PR #1269, opened 2019-02-18
Installing tools from workflows

bedroesb:
A workflow.ga file in the workflow directory will be used to install the necessary tools.
At this moment the tools are installed in a section/tool_panel_section_label named after the tutorial name, but this can also be the topic name, suggestions?

I am at this moment testing all the topics for correct container buildup.


bedroesb:
@bebatut Maybe it is also necessary to change the planemo training_init module so it generates Docker files in the latest format?

bin/workflow_tools_install.py
```diff
@@ -0,0 +1,22 @@
+import os
+import yaml
+import sys
+
+HEAD = """
+---
```
bgruening: @bedroesb if you have time I think a better way would be to integrate this deeply into shed-install (ephemeris project). Maybe even without conversions. Also please note that the workflow format will change (optionally) to be a yaml file. I'm in favor of going all with this new format and do not include any support for the old ga files here in the training.

bgruening:
Nice improvement. Thanks @bedroesb!

Maybe we pin the training containers to a specific Galaxy version at some point, but for the moment I think it's ok to go with the latest.

PR #1273, opened 2019-02-20
Adding new mass spectrometry imaging tutorial

foellmelanie:
Paste zenodo url does not work for composite imzML files, needs to be donwloaded locally as written in the tutorial


bgruening:
@bedroesb this data library is interesting. It's a public that was re-used not created by @foellmelanie. Means more files are included than needed. Moreover, the files included are a so-called composite-datatype. And to my knowledge cannot be (yet) uploaded via ephemeris.

bedroesb:
@bgruening  I found a similar problem in the metabolomics tutorial lcms-data-import where ephemeris could import the data, but got stuck in some kind of loop where it kept asking for requests?  (these are mzXML files)

PR #1275, opened 2019-02-21
 Updated install_tutorial_requirements.sh + minor fixes

bedroesb:
* Install_tutorial_requirements.sh now installs tools from workflows
* Extensions of LC-MS - Data Import data-library fixed
* Used unzipped DOI of zenodo for iwtomics tutorial
* Deleted empty workflow files
* Changed revision of tools NCBI blast +, busco, custom_pro_db and goseq to latest version in workflows


topics/proteomics/tutorials/mass-spectrometry-imaging-loading-exploring-data/data-library.yaml
```diff
@@ -15,9 +15,9 @@ items:
       items:
       - url: https://zenodo.org/api/files/463f6b73-0c42-4be4-8b52-181a8e093abd/mouse_kidney_cut.ibd
         src: url
-        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
+        ext: ibd
         info: https://doi.org/10.5281/zenodo.1560645
       - url: https://zenodo.org/api/files/463f6b73-0c42-4be4-8b52-181a8e093abd/mouse_kidney_cut.imzML
         src: url
-        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
+        ext: imzML
```
bgruening: shouldn't this be imzml?

shared/datatypes.yaml
```diff
@@ -17,20 +17,25 @@ fastq.gz: fastqsanger.gz
...
 gtf: gtf
 gz: gz
 html: html
+ibd: ibd
 idat: idat
+imzML: imzML
 metadata: tabular
 mgf: mgf
 mzml: mzml
+mzXML: mzXML
```
bgruening: mzXML: mzxml?

shared/datatypes.yaml
```diff
@@ -17,20 +17,25 @@ fastq.gz: fastqsanger.gz
...
 fq: fastqsanger
 gbk: gbk
 gff: gff
 gff3: gff3
 gtf: gtf
 gz: gz
 html: html
+ibd: ibd
 idat: idat
+imzML: imzML
```
bgruening: imzML: imzML

bedroesb:
I am still checking if the transcriptomics container can be built correctly, because the version of goseq that was being used was broken. I hope that the latest version of goseq that I use now will not effect the workflow.

shared/datatypes.yaml
```diff
@@ -17,20 +17,25 @@ fastq.gz: fastqsanger.gz
 fastq.bz: fastqsanger.bz2
 fastqsanger: fastqsanger
 fastqsanger.gz: fastqsanger.gz
-fastqsanger.bz: fastqsanger.bz2
-fna: fasta
+fastqsanger.bz: fastqsanger.bz
 fa: fasta
+fb: tabular
```
bgruening: @bedroesb do you know form where this is coming? what is `fb`? Is that just a typo?

bedroesb: The `.fb` file is found in the goenrichment tutorial in transcriptomics. See [zenodo](https://zenodo.org/record/2565417)
While searching for the extension I found some data-library files that were named incorrectly (see latest push).

bgruening: I see this refers to FlyBase probably. I don't think we should add this here instead we should fix this by hand and ask @dsobral to change this to tabular the next time he changes the data.

bedroesb: I delete the .fb extention?

dsobral: Yes, actually the file is a gaf file, it is not even tabular...
At some point it was renamed and the file format was "lost in translation".

Sorry, I need to change it to .gaf  
This will also imply changing a few things at different points in the process.

Answering @bedroesb the .fb extension, at least in what regards this dataset, should not be included.

Best,
Daniel

bedroesb: ok thank you. Done

PR #1283, opened 2019-02-27
Fix to transcriptomics/ref-based tutorial

mtekman:
Fix to some tool names which have duplicates such as Cut and Concat


mtekman:
these changes might not be necessary across all Galaxy instances.

Please re-open when relevant

PR #1300, opened 2019-03-05
Make the tip boxes expandable

bebatut:
A request from @wm75 

Current state:

screenshot 2019-03-05 at 08 45 31

With the PR:

screenshot 2019-03-05 at 08 42 21

Let me know what you think about that


shiltemann:
oh yeah, I like it :) looks cleaner and people can treat it more like little exercises, try to do it themselves first before looking at "solution" :+1:

erasche:
We should probably review all other uses of tip boxes before we merge? After reviewing my tutorials it was clear that I misused them in a bunch of places when comment/details boxes are better :(

wm75:
Me too, occasionally, I assume. OTOH, no tutorial will be broken by this (just may become less convenient to follow in single places). So maybe its ok to fix inappropriate uses of tip boxes as we come across them?

erasche:
Sure @wm75, I just thought it might make sense to make a quick survey/check of the topics of interest to us? I didn't mean to imply that needed to happen in this PR, just that that's a downstream result.

wm75:
Completely agree on that :+1:

shiltemann:
yes, I also abuse the tip box for background info boxes in my tutorials and would like to keep those expanded by default, so will change the box type for those, and agree it makes sense to review any other tutorials this affects as well

erasche:
I'll look at admin and dev sections on thursday, I'll have some time on the train.

bebatut:
as you like. @wm75, @shiltemann what do you prefer?

wm75:
Obviously, I'll take care to use Tips correctly in the somatic variant calling tutorial I'm currently writing.
Then after next week, I need to update tutorials making use of gemini anyway so I can also check them. I would also volunteer to go through everything in variant analysis then.

shiltemann:
@wm75 sounds good, thanks a lot! ..then let's just add to this PR? that way we don't leave any tutorials in sub-optimal state in the meantime ..I should have time to add updates for my tutorials in the coming days :+1:

erasche:
I've made some progress before lunch. I've also replaced some things that were tips with their equivalent snippets, but this was mostly to clear up the results of my grepping.

erasche:
Between @shiltemann and myself we've fixed the rest of the boxes :)

bebatut:
Thanks a lot @shiltemann, @wm75 and @erasche for the work and the feedback

PR #1305, opened 2019-03-09
Fix to Pre-processing

mtekman:
Minor fix to links (Zenodo record updated thanks to fix from their team)
Input names fixed (Thanks to J Kim for spotting these)


bgruening:
Thanks @mtekman!

PR #1316, opened 2019-03-14
Fix a couple of bugs

erasche:


bgruening:
Thanks, Helena and Anthony!

PR #1263, opened 2019-02-12
Fix requirement display on slides and support follow up for slides

bebatut:
It should fix requirements for slides to follow the new way to represent requirements, and implement the possible follow up (as for hands-on tutorial)

screenshot 2019-02-12 at 16 43 03
screenshot 2019-02-12 at 16 45 43

I also fix the metadata for some slide decks not following the new requirement declaration

PR #1317, opened 2019-03-15
Fix link to external tutorial for required or follow-up tutorials

bebatut:
Spotted by @nsoranzo in #1263


nsoranzo:
Errors are still there unfortunately: https://travis-ci.org/galaxyproject/training-material/jobs/506718736#L2107

Also, this still doesn't allow for using slide decks as requirements.

bebatut:
I will investigate more on Monday

bebatut:
The error should be fixed.
@nsoranzo I do not understand your 2nd point

PR #1327, opened 2019-04-04
Unflattened one-line workflows

bedroesb:
Changed all oneline workflows to the unflattened JSON one using 
```
cat wf.ga | jq . -S > out.ga
```
from https://github.com/usegalaxy-eu/workflow-testing

This way small changes to the workflow can be easily compared.


bedroesb:
This is again one step towards deleting the tools.yaml files. After this I will change some workflows in a way that `workflow-to-tools` from ephermeris is able to detect all the tools.

PR #1332, opened 2019-04-10
Data fetched at docker run to decrease container size

bedroesb:
At image build, `docker build -t containername -f topics/topicname/docker/Dockerfile .`, only the data-library.yaml files are merged, the way it used to, but the data itself is not downloaded.

When you use the normal run command  `docker run -p "8080:80" containername` the container will run without the example data.
When the command  `docker run -p "8080:80" containername -d` is used, the corresponding data libraries of that topic will be downloaded and accessible within the training environment. 

This reduces the size of the images enormously, enabling them to be uploaded to docker hub or quay.io


bedroesb:
if this idea will be implemented, we should also change planemo to generate the new docker files with the new 
```
ADD bin/datalibrarydownload.sh /datalibrarydownload.sh
ENTRYPOINT ["/datalibrarydownload.sh"]
```
lines

bedroesb:
@bgruening What do you think of this implementation? Or would you still prefer to incorporate this function inside the galaxy-stable base image.

bedroesb:
@bgruening I checked the building of several topics, and it works like a charm. Only concern: when I wanted to build the containers in windows, I had some line ending problems in the bash scripts. (although on github LF is used ...  ). But that is a windows thingy.. I think it can be merged. Biggest container now is 28 gig for transcriptomics.

topics/epigenetics/docker/Dockerfile
```diff
@@ -18,4 +19,7 @@ ADD topics/epigenetics/tutorials/ /tutorials/
 # install everything for tutorials
 ADD bin/docker-install-tutorials.sh /setup-tutorials.sh
 ADD bin/mergeyaml.py /mergeyaml.py
+ADD bin/datalibrarydownload.sh /datalibrarydownload.sh
 RUN /setup-tutorials.sh
+
+ENTRYPOINT ["/datalibrarydownload.sh"]
```
bgruening: That is a little bit nitpicking, but could you call this script `data_libarary_download.sh`. Its a little bit more readbale.

bedroesb:
@bgruening Done !

bgruening:
Merci!

bgruening:
@bedroesb you might want to inform Frederik about this. He probably wants to put this in some fancy GDoc for ELIXIR.

PR #1335, opened 2019-04-12
[16S rRNA with mothur] clarify pair naming step and importance

shiltemann:
better stress the importance of using the correct pair names in collection creation step


bgruening:
:) thanks!!!

PR #1326, opened 2019-03-31
Tiding up tutorial.md

kpoterlowicz:

PR #1336, opened 2019-04-12
ewas suite image upload

kpbioteam:
I would like to upload this two images in response to following request https://github.com/galaxyproject/training-material/pull/1326


kpbioteam:
updated tour.yaml and, workflow @bgruening

PR #1337, opened 2019-04-15
Added some extra information

subinamehta:


shiltemann:
thanks for the updates @subinamehta!

I notice you often have a step "inspect the resulting files" ..It could be really great if you could add some questions here, what do you want learners to understand about the files (doesn't have to be in this PR, but just a general suggestion :))

PR #1342, opened 2019-04-23
remove a duplicated lesson

gmauro:
I have merged the two lessons on Heterogeneous compute resources


topics/admin/tutorials/heterogeneous-compute/slides.html
```diff
@@ -136,3 +136,7 @@
     * [https://github.com/galaxyproject/pulsar](https://github.com/galaxyproject/pulsar)
 * Pulsar Ansible
     * [https://github.com/galaxyproject/ansible-pulsar](https://github.com/galaxyproject/ansible-pulsar)
+
+# Exercise
```
erasche: Please remove this section, we don't want to link to outdated training resources.

topics/admin/tutorials/heterogeneous-compute/slides.html
```diff
@@ -2,13 +2,13 @@
 layout: base_slides
 logo: "GTN"
 
-title: "Using Heterogeneous Compute Resources"
+title: "Running Jobs on Remote Resources with Pulsar"
```
erasche: While they can be remote systems they don't have to be.

I'd prefer to leave the original title / first section heading, but if you can think of a good way to say "heterogenous and/or remote", we can change to that?

gmauro: I think Pulsar should be on the title. What about "Using Pulsar to handle heterogeneous remote resources"?

gmauro: or better "Using Pulsar to handle heterogeneous compute resources"

erasche: > Using Pulsar to handle heterogeneous compute resources

sounds great!

topics/admin/tutorials/heterogeneous-compute/slides.html
gmauro: ok

PR #1344, opened 2019-04-23
Try to fix makefile

erasche:
> miniconda changed, not dropping the slash, just now has an additional folder named condabin, which contains just the executable conda, and the rest like activate are still in regular bin folder

solution finally figured out by @shiltemann <3

Apparently makefile doesn't have a parent directory function (and everything online just implements their own), so shelling out.


shiltemann:
thanks @erasche and @bebatut

PR #1349, opened 2019-04-24
Add pip as requirement in environment.yml

blankenberg:
Fixes warning during `make create-env`


shiltemann:
thanks @blankenberg :)

PR #1352, opened 2019-04-26
Standardise contributor lists everywhere

erasche:
Slides and topics now use the same icon+name, as in the byline of the tutorials.

image

image

### DIFF
--- a/_includes/contributor-badge-list.html
+++ b/_includes/contributor-badge-list.html
@@ -1,0 +1,6 @@
+{% if include.contributors %}
+{% capture people %}
+{% for id in include.contributors %}, {% assign name = contributors[id].name | default: id -%}{% include _includes/contributor-badge.html id=id name=name %}{% endfor %}
+{% endcapture %}
+{% endif %}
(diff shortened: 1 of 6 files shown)


shiltemann:
great, thanks @erasche!

nsoranzo:
I think this is not working for tutorials, see e.g. the bottom of vhttps://training.galaxyproject.org/training-material/topics/introduction/tutorials/galaxy-intro-short/tutorial.html

shiltemann:
oops, thanks @nsoranzo! I only looked at the list at the top, will fix :)

erasche:
Huh, ok, we'd seen that before somewhere but didn't see that when I was building locally, odd.

erasche:
Ok, that re-used a variable definition that I thought was more local than it was. https://github.com/galaxyproject/training-material/pull/1352/files#diff-397e461c49d6c29f90c7fc3ab7839e4bR6 is overwritten below, that explains the issue.

I have a PR to correct it but I think we could honestly drop it, since we now credit the contributors much more prominently at the top.

PR #1358, opened 2019-04-29
Fix contributor list

erasche:
Bug noticed by @nsoranzo in https://github.com/galaxyproject/training-material/pull/1352


shiltemann:
thanks @erasche !

PR #1361, opened 2019-05-02
Short version of the 16S with mothur tutorial

shiltemann:
I created a shorter version of the 16S mothur tutorial. It follows all the same steps, but replaces some groups of tools with workflow runs to speed up the tutorial a bit. To avoid duplication between the two as much as possible, I basically turned the tutorial into a giant snippet.

@AnnaSyme I remember you mentioning wanting to make a shorter version of this tutorial as well, let me know if you have any thoughts about this :)

@erasche kindly agreed to test the shortened tutorial to check that it still flows nicely and makes sense, thanks!

Also tried to address [this feedback](https://github.com/galaxyproject/training-material/issues/989#issuecomment-489054194) and clarify skipping of the mock community analysis section

P.S. I also plan to use this tutorial at GCC workshop where I will not have enough time for the full tutorial.


erasche:
Awesome! Thanks @shiltemann

PR #1368, opened 2019-05-09
Maker updates

abretaud:
So here's an update for the maker tutorial:
- added some introduction slides
- updated the workflow
- a few corrections to the tutorial


shiltemann:
thanks @abretaud and @erasche !

PR #1369, opened 2019-05-09
tool names updated with minor corrections

kpbioteam:


kpoterlowicz:
Hi All, we are trying to publish our EWAS-Galaxy tool suite and we have received the following feedback regarding associated GTN website " training isn’t actually ‘enabled’ so is hidden except for direct tutorial link", what can we do to make training enabled in epigenetic section ?

PR #1377, opened 2019-05-13
Update ewas_training.ga

kpoterlowicz:


bedroesb:
Thank you for updating the workflow!

In case you want to unflatten the workflow, you can always use following command:
```
cat wf.ga | jq . -S > out.ga
```
based on following repo:
https://github.com/usegalaxy-eu/workflow-testing

nsoranzo:
Replaced by https://github.com/galaxyproject/training-material/pull/1560

PR #1393, opened 2019-05-31
Workflow Parameters

mvdbeek:


mvdbeek:
I think this is ready, if I can get a review that'd be great!

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
```diff
@@ -0,0 +1,160 @@
...
+contributors:
+- mvdbeek
+
+---
+
+
+# Introduction
+{:.no_toc}
+
+Workflows are a powerful feature in Galaxy that allow you to chain multiple steps of an anlysis together.
```
shiltemann:
```suggestion
Workflows are a powerful feature in Galaxy that allow you to chain multiple steps of an analysis together.
```

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
```diff
@@ -0,0 +1,160 @@
...
+Instead of selecting a specific number of lines we can choose to insert a workflow parameter
+that allows the user to set this parameter when running the workflow.
+To do that click on the `Select first` tool in the workflow and find the `Number of lines`
+parameter in the `Details` panel on the right hand side of the workflow.
+If you hover over the second symbol you will see a small text box appearing that says
+"Add connection to module". Click on this symbol and you will see that a new input with the text `Number of lines` has
+appeared in the `Select first` tool in the editor.
+
+To insert the workflow parameter click on Inputs in the tool panel on the right side of the workflow editor
+and select `Simple inputs used for workflow logic`. This will insert a new box into the workfow editor.
```
shiltemann:
```suggestion
and select `Simple inputs used for workflow logic`. This will insert a new box into the workflow editor.
```

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
```diff
@@ -0,0 +1,160 @@
...
+>   You can create or compute a dataset for each parameter you would like to try and run the
+>   `Parse parameter value` tool on it, running all downstream tools once for each parameter.
+{: .tip}
+
+# Conclusion
+{:.no_toc}
+
+Galaxy Workflows chain together different steps of an analysis. To make your workflows
+more useful to your colleagues you can add workflow parameters. Sometimes a parameter is not
+known in advance, but can be calculated as part of the workfow. Now you know how to read these
```
shiltemann:
```suggestion
known in advance, but can be calculated as part of the workflow. Now you know how to read these
```

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
```diff
@@ -0,0 +1,160 @@
...
+`Integer` on the right hand side you will be able to connect the workflow parameter to the `Number of lines` parameter.
+Make sure to also set a descriptive label and annotation, and save the workflow by clicking on the `save` icon.
+If you run the workflow you will now see the new parameter that needs to be set before the workflow can be run.
+
+
+![Animation of simple integer workflow parameter](../../images/workflow_integer_param.gif "Integer workflow parameter")
+
+> ### {% icon tip %} Tip: You can use a single parameter multiple times
+> 
+> * You can connect a parameter to multiple steps or to mulitple parameters within a step
```
shiltemann:
```suggestion
> * You can connect a parameter to multiple steps or to multiple parameters within a step
```

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
```diff
@@ -0,0 +1,160 @@
...
+    - *"Replacement"*: `\1bar`
+    - Connect the output of the  `Compose text parameter value` tool to the `Find Regex` parameter.
+    - Connect the data input to the `Select lines from` input of the `Regex Find And Replace` tool
+
+Now upload a text dataset with the contents `wunder`
+If you run this workflow on the dataset and you select `wunder` as the newly defined parameter  you will see 2 new datasets in your history. The first dataset has the data type ``expression.json`` and contains the composed parameter value `(wunder)`, the second dataset will contain the output of the `Regex Find And Replace` step. A click on the `i` button will show the used parameters. You will see that the `Find Regex` parameter will contain the values that you entered in the workflow run form. If you look at the dataset content you will see it is `wunderbar`.
+
+# Read a parameter from a dataset
+
+Often times it is necessary to calculate a parameter in one step of a workflow and then to use it in another step of the same workflow. This can be accomplished by reading the parameter from a dataset (As long as it is a text, integert, float, boolean or color parameter).
```
shiltemann:
```suggestion
Often times it is necessary to calculate a parameter in one step of a workflow and then to use it in another step of the same workflow. This can be accomplished by reading the parameter from a dataset (As long as it is a text, integer, float, boolean or color parameter).
```

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
```diff
@@ -0,0 +1,160 @@
...
+In this example we will construct a workflow where we calculate the sum of all values in a dataset and then divide the values in this dataset by the sum calculated in the previous step.
+
+## Construct example workflow for reading parameters from a file
+
+1. **Add a data input to a new workflow** {% icon tool %}
```
shiltemann: hands_on box?

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
```diff
@@ -0,0 +1,160 @@
...
+
+It is often necessary to create a text parameter that is flanked by additional text.
+Take for instance a regular expression that finds `foo` in a string and replaces it
+with `foobar` as in `s/(foo)/\1bar/`.
+The Galaxy tool ``Regex Find And Replace`` can be used to tun such a regular expression. The find part of the regex (`s/(foo)/`) can be defined in the `Find regex` parameter and the replacement part can be entered in the `Replacement` parameter.
+If we want to make the `foo` part of the regular expression configurable we can
+compose this text parameter using the `Compose text parameter value` tool.
+
+## Compose a text parameter
+1. **Add a data input to the workflow**
```
shiltemann: Put these steps in a proper hands-on box?

topics/galaxy-ui/tutorials/workflow-parameters/tutorial.md
mvdbeek: Yep, I can try that.

mvdbeek: I tried it in https://github.com/galaxyproject/training-material/pull/1393/commits/6817df43a7e8c98d1f1bff43bf6ac28f114812ad -- looks much better, thanks!

erasche:
@mvdbeek have a couple of fixes I'll push but I'm stuck at `Compose text parameter value`, where should this be found? Is that something from toolshed that's required? Or something in dev only?

erasche:
Update: found them, we'll just have to update .eu

mvdbeek:
yep, they're on the toolshed, https://toolshed.g2.bx.psu.edu/view/iuc/compose_text_param/feb3acba1e0a and should work starting from 19.05

erasche:
> If you run the workflow you will now see the new parameter that needs to be set before the workflow can be run.

For me it doesn't need to be set, in fact it defaults to 0 and is collapsed by default. This is with `optional: no`. Also there doesn't seem to be a way to provide a default value. Is this expected?

mvdbeek:
> Is this expected?

Unfortunately yes, at this point. Though we'll tackle this soon (probably at the encore fest if there is interest).

erasche:
Got a bit closer this morning but the expression tool is failing on EU with

```
Traceback (most recent call last):
  File "_evaluate_expression_.py", line 1, in <module>
    from galaxy_ext.expressions.handle_job import run; run()
  File "/opt/galaxy/server/lib/galaxy_ext/expressions/handle_job.py", line 31, in run
    with open(environment_path, "r") as f:
IOError: [Errno 2] No such file or directory: ''

cat: tmp6CL_1e: No such file or directory
```

erasche:
**Bug**

Upon clicking the workflow connection button here, no input was added.

image

**Progress**

Ok, I've finished most of the stylistic updates (added myself as contributor since it took a bit of time, hope that's ok for you) and I think it's fine (but again can't test easily, not working on .eu)

I've added some example datasets in places but if you have better ones, great, and if we could run the workflows on those inputs then we can have students check their answers maybe in question/solution boxes?

mvdbeek:
> also .eu seems to be eating my workflow, the regex step never gets scheduled and I'm not seeing an error in the logs.

That's to be expected if the expression tool doesn't run. Do you have node js installed and on galaxy's PATH ?
I also added https://github.com/bardin-lab/galaxy/commit/3a6ea0cbb61f06f4495fda6570a0214c4f923b3b on our instance, but didn't quite figure out is that was really necessary.

> Upon clicking the workflow connection button here, no input was added.

I can reproduce that on usegalaxy.eu, but it works fine on our local instance and on my dev instance. I see that https://github.com/usegalaxy-eu/galaxy/compare/release_19.05_europe...galaxyproject:release_19.05 is up to date, maybe a client rebuild is necessary ? Also usegalaxy.org is missing the expression tools in its entirety, so couldn't check there.

erasche:
>  Do you have node js installed and on galaxy's PATH ?

Oh gosh no! That's a requirement? It doesn't come from conda / normal dependencies? Our path is extremely clean. It's not in the .venv either for us, I guess it's only installed during the `make client` step and we do that somewhere else.

Clientbuild was latest commit on that branch, I can rebase and rebuild again later?

mvdbeek:
> Oh gosh no! That's a requirement? It doesn't come from conda / normal dependencies?

It's installed with `common_startup.sh` via nodeenv and should then be available via the virtualenv galaxy runs in in (https://github.com/galaxyproject/galaxy/blob/release_19.05/scripts/common_startup.sh#L233)

> Clientbuild was latest commit on that branch, I can rebase and rebuild again later?

Hmm, the commits look fine, nothing relevant missing as far as I can tell.

erasche:
If you don't build the client server side, then common startup isn't part of the ansible playbooks, so it's never been installed for us on EU.

Is there any reason that it isn't a more normal dependency?

mvdbeek:
> Is there any reason that it isn't a more normal dependency?

Just that it's not a python dependency that can be installed via pip.

erasche:
Oh, I meant more like a conda dependency / run as a normal tool?

mvdbeek:
That might work, we could give that a try.

mvdbeek:
https://github.com/galaxyproject/galaxy/pull/8224 for making node a regular tool dependency.
also it looks like the bug has disappeared after a restart ... yay!

erasche:
And after more restarts it's back!

> That's to be expected if the expression tool doesn't run. Do you have node js installed and on galaxy's PATH ?

apparently we actually do, just not in venv, I'd completely forgotten about that.

```
galaxy@sn04:~/server$ node -v
v8.16.0
galaxy@sn04:~/server$ which node
/usr/bin/node
```

so I'm even more confused now. but in theory, knowing how all of these parts fit together, the training at least looks good. So, I'm happy to merge it now. EU/tool/config bugs we can figure out at some other date.

erasche:
Thanks @mvdbeek!

PR #1394, opened 2019-06-03
Check trailing newlines in snippets

erasche:
Just saw this on my todo list from a while and it only took a minute


shiltemann:
thanks!

PR #1410, opened 2019-06-15
Age prediction tutorial updated with the new scikit-learn tools (1.0.0.4) in Galaxy

anuprulez:
I have updated the tutorial to make it work with the new version of Scikit-learn tools (1.0.0.4) in Galaxy.
@khanteymoori and @qiagu: Can you test this tutorial and add your comments? 
We are planning to teach this tutorial during the GCC 2019.

ping @bgruening 

Thanks!


khanteymoori:
Many thanks to Anup for updated tutorial. I already tested it, it is OK without any problem, These are some comments:
1- Unification of toolbox name:  "estimator attributes" or "Estimator attributes"
2-     “Whether to shuffle data before splitting” is True across successive experiments,
True should be replaced by Yes
3-  In "Hyperparameter Search" step, some of the options such as "refit" and .. are not mentioned and it can be accepted because they are kept in their default value. But, this line take about an option with default value:
“Whether to hold a portion of samples for test exclusively?”: Nope
I think it would be better if we have a similar policy, either mention all options or only changed options. 
4- New tool for "Ensemble methods" when we use in "Predict age" part, has a problem! It is due to missing header option in new version, so we should use the older one.
Screenshot_2019-06-18_19-09-11

anuprulez:
@khanteymoori thanks for your comments :)

bgruening:
Thanks a lot, great team work!

PR #1414, opened 2019-06-19
Prokka annotation tutorial: removed unnecessary commands; added slides

AnnaSyme:
For GCC 2019 workshop, took out some extra commands in Prokka and JBrowse that aren't needed anymore; also added a few more annotation slides. Moved slide images up into a higher level "images" directory.


package-lock.json
```diff
@@ -5,14 +5,14 @@
   "requires": true,
   "dependencies": {
```
bgruening: @AnnaSyme can you remove this change?

package.json
```diff
@@ -12,6 +12,6 @@
   },
   "homepage": "https://github.com/galaxyproject/training-material#readme",
   "dependencies": {
-    "decktape": "^2.9.1"
```
bgruening: can you also remove this change?

package-lock.json
AnnaSyme: thanks @bgruening  - do I just click delete for this file (and the next one) ?

AnnaSyme:
thanks @erasche looks good!

erasche:
Thanks so much @AnnaSyme!

PR #1425, opened 2019-06-26
Update variant-analysis metadata

wm75:
- Provide an adequate summary of the material in the variant analysis section.
- Add myself as a maintainer.


wm75:
Ah, thanks @bgruening

PR #1443, opened 2019-07-01
Miniconda prefix must be in double quotes

profgiuseppe:
As opposed to other values, miniconda_prefix must be in double quotes (as the solution shows), otherwise the user gets an error.


shiltemann:
Thanks a lot for your contribution(s) @profgiuseppe! feel free to add yourself to the CONTRIBUTORS.yaml file in the root of the repository so you will appear in our [contributor hall of fame](https://training.galaxyproject.org/training-material/hall-of-fame)

PR #1444, opened 2019-07-01
Add minor content, syntax fixes

jraysajulga:
Just catching minor recommendations as we go through the admin documentation for GCC2019 training.


jraysajulga:
It's the least I could do for that great workshop!

PR #1446, opened 2019-07-01
Nginx templates should be in the main directory

profgiuseppe:
Less confusing wording for the templates. That main level directory was never mentioned before (and problably it does not exists). Also, these templates could go into the `galaxyproject.nginx` role, but it may be too confusing so it's better to avoid mentioning that here.


erasche:
Thanks @profgiuseppe!

PR #1455, opened 2019-07-03
scRNApreprocessing use stranded options

lldelisle:
Hi,
I noticed that the data we are analysing are stranded so we should use the stranded parameters for countfeatures.
If I made something wrong or something is missing, please, tell me,

Lucille


shiltemann:
thanks a lot for your contribution @lldelisle! please feel free to add yourself to the CONTRIBUTORS.yaml file :)

lldelisle:
I cannot really call it a contribution... Just a correction to fit with the biological input...

shiltemann:
still, every contribution helps, big or small, it's what keeps this project going! so thank you a lot!

lldelisle:
To generate the small input dataset this is what I did:
1) Take only 20M R1 and R2 reads from the original dataset using seqtk sample (which is about 200M R1 and R2).
2) Evaluate that chr22 is about 1% of the other numbered chr
- run the workflow: https://raw.githubusercontent.com/lldelisle/myWorkflows/master/Galaxy-Workflow-MakeAFakeInput.ga
which: 1) Take randomly the id of 1% of the input R1 fastq
2) Take all id of reads which map to chr22
3) Concatenate the two and remove duplicates
4) use seqtk subseq to select R1 and R2...
We are now down to 380k reads.

bebatut:
@mtekman can you check this change? thanks a lot

mtekman:
Thanks! This was brought up during the tutorial and I was planning to reference this with a later edit, but I am very happy to see this here now :+1:

mtekman:
@lldelisle Please do add yourself as a contributor before this is merged!

lldelisle:
I am sorry but I needed to remove the repository to work with Maria on the ATAC-seq WF... 
I will add myself as a contributor in the https://github.com/galaxyproject/training-material/pull/1472
Many thanks,

bebatut:
Thanks a lot @lldelisle!

PR #1460, opened 2019-07-04
Remove tool.yaml files and generate them from workflows for instance annotation during travis build

shiltemann:
`tools.yaml` files can be generated from the workflow files using ephemeris, so I've removed them from this repo. 

I have added a script to generate the tool.yaml files from the workflows for all the tutorials. The tools.yaml files are still needed for the instance annotation and badge generation, and the script might also be useful for Galaxy admins looking to support certain tutorials

(wip until I test a bit more)


erasche:
:tada: :tada:

bebatut:
Do we need the `tools.yaml` files if we have the workflow?

erasche:
If we don't, I think the issue is that the instances that it can run on won't appear

bebatut:
Sorry I misread and did not take the time to check the PR. I thought you were generating again all `tools.yaml` files from workflow and stored them here... 
Looks good now that I understood

shiltemann:
k, got a chance to test this, should be good now :)

bebatut:
Thank you!

PR #1472, opened 2019-07-05
ATAC-seq tutorial

mblue9:
First draft for ATAC-seq tutorial created at CoFest. Currently being worked on by myself and @lldelisle, @heylf, @glorybsm

- [x] Create tutorial skeleton created from workflow with planemo (`planemo training_init --topic_name epigenetics --tutorial_name atac-seq --workflow_id bfffdc9ed6c05226 --galaxy_url https://usegalaxy.eu --galaxy_api_key xxxxxx`)
- [x] Add initial description of workflow steps
- [x] Add input datasets to zenodo
- [x] Add introduction
- [x] Add conclusion
- [x] Expand descriptions of steps
- [x] Add visualisation step
- [x] Replace MACS2 with Genrich
- [x] Check all steps/parameters
- [x] Check if need to add more datasets to Zenodo (then update zenodo links)
- [x] Add some Questions
- [x] Add Key Points
- [x] Add contributor names (have to exist in contributors.yml) or Jekyll will fail

Note: If update workflow steps, then to update skeleton can run `planemo training_generate_from_wf --topic_name epigenetics --tutorial_name atac-seq --workflow_id bfffdc9ed6c05226 --galaxy_url https://usegalaxy.eu --galaxy_api_key xxxxxx`


lldelisle:
To generate the small input dataset this is what I did:

    Take only 20M R1 and R2 reads from the original dataset using seqtk sample (which is about 200M R1 and R2).
    Evaluate that chr22 is about 1% of the other numbered chr

    run the workflow: https://raw.githubusercontent.com/lldelisle/myWorkflows/master/Galaxy-Workflow-MakeAFakeInput.ga
    which: 1) Take randomly the id of 1% of the input R1 fastq

    Take all id of reads which map to chr22
    Concatenate the two and remove duplicates
    use seqtk subseq to select R1 and R2...
    We are now down to 380k reads.

mblue9:
Ok I think we're done here with the first version of an ATAC-Seq tutorial and ready for review. ping @lldelisle @heylf

Gemfile
```diff
@@ -1,13 +0,0 @@
-source "https://rubygems.org"
-gem 'nokogiri'
```
bgruening: these files should not be deleted

bgruening:
:heart: thanks all!

tools.yml
```diff
@@ -0,0 +1,65 @@
+install_tool_dependencies: True
```
nsoranzo: @bgruening This file shouldn't have been added, right?

This is generated by `planemo training_init`, probably should be created as a temporary file.

lldelisle: Sorry, it was created when I ran planemo but I should not have added it.

PR #1485, opened 2019-07-07
Heroku revival

martenson:
This reintroduces the gemfiles to support ruby-standard deployment on Heroku.

Heroku will deploy with these limitations:

- it has to be pull request
- source and target branch both have to be from this repository for automatic deploy
- for deploying forks the launch has to be explicitly allowed through Heroku console that @erasche @bebatut @shiltemann and @martenson have access to
- atm we have limit of 5 concurrent deployments (I'll try to change that)
- deployments self-destroy after a day (configurable)


martenson:
Jekyll is very stubborn in using Gemfile when it detects it, I have found no configuration option to ignore it, so as a part of `make serve` and `make detached-serve` I've added commands to mv the gemfile(lock) away which preserves the original conda-stack behavior of building things,

erasche:
@martenson, ok to merge?

martenson:
@erasche ok to review, I think it'll work

erasche:
:tada:

PR #1489, opened 2019-07-08
Markdownify front matter

tobyhodges:
This should allow your learning objectives, questions, and key points, from the YAML front matter to render nicely in the HTML.


erasche:
It looks fine, but I'm not sure we want to encourage formatting like this? It could be distracting. But maybe for things like bacteria names

bebatut:
Thanks a lot @tobyhodges

PR #1493, opened 2019-07-08
Transcriptomics: Add step to annotate genes with name, location from GTF

bebatut:
using `Annotate DESeq2/DEXSeq output tables` tool directly to DESeq2 output

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -111,7 +111,6 @@ icon-tag:
   galaxy-dropdown: fa-caret-down
   galaxy-wf-connection: fa-arrows-h
   galaxy-wf-new: fa-plus
-  galaxy-scratchbook: fa-th
 
   # menus, links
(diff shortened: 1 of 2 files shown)


_config.yml
```diff
@@ -111,6 +111,7 @@ icon-tag:
   galaxy-dropdown: fa-caret-down
   galaxy-wf-connection: fa-arrows-h
   galaxy-wf-new: fa-plus
+  galaxy-scratchbook: fa-th
```
shiltemann: we already have this icon 4 lines above :)

erasche:
```suggestion
```

bebatut:
@galaxyproject/training-transcriptomics any feedback on that PR?

bebatut:
This PR is out of date and most of it has been fixed with #1549

PR #1497, opened 2019-07-08
Workflow for non-dip

MoHeydarian:
Adding workflow for non-diploid variant analysis tutorial.


bebatut:
@wm75 could you check this workflow? thanks a lot

wm75:
To make this more robust (e.g. for automated testing with zenodo uploads), it may be a good idea to set the read group ID and sample names explicitly at the bwa-mem steps. If you have them autodetected they will be determined from the dataset name, and that will be different as soon as you upload via link. This will affect the final output, i.e., will break testing.

wm75:
@MoHeydarian are you sure these changes belong here? Looks like you forgot to check out a new topic branch first.

MoHeydarian:
@wm75 correct, I did not intend to push anything to the GTN branch. Apologies for the confusion.

PR #1508, opened 2019-07-10
Add select & concatenate alternative to awk

yvanlebras:


yvanlebras:
Hi @shiltemann , @bgruening , as I want to use this tuto to spread if the ecology.usegalaxy.org existence and give the opportunity to new Galaxy people to directly test it, I think we can add this warning if the awk tool is not installed or like here doesn't work.

shiltemann:
@yvanlebras is the `NLBMS` term you're selecting for always in the same column? if so you could use the `Filter - data on any column using simple expressions` tool to do it (and you can tell it keep the header line so it can be one step)

yvanlebras:
problem is that the filter tool output is a tabular one and this will not be usable as input (and the tabular to csv converter don't output text delimiter ...)

shiltemann:
The input file for the awk step is currently described as:

` - {% icon param-file %} *"Select cells from"*: output of **CSV to tabular** {% icon tool %}` 

so I assumed input and output were both tabular for this step, is this not the case? should that line be changed then (I see we indeed don't run that CSV to tabular tool in this tutorial?)

if we can find a way to do this without awk it could be good, since this also always feels a bit like magic for many users, whereas simple filter tools are a bit easier to understand. I will have a look if I can find an easy way to do this at some point

yvanlebras:
Arf right, I updated the tuto to remove the csv to tabular step, but didn't update the "input" info from next step (here awk). We definitely don't have to use any csv to tabular or inverse in this tuto ;) Thank you for pointing this!

yvanlebras:
Thank you so much @shiltemann ! Do you think you can merge now ? I think it's ok so I can then write some blog posts, and widespread the info of ecology.usegalaxy.eu ;)

shiltemann:
thanks @yvanlebras! will merge as soon as tests pass again :)

yvanlebras:
Thank you so much!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #1509, opened 2019-07-11
Support for subtopics

erasche:
The admin topic has a huge number of tutorials, and we (@Slugger70, others, myself) found it confusing and overwhelming quite often. This is an experiment to segregate them into three groups:

- core (base / foundational knowledge)
- features (places to start expanding on that foundation)
- extras (in depth dives / less related topics to core of admin)

image

Even if this is not generalisable (yet/ever), we would like to try deploying this experiment for a while and seeing how people respond to it. At least for admin, I think this separation is important, for other topics with maybe 3-10 tutorial it is less important?

NB: I know it adds a duplicate template, but it allows us to experiment with it a bit more without going through jekyll hell to make it generally working.


bebatut:
Good idea to organize the topics better :smile:

We have more and more tutorials in Transcriptomics. We also need such organization also. So what about doing it directly and avoid template duplication.
We could use the first tag (if exist) as subtopic or create another metadata called subtopic. 

What do you think?

erasche:
> So what about doing it directly and avoid template duplication.

I avoided this because it would be a fair bit of extra work :( the templating syntax is quite unpleasant, the templating I wrote for the admin portions requires a lot of ugly if/else/unless. I.e. the last line of my original post:

    NB: I know it adds a duplicate template, but it allows us to 
    experiment with it a bit more without going through jekyll hell
    to make it generally working.

> We could ... create another metadata called subtopic.

I talked a bit with @shiltemann in person, it sounds like a nice way to go forward but it would take a lot more time/energy than I have currently? And maybe first as I mentioned let's try this experiment to see if it's nice here and then decide on a final way forward?

Lastly, I currently feel also a bit that admin is "special", it has definitely a progression from "learn these", "then build", "+ unrelated stuff"? I'm not sure other topics have a similar progression but you would know better for transcriptomics.

erasche:
@bebatut re: in person conversation, I will find time to work on this soon. Leave it to me :)

erasche:
@shiltemann fixed all of the bugs with my original and made it more generic! It works way better now!!

Open todos for me, for monday

- [ ] add "Further learning" as title again instead of features
- [ ] fix the extra boxes in admin/dev

shiltemann:
merge #1510 first, I merged that branch into this one to work on search box for pages with subtopics, any conflicts should be minimal and I will resolve here after :) thanks!

shiltemann:
ok, looks like no conflicts :)

erasche:
@bebatut does this need anything else?

bebatut:
Looks good. Thanks!

PR #1510, opened 2019-07-12
Search box for tutorials in a topic

shiltemann:
Some topics (like admin) are getting to have a lot of tutorials in them, making it hard to find tutorials. This implements a search box to filter the tutorial list (using [List.js](https://listjs.com/)). You can type keywords and it starts filtering, or you click on a tutorial tag to filter the list to all tutorials having that label

topic-search


_layouts/topic.html
```diff
@@ -155,4 +160,23 @@ <h2 id="references">References</h2>
     </section>
 </div>
 
+<script src="{{ site.baseurl }}/assets/js/list.min.js"></script>
+<script type="text/javascript">
```
bebatut: Could you move this js script to the `assets/js/main.js` file? Thanks

PR #1514, opened 2019-07-17
Uniprot version info addition

combesf:
Addition of the version of Uniprot (ID conversion tool)


shiltemann:
thanks @combesf !

combesf:
Thank you Saskia !

PR #1515, opened 2019-07-17
CODEOWNERS: update to include subfolders

shiltemann:
the trailing `*` only matches files directly in the folder, while intended match was any files in any subfolders


erasche:
Not under any testing, no point in waiting.

erasche:
Thanks @shiltemann :)

Issue #1518, opened 2019-07-17
Create tutorial from workflow bug


sbenateau:
Hello,
I wanted to create a tutorial from a workflow a colleague had build. I used the code from this page: https://galaxyproject.github.io/training-material/topics/contributing/tutorials/create-new-tutorial/tutorial.html

```
planemo training_init
        --topic_name "ecology"\
        --tutorial_name "STOC-population-trend"\
        --tutorial_title "STOC population trend"\
        --workflow "../../Downloads/Galaxy-Workflow-Workflow_STOC_population_trend.ga"\
        --zenodo_link "https://doi.org/10.5281/zenodo.3339190"
```

It works perfectly with the example workflow:
topics/contributing/tutorials/create-new-tutorial/workflows/example-workflow.ga
But not with the workflow that my colleague send me.
You can find the .ga file here with the result of the code provided above:
https://github.com/sbenateau/training-material/tree/ecology-STOC-newTuto/topics/ecology/tutorials/STOC-population-trend/workflows

bebatut:
Hi,

Which error did you get?

I tried also locally. I got an error because the tools can not be found on the main ToolShed (planemo needs to install the tools to extract their information). And all tools in the workflow are from the test ToolShed. It may the reason why it is failing, at least locally.

If the workflow is on a running Galaxy instance, you could also use it without downloading the workflow locally and having issues of tool installation:

1. Extract the workflow id (on the URL when you edit or run the workflow)
2. Get your API key
3. Run: 

   ```
   planemo training_init \
        --topic_name "ecology" \
        --tutorial_name "STOC-population-trend" \
        --tutorial_title "STOC population trend" \
        --workflow_id <WORKFLOW_ID> \
        --galaxy_url <GALAXY_URL> \
        --galaxy_api_key <YOUR_API_KEY> \
        --zenodo_link "https://doi.org/10.5281/zenodo.3339190"
   ```

I hope this will help you!

sbenateau:
Thank you very much @bebatut. The problem comes indeed from the tools (they are only on our development instance that wasn't accessible anymore).

sbenateau:
I went one step further with the code you provided @bebatut . 

The tutorial is generated but I get this error:
`bioblend ERROR: GET: error 500: b'{"err_msg": "Uncaught exception in exposed API method:", "err_code": 0}', 0 attempts left`

And after I created the tutorial, I get the following error when I use `make serve`
[...]

PR #1521, opened 2019-07-19
add warning for generation of tutorials from workflows files

sbenateau:


topics/contributing/tutorials/create-new-tutorial/tutorial.md
```diff
@@ -256,7 +256,7 @@ For the next times, you can make it quicker.
 >             --workflow_id "ID of the workflow on the Galaxy instance" \
 >             --zenodo_link "URL to the Zenodo record"
 >      ```
->    - option 2: from a local workflow file (`.ga`)
+>    - option 2: from a local workflow file (`.ga`) (use only if your workflow is composed of tools from the main ToolShed)
```
bgruening: That is not really true I guess. The crucial point is that the tools in your workflow need to be installed in the Galaxy that you use. Where the tools coming from should not matter, but they need to be installed.

yvanlebras: Hmmm, really strange effectively, but using .ga local workflow file seems to have this issue if I well understood https://github.com/galaxyproject/training-material/issues/1518 B! ;) comment

PR #1522, opened 2019-07-23
Sequence analysis: Expand QC tutorial with more explanation

bebatut:
I tried to add more details about FastQC reports, specially for the different type of data.
I tried also to clarify some steps for Cutadapt and paired-end explanation.

Feedbacks there would be really great!


topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -110,21 +111,7 @@ But what does this quality mean?
...
- !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefgh
- |                         |    |        |                              |
-33                        59   64       73                            104
- 0........................26...31.......40
-                          -5....0........9.............................40
-                                0........9.............................40
-                                   3.....9.............................40
- 0.2......................26...31........41
-```
+![Encoding of the quality score with ASCII characters for different Phred encoding](../../images/quality_score_encoding.png)
```
bernt-matthias: Technically the Illumina 1.8 version of the format seems to be ASCII 33 to 93:

http://resources.qiagenbioinformatics.com/manuals/clcgenomicsworkbench/700/Quality_scores_in_Illumina_platform.html

This is also discussed in the discussion of the wikipedia entry. Though, to me it seems that values >40 don't appear with current sequencers/software. 

This is also a problem for the data format sniffer. As I remember Galaxy just takes everything as fastqsanger.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
-> 2. How is the mean score changing along the sequence?
-> 3. Is this tendency seen in all sequences?
+> Which Phred encoding is used in the FASTQ file for these sequences?
 >
 > > ### {% icon solution %} Solution
-> > 1. The Phred scores are encoded using `Sanger / Illumina 1.9` (`Encoding` in the top table)
-> > 2. The mean score over the sequence is dropping at the end of the sequences. It is usual: the sequencers are incorporating more errored nucleotides at the end. But the score stays good: over 28.
-> > 3. The box plots are getting wider at the end of the sequences. It means a lot of sequences have their score dropping at the end of the sequence. At after 31 nucleotides, more than 10% of the sequences have scores below 20.
-> >
+> > The Phred scores are encoded using `Sanger / Illumina 1.9` (`Encoding` in the top table).
```
bernt-matthias: 1.8?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+
+![Per base sequence quality](../../images/quality-control/per_base_sequence_quality.png "Per base sequence quality")
+
+On the x-axis are the base position in the read. In this example, the sample contains reads that are 37 bp long.
+
+> ### {% icon comment %} Non uniform X-axis
 >
-> - **Per sequence quality scores**
+> The X-axis is not uniform. It starts out with individual 1-10 bases. After that, bases are binned across a window a certain number of bases wide. The number of base positions binned together depends on the length of the read. With 150 bp reads, the latter part of the plot will report aggregate statistics for 5bp windows. Shorter reads will have smaller windows and longer reads larger windows. 
+{: .comment}
```
bernt-matthias: Note that binning can be changed now.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+For each position, a boxplot (BoxWhisker type ) is drawn with:
+
+- the median value, represented by the central red line
+- the inter-quartile range (25-75%), represented by the yellow box
+- the 10% and 90% points in the upper and lower whiskers
+- the mean quality in the blue line
+
+The y-axis shows the quality scores. The higher the score, the better the base call. The background of the  graph divides the y-axis into very good quality scores, scores of reasonable quality (orange), and reads of poor quality (red).
+
+It is normal with all Illumina sequencers for the median quality score to start out lower over the first 5-7 bases and to then rise. The quality of reads on most platforms will after steadily drop over the length of the read. This is often due to signal decay or phasing during the sequencing run:
```
bernt-matthias: will after -> will

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+  The fluorescent signal intensity decays with each cycle of the sequencing process. Due to the degrading fluorophores, a proportion of the strands in the cluster are not being elongated. The proportion of signal being emitted continues to decrease with each cycle, yielding to a decrease of quality scores at the 3' end of the read. 
+
+- Phasing
+
+  The signal starts to blur as the number of cycles increases because the cluster loses synchronicity. As the cycles progress, some strands get random failure of nucleotides to incorporate due to:
+
+  - Incomplete removal of the 3' terminators and fluorophores
+  - Incorporation of nucleotides without effective 3' terminators
+
+  It yields to a decrease in quality scores at the 3' end of the read.
```
bernt-matthias: It yields -> This leads (maybe)

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
 >
->     ![Per sequence GC content](../../images/quality-control/per_sequence_gc_content.png "Per sequence GC content")
+> > ### {% icon solution %} Solution
+> > 1. The mean score over the sequence is dropping at the end of the sequences. It is usual: the sequencers are incorporating more errored nucleotides at the end. But the score stays good: over 28.
+> > 2. The box plots are getting wider at the end of the sequences. It means a lot of sequences have their score dropping at the end of the sequence. At after 31 nucleotides, more than 10% of the sequences have scores below 20.
+> >
+> {: .solution }
+{: .question}
+
+When the median quality is to a phred score of ~20, we should consider trimming the sequence at that point, since sequence with lower quality than that is likely to cause more problems than it fixes. We will explain that process in the next section
```
bernt-matthias: is to a phred ... probably some word missing?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+
+In a random library we would expect that there would be little to no difference between the four bases. The proportion of each of the four bases should remain relatively constant over the length of the read with `%A=%T` and `%G=%C`, and the lines in this plot should run parallel with each other. 
+
+> ### {% icon details %} Biases by library type
+> 
+> It's worth noting that some types of library will always produce biased sequence composition, normally at the start of the read. Libraries produced by priming using random hexamers (including nearly all RNA-Seq libraries as in the previous plot), and those which were fragmented using transposases, will contain an intrinsic bias in the positions at which reads start (the first 10-12 bases). This bias does not involve a specific sequence, but instead provides enrichment of a number of different K-mers at the 5' end of the reads. Whilst this is a true technical bias, it isn't something which can be corrected by trimming and in most cases doesn't seem to adversely affect the downstream analysis. It will, however, produce a warning or error in this module.
```
bernt-matthias: types of library -> library types

bernt-matthias: Also a funny effect that I found recently is that the phred score is very low if the first nucleotides are always the same (e.g. in RADSeq where a restriction enzyme always cuts at a certain pattern). Then the complete flow cell always sees the same nucleotide which confuses the error model of the base caller.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+> > In the beginning of sequences, the sequence content per base is not really good and the percentages are not equal, as expected for RNA-seq data.
+> {: .solution }
+{: .question}
+
+
+## Per sequence GC content
+
+![Per sequence GC content](../../images/quality-control/per_sequence_gc_content.png "Per sequence GC content")
+
+This plot displays the number of reads vs. GC% per read. It is compared it to a theoretical distribution assuming an uniform GC content for all reads, expected for whole genome shotgun sequencing, where the central peak corresponds to the overall GC content of the underlying genome. Since the GC content of the genome is not known, the modal GC content is calculated from the observed data and used to build a reference distribution.
```
bernt-matthias: It is compared it to -> remove 2nd it

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+> ### {% icon details %} More details about other FastQC plots
 >
->     It plots out the percentage of base calls at each position for which an N was called.
+> #### Per tile sequence quality
+> 
+> ![Per tile sequence quality](../../images/quality-control/per_tile_sequence_quality.png "Per tile sequence quality")
+> 
+> This graph will only appear for Illumina library which retains its original sequence identifiers. Encoded in these is the flowcell tile from which each read came. The graph enables you to look at the quality scores from each tile across all of your bases to see if there was a loss in quality associated with only one part of the flowcell.
+>   
+> The plot shows the deviation from the average quality for each tile. The colours are on a cold to hot scale, with cold colours being positions where the quality was at, or above, the average for that base in the run, and hotter colours indicate that a tile had worse qualities than other tiles for that base. In the example above you can see that certain tiles show consistently poor quality. A good plot should be blue all over.
```
bernt-matthias: I often observe something like this in our RNASeq data:

pertile

we analysed this a bit further and have noticed that the horizontal lines are tiles where the chemicals arrive last. That is the chemistry is a bit worse there.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
->     This graph shows the distribution of fragment sizes in the file which was analysed. In many cases this will produce a simple graph showing a peak only at one size, but for variable length FASTQ files this will show the relative amounts of each different size of sequence fragment.
+> ![Per base N content](../../images/quality-control/per_base_n_content_error.png)
+> 
+> #### Sequence length distribution
 >
-> - **Sequence Duplication Levels**
+> ![Sequence length distribution](../../images/quality-control/sequence_length_distribution.png "Sequence length distribution")
 >
->     ![Sequence Duplication Levels](../../images/quality-control/sequence_duplication_levels.png "Sequence Duplication Levels")
+> Some high throughput sequencers generate sequence fragments of uniform length, but others can contain reads of wildly varying lengths. Even within uniform length libraries some pipelines will trim sequences to remove poor quality base calls from the end.
```
bernt-matthias: For instance the Illumina base caller (bcl2fastq) does adapter trimming (by default) in a way that the first n bases are trimmed if they matche the first n bases of the adapter up to 90% (by default). The problem is that n can also be 1. 

In the plot you see then that ~1/4 of the sequences are shorter by 1, ~1/16 of the data shorter by 2, ....

So we disabled this trimming now and use cutadapt.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+> #### Kmer Content
 >
-> - **Adapter Content**
+> FastQC does a generic analysis of all of the short nucleotide of length k (kmer, with k = 7 by default) starting at each positon along the read in the library to find those which do not have even coverage through the length of your reads. Any given kmer should be evenly represented across the length of the read. 
 >
->     ![Adapter Content](../../images/quality-control/adapter_content.png "Adapter Content")
+> FastQC will report the list of kmers which appear at specific positions with greater than expected frequency are reported. This can find a number of different sources of bias in the library which can include the presence of read-through adapter sequences building up on the end of the sequences. The presence of any overrepresented sequences in the library (such as adapter dimers) causes the kmer plot to be dominated by the kmer these sequences contain, and then it's not always easy to see if there are other interesintg biases are present.
```
bernt-matthias: remove "are reported"

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
-{: .question}
-
+> #### Amplicon
+> 
+> Amplicon libraries are prepared by PCR amplification of a specific target. For example, the V4 hypervariable region of the bacterial 16S rRNA gene. All reads from this type of library are expected to be nearly identical. It will result in:
+>
+> - Extremely biased per base sequence content
+> - Extremely narrow distribution of GC content
+> - Very high sequence duplication levels
+> - Abundance of overrepresented sequences
```
bernt-matthias: should have the same quality problems as the RADSeq, or are specific base caller parameters used?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,297 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+The quality of the sequences drops at the end of the sequences. This could cause bias in downstream analyses with these potentially incorrectly called nucleotides. Sequences must be treated to reduce bias in downstream analyis. In general, quality treatments include:
 
-- Filtering of sequences
+1. Cutting/Trimming/masking sequences
+    - from low quality score regions
+    - beginning/end of sequence
+    - removing adapters
+2. Filtering of sequences
     - with low mean quality score
     - too short
```
bernt-matthias: that are to short?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -318,21 +462,21 @@ To accomplish this task we will use [Cutadapt](https://cutadapt.readthedocs.io/e
...
 >       - In *Read 1 Options*
->
->          In this training, no adapters were already removed. If you know which adapter sequences were used during library preparation, provide their sequences there.
->
->    - In *"Adapter Options"*
->       - *"Minimum overlap length"*: `3`
+>          
+>          > ### {% icon comment %} Know adapters
+>          > In this dataset, no adapters were found as we saw in FastQC report. They were already removed. 
+>          > If you see or know which adapter sequences were used during library preparation, provide their sequences there.
```
bernt-matthias: Would be interesting to have a data set with adapters. With different adapters for forward and reverse strand we often found that using the common part is sufficient and maybe even better.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -374,44 +518,76 @@ The quality of the previous dataset was pretty good from the beginning and we im
...
+> ```
+> 42 40 26 27
+> ```
+>
+> Note that thereby also positions with a quality value larger than the chosen threshold are removed if they are embedded in regions with lower quality (note that the partial sum is decreasing if the quality values are smaller than the threshold). The advantage of this procedure is that it is robust against a small number of positions with a quality higher than the threshold.
+>
+>
+> Alternatives to this procedure would be:
+>
+> * Cut all positions with a quality smaller than the threshold
```
bernt-matthias: Cut after the first position after ... ? Cutting single positions somehow makes no sense.

But this is essentially a sliding window of size 1.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -374,44 +518,76 @@ The quality of the previous dataset was pretty good from the beginning and we im
...
+> * Sliding window approach
+>
+>     The sliding window approach checks that the average quality of each sequence window of specified length is larger than the threshold. Note that in contrast to cutadapt's approach, this approach has one more parameter and the robustness depends of the length of the windows (in combination with the quality threshold). Both approaches are implemented in Trimmomatic.
```
bernt-matthias: windows -> window

bernt-matthias:
Really nice improvement. 

Just in case you dont know yet, I really like this resource: https://sequencing.qcfail.com/

bebatut:
Thanks @bernt-matthias for the comments

> Just in case you dont know yet, I really like this resource: sequencing.qcfail.com

I forgot about it. Nice reminder. I added as ressource to check

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: We could have a dataset with adapters as an extra exercise. Do you one dataset with some?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: Rephrased

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: I never analyzed RADSeq data. Do you want to add the use case here?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: I added this example in the explanation. Let me know what you think about it

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: Should we mention that here? How would you phrase it?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: Note added

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: From FastQC report, it is `Sanger / Illumina 1.9`

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: Should we add a note for that? I am afraid it makes it even more complicated

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bernt-matthias: Ah, OK. I guess (but not sure) 1.9 just has a different header format. IMHO the most important piece of information is if its phred33 or phred64 encoding. Everything else is just nasty details :)

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bernt-matthias: Maybe something along the following lines: 

"""
An extreme example is the RADSeq protocol where fragmentation is done with restriction enzymes that cut at a specific short sequence motif, i.e. each fragment starts with the same nucleotides which are processed in lock step by the sequencer. Thus, the sequencer gets the same signal for the complete lane or flow cell while processing these nucleotides. Since the error model of the base caller can't cope with this case it assigns very low quality values to these sequences.
"""

ping @yvanlebras do you agree with this?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bernt-matthias: fine for me

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bernt-matthias: not necessarily

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bernt-matthias: unfortunately not.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -420,11 +603,8 @@ Let's first have a look at the quality of our reads!
 >    ```
 >
 > 2. Rename the file to `reads_2`
-> 3. **FastQC** {% icon tool %} with the reverse reads
->
->     FastQC has already been run on the forward reads
->
-> 4. **MultiQC** {% icon tool %} with the following parameters to aggregate the FastQC reports
+> 3. **FastQC** {% icon tool %} with the reverse reads>
```
anuprulez: Looks like `>` at the end of the line is not needed.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -110,21 +111,7 @@ But what does this quality mean?
 
 The quality for each sequence is a string of characters, one for each base of the nucleic sequence, used to characterize the probability of mis-indentification of each base. The score is encoded using the ASCII character table (with [some historical differences](https://en.wikipedia.org/wiki/FASTQ_format#Encoding)):
```
erxleben: The quality score for each sequence is displayed as a string of characters

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: Nice catch!

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,304 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+
+But there are also other situations in which an unusually-shaped distribution may occur. For example, with RNA sequencing there may be a greater or lesser distribution of mean GC content among transcripts causing the observed plot to be wider or narrower than an idealized normal distribution.
+
+> ### {% icon question %} Questions
 >
->     An unusually-shaped distribution could indicate a contaminated library or some other kind of biased subset. A normal distribution which is shifted indicates some systematic bias, which is independent of base position. If there is a systematic bias which creates a shifted normal distribution then this won't be flagged as an error by the module since it doesn't know what your genome's GC content should be.
+> Why is there a warning for the per sequence GC content graphs?
```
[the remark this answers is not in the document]
bebatut: I used graphs everywhere

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,304 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+
 Rather than looking at quality scores for each individual read, FastQC looks at quality collectively across all reads within a sample:
 
 ![Per base sequence quality](../../images/quality-control/per_base_sequence_quality.png "Per base sequence quality")
 
-On the x-axis are the base position in the read, and on the y-axis quality scores. In this example, the sample contains reads that are 37 bp long. For each position, there is a box plotted to illustrate the distribution of values (with the whiskers indicating the 90th and 10th percentile scores). The blue line represents the mean score at the postion. The plot background is color-coded to identify good (green), acceptable (yellow), and bad (red) quality scores.
+On the x-axis are the base position in the read. In this example, the sample contains reads that are 37 bp long.
 
-> ### {% icon question %} Questions
+> ### {% icon comment %} Non uniform x-axis
```
shiltemann: is this maybe too much info? in the example shown here the axis /is/ uniform, so I would say either leave this out, or show it as a details box with an example image included of a case where this nonuniformity is present

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -106,27 +107,13 @@ IIIIIIIIIIIIIIIIIIIIIIIIIIIII8IIIIIII
...
- 0........................26...31.......40
-                          -5....0........9.............................40
-                                0........9.............................40
-                                   3.....9.............................40
- 0.2......................26...31........41
-```
+![Encoding of the quality score with ASCII characters for different Phred encoding](../../images/quality_score_encoding.png)
 
-So each nucleotide is assigned an ASCII character, representing its [Phred quality score](https://en.wikipedia.org/wiki/Phred_quality_score), the probability of an incorrect base call:
+So each nucleotide is assigned to an ASCII character, representing its [Phred quality score](https://en.wikipedia.org/wiki/Phred_quality_score), the probability of an incorrect base call:
```
shiltemann:
```suggestion
So there is an ASCII character associated with each nucleotide, representing its [Phred quality score](https://en.wikipedia.org/wiki/Phred_quality_score), the probability of an incorrect base call:
```

topics/sequence-analysis/tutorials/quality-control/tutorial.md
shiltemann: I agree this would probably only be confusing for people, and we can probably leave it out since this tutorial is meant as a first introduction to QC?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,304 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+> #### Kmer Content
 >
-> - **Adapter Content**
+> FastQC does a generic analysis of all of the short nucleotide sequences of length k (kmer, with k = 7 by default) starting at each position along the read in the library to find those which do not have an even coverage through the length of your reads. Any given kmer should be evenly represented across the length of the read. 
 >
->     ![Adapter Content](../../images/quality-control/adapter_content.png "Adapter Content")
+> FastQC will report the list of kmers which appear at specific positions with a greater frequency than expected. This can be due to different sources of bias in the library, including the presence of read-through adapter sequences building up on the end of the sequences. The presence of any overrepresented sequences in the library (such as adapter dimers) causes the kmer plot to be dominated by the kmer from these sequences. Any biased kmer due to other interesting biaises may be then diluted and not easy to see.
```
shiltemann:
```suggestion
> FastQC will report the list of kmers which appear at specific positions with a greater frequency than expected. This can be due to different sources of bias in the library, including the presence of read-through adapter sequences building up on the end of the sequences. The presence of any overrepresented sequences in the library (such as adapter dimers) causes the kmer plot to be dominated by the kmer from these sequences. Any biased kmer due to other interesting biases may be then diluted and not easy to see.
```

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,304 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+> - Adapter content > 0% starting at base 1
+>
+{: .details}
 
 # Filter and Trim
 
-We will filter and trim the sequences based on the information provided by the quality graphs, the quality of the sequences drops at the end of the sequences. This could cause bias in downstream analyses with these potentially incorrectly called nucleotides.
-
-Sequences must be treated to reduce bias in downstream analyis. In general, quality treatments include:
+The quality of the sequences drops at the end of the sequences. This could cause bias in downstream analyses with these potentially incorrectly called nucleotides. Sequences must be treated to reduce bias in downstream analyis. In general, quality treatments include:
```
shiltemann:
```suggestion
The quality of the sequences drops at the end of the sequences. This could cause bias in downstream analyses with these potentially incorrectly called nucleotides. Sequences must be treated to reduce bias in downstream analysis. In general, quality treatments include:
```

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -371,47 +522,81 @@ The quality of the previous dataset was pretty good from the beginning and we im
 > ### {% icon comment %} Bad quality sequences
 > If the quality of the reads is not good, we should always first check what is wrong and think about it: it may come from the type of sequencing or what we sequenced (high quantity of overrepresented sequences in transcriptomics data, biaised percentage of bases in HiC data).
```
shiltemann:
```suggestion
> If the quality of the reads is not good, we should always first check what is wrong and think about it: it may come from the type of sequencing or what we sequenced (high quantity of overrepresented sequences in transcriptomics data, biased percentage of bases in HiC data).
```

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -371,47 +522,81 @@ The quality of the previous dataset was pretty good from the beginning and we im
...
 {: .comment}
 
-# Process paired-end data
+> ### {% icon details %} Trimming with Cutadapt
+>
+> One of the biggest advantage of Cutadapt compared to other trimming tools (e.g. TrimGalore!) is that it has a good [documentation](https://cutadapt.readthedocs.io) explaining how the tool works in detail.
+>
+> Cutadapt quality trimming algorithm consists of three simple steps:
+>
+> 1. Substract the chosen threshold value from the quality value of each position
```
shiltemann:
```suggestion
> 1. Subtract the chosen threshold value from the quality value of each position
```

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -22,6 +22,7 @@ follow_up_training:
     tutorials:
       - mapping
 time_estimation: "1H"
```
shiltemann: is this still realistic with all the extra content?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -170,140 +157,304 @@ To estimate sequence quality along all sequences, we now use [FastQC](https://ww
...
+- the median value, represented by the central red line
+- the inter-quartile range (25-75%), represented by the yellow box
+- the 10% and 90% values in the upper and lower whiskers
+- the mean quality, represented by the blue line
+
+The y-axis shows the quality scores. The higher the score, the better the base call. The background of the  graph divides the y-axis into very good quality scores (green), scores of reasonable quality (orange), and reads of poor quality (red).
+
+It is normal with all Illumina sequencers for the median quality score to start out lower over the first 5-7 bases and to then rise. The quality of reads on most platforms will drop at the end of the read. This is often due to signal decay or phasing during the sequencing run:
+
+- Signal decay
```
shiltemann: explanation of signal decay and phasing maybe in a details box?

topics/sequence-analysis/tutorials/quality-control/tutorial.md
shiltemann: good idea, I think I might be able to find us a good dataset, will look into it (but let's not wait for it here)

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: Maybe not...

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: Yeah maybe a detail box is better there

topics/sequence-analysis/tutorials/quality-control/tutorial.md
bebatut: yeah maybe better

shiltemann:
thanks a lot @bebatut for the enhancements, and thanks @bernt-matthias, @erxleben @anuprulez for the reviews!

PR #1523, opened 2019-07-23
Fix topic table display for introduction slides

bebatut:
To fix the display of the first line (Introduction) in topics:

Screenshot 2019-07-23 at 17 26 21

And also remove the line extension if no Galaxy instances:

Screenshot 2019-07-23 at 17 47 13


shiltemann:
thanks!

PR #1528, opened 2019-07-31
travis: use language: ruby

shiltemann:
When our travis jobs are assigned a xenial machine, the wrong version of ruby was being picked up in the tests (e.g. like in https://github.com/galaxyproject/training-material/pull/1527) this appears to fix that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 ---
 sudo: required
-language: python
-python: 3.6
+language: ruby
+rvm:
+  - 2.4.4
 branches:
   only:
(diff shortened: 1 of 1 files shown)


erasche:
Discussed with @shiltemann out of band, conclusion was that we aren't certain that we can ensure a truly clean path that doesn't include anything from Travis, so the easiest solution is to force Travis to use that version of ruby so at least we don't get any conflicts as a result of different versions having different packages, and it doesn't matter as much if something is installed to the conda environment, or some system location.

PR #1529, opened 2019-07-31
WIP: Minor amplicon tutorial improvements

bernt-matthias:
Just going through the amplicon analysis tutorial and found a few point that might be improved. 

Not finished yet.


bernt-matthias:
Don't like that at the moment

- Background: Rarefaction
- Background: What are Operational Taxonomic Units (OTUs)?

appear multiple times.

shiltemann:
@bernt-matthias I agree, but they appear for the first time in the Mock community section, which is optional. The information is needed at that point for optimal understanding, but for those who skipped the section the information should also be shown at some point. 

I am not sure of a good way around it for now, do you have any suggestions? My thinking was that by using a comment box, the background section is at least delineated clearly and easy to skip over>

bernt-matthias:
@shiltemann agreed

bernt-matthias:
Just another question: I remember that previously there was a statistical analysis at the end of the tutorial. Has this been moved somewhere else?

bernt-matthias:
Also population level analysis seems gone?

shiltemann:
hi @bernt-matthias I moved it to the extra-credit.md file for the moment. We can add it back in as long as we change all the text (some of it is still copied from the mothur SOP). But perhaps it's better as a separate follow-up tutorial since it's already quite long right now. what do you think?

bernt-matthias:
Separate sounds better.

bernt-matthias:
For the duplicated rarefaction curves: The later one in the `Diversity Analysis` section one could have an extra text: 

How i.e. evenness and richness affect the curves: 

- richness -> asymptotic value
- evenness -> how long it takes until the asymptote is reached

bernt-matthias:
A probably more important point: In alpha diversity analysis the workflow uses `summary.single`

- shared file from Make.shared (i.e. not the subsample.shared)
- sets the size parameter to 2389

In the mothur docs the size parameter is documented as:

> size: Within the suite of calculators available in mothur are a set that will predict the number of 
> additional OTUs that will be observed for a given sample size. By default these calculators will base > the prediction on a sample that is the same size as the initial sampling. 

The Galaxy tool documents this as: 

> size - enter the size of the sample, if not set mothur will use the size of your smallest group

Sound like subsampling, but I think its not. There is an additional subsample parameter which should probably be set to `true`. Note that the mothur SOP does exactly this `summary.single(shared=stability.opti_mcc.shared, calc=nseqs-coverage-sobs-invsimpson, subsample=T)`

An alternative might be to use subsample.shared as input?

bernt-matthias:
Just tested with the subsampled input .. quite similar results

Subsampling in the tool does not work at the moment: https://github.com/galaxyproject/tools-iuc/pull/2539

shiltemann:
Hey @bernt-matthias, working through open PRs and came across this, I think there were some things left on the wishlist for this, but perhaps we should just merge what is here? What do you think?

bernt-matthias:
> Hey @bernt-matthias, working through open PRs and came across this, I think there were some things left on the wishlist for this, but perhaps we should just merge what is here? What do you think?

Fine with me. Not sure about the conflicts.

shiltemann:
Thanks @bernt-matthias :+1:  I will take care of the conflicts

shiltemann (approved):

PR #1531, opened 2019-08-01
Fix build ordering

erasche:
The monthly cron job deployed to gh-pages, because that task was second. I've moved it to be first so that the proper version gets deployed to gh-pages before the custom version is built in `bin/publish-archive`.


shiltemann:
thanks!

PR #1540, opened 2019-08-15
Change the default listen host

erasche:
Several people have reported issues and the solution is the same, tell them to the anycast IP address. Given that jekyll is a read-only service that's being run, and everyone contributing is intending to share this content anyway, I propose that we default to listening on this address. The impact should be relatively small.

Came up in gitter with @bernt-matthias and a few others.


bernt-matthias:
Or should it be HOSTNAME?

erasche:
we could do `$(shell hostname)` but I am less confident in this option. The proposed change has fixed the issue for the last three times it came up

Makefile
```diff
@@ -1,7 +1,7 @@
 # Settings
 JEKYLL=jekyll
 PORT?=4000
-HOST?=localhost
+HOST?=0.0.0.0
```
nsoranzo: I think this will break `SITE_URL` env at line 10 below.

Makefile
```diff
@@ -1,13 +1,14 @@
...
 FLAGS?=""
 CHROME=google-chrome-stable
 TUTORIALS=$(shell find _site/training-material -name 'tutorial.html' | sed 's/_site\/training-material\///')
 SLIDES=$(shell find _site/training-material -name 'slides.html' | sed 's/_site\/training-material\///')
 SLIDES+=$(shell find _site/training-material/*/*/slides/* | sed 's/_site\/training-material\///')
-SITE_URL=http://${HOST}:${PORT}/training-material
+PDF_HOST?=localhost
```
nsoranzo: Would `127.0.0.1` be a safer alternative?

bgruening: ping @erasche

erasche: done

PR #1541, opened 2019-08-16
Small fix to the job resource params selector docs

hexylena:
Thanks to a comment from @charmichaeld
https://github.com/galaxyproject/galaxy/issues/6232

### DIFF
--- a/topics/admin/tutorials/connect-to-compute-cluster/tutorial.md
+++ b/topics/admin/tutorials/connect-to-compute-cluster/tutorial.md
@@ -814,7 +814,7 @@ Lastly, we need to write the rule that will read the value of the job resource p
 >            # validate params
 >            cores = int(param_dict['__job_resource']['cores'])
 >            time = int(param_dict['__job_resource']['time'])
->            destination_id = DESTINATION_IDS[cores]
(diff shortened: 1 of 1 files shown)


nsoranzo:
Are we sure it was actually an issue? I think the original code is correct, while this change probably generates an infinite loop.

topics/admin/tutorials/connect-to-compute-cluster/tutorial.md
```diff
@@ -814,7 +814,7 @@ Lastly, we need to write the rule that will read the value of the job resource p
 >            # validate params
 >            cores = int(param_dict['__job_resource']['cores'])
 >            time = int(param_dict['__job_resource']['time'])
->            destination_id = DESTINATION_IDS[cores]
+>            destination_id = 'dynamic_cores_time' # DESTINATION_IDS[cores]
```
bgruening: @erasche I think this should be something like `DESTINATION_ID.get(cores, 1)`

hexylena:
I've reviewed and merged some updates in #1752 and this is not an issue, I'm not sure why I thought it was. the code is correct as written.

PR #1545, opened 2019-08-22
Rename scRNA preprocessing directory

mtekman:
Literally from scrna_preprocessing → scrna-preprocessing.

Just for consistency


erasche:
This was live under the old url right? We should probably include a redirect or so?

bgruening:
Mh, indeed. Would a general redirect to a neural site be better ... lets say a general page that says this tutorial does not exist anymore you can look up our archive or please have a look at the main site in case the tutorial was just renamed?

erasche:
That would be easier to implement for sure, but I guess less 'fair' :/ We should probably do that in addition, for anything that was previously renamed without a redirect.

bgruening:
> Edit: we have a 404 page, just need to include links to the archive + maybe search

I like that!

PR #1547, opened 2019-08-27
Update ecoli comparison for JBrowse

hexylena:
Per GCC discussion with @nekrut 

- [x] replaced google sheets with in-galaxy work
- [x] fixed issue with lastz run (columns transposed)
- [x] reformatting for standards (videos, icons, etc)
- [x] ~updated to use JBrowse for MAF viz~ Just using BED
  - [x] ~waiting on https://github.com/galaxyproject/tools-iuc/pull/2551~
- [x] Circos

I don't think we'll use apollo here, no need for interactive annotation, just for the viz portion. I'll see if we can make something nice for circos though.


erasche:
Instead of IGV now includes a live copy of JBrowse with the final data that the users can analyse and play with:

image

erasche:
And circos!

erasche:
Moved it back to WIP because new circos version will have a MAF-to-links tool and would be better to use that maybe?

But otherwise free to review if anyone feels motivated

erasche:
Ready for review. Would you have time or interest @Delphine-L?

I'm sorry I didn't coordinate more for this, Anton talked to me and I didn't think it through.

shiltemann:
awesome, the embedded Jbrowse is amazing :+1:

shiltemann:
thanks @hexylena!

PR #1550, opened 2019-08-28
Update zenodo for chip seq xi

erasche:
Per a user complaint, cc @bgruening


bgruening:
Awesome, we ping jekyll to <4, which gets installed, but something is pulling in jekyll-4 just a few lines later https://travis-ci.org/galaxyproject/training-material/jobs/577850831#L1451

bebatut:
It seems like there is again a conflict between the gem dependencies. I am investigating

bgruening:
@bebatut it seem we are not using the lock file, can this be?

bebatut:
we are not using the lock file on Travis, just for Heroku

PR #1551, opened 2019-08-28
Update to Jekyll 4

bebatut:
As it seems, we will need to update to [Jekyll 4](https://jekyllrb.com/news/2019/08/20/jekyll-4-0-0-released/) at some point. The website should build quicker as the cache is enable by default now.

We need to:
- [x] Remove all `{{ site.baseurl }}` before `{% link ...`
- [x] Update `Gemfile`, `Gemfile.lock`

   Need to wait for `jekyll-github-metadata` to relax their dependency

I started the previous tasks.

Jekyll is also using environment variables for production and it seems to be the recommended way for Google Analytics. So I tried to get it and then avoid to rely on an extra plugin which was not updated in the last 4 years (and then not sure when it would be ready for Jekyll 4)


bebatut:
Sorry for pinging everyone!

bgruening:
Just as a heads-up ... we do have Jekyll now in conda as well, at least Jekyll 3.x. 4.x will come in the next days. So we might want to go for conda only in the future. Is this is possible with Heroku?

erasche:
> Is this is possible with Heroku?

unfortunately no, we had to re-add the gemfile to get it working under heroku.

martenson:
Heroku is app deployment platform, not a "here is your VM do what you want"-thingy. So we need to stick to the standard ruby deployment to use it.

erasche:
@bebatut is the github-metadata dependency still an issue? https://github.com/jekyll/github-metadata/pull/164 I think you also mentioned you opened a PR to fix one of dependencies as well?

bebatut:
The problem with `jekyll-github-metadata` is that we need a release (https://github.com/jekyll/github-metadata/issues/169) to be able to install it with `gem install`. I need to check if there will be another way to install it without waiting for them to make a release

erasche:
> I need to check if there will be another way to install it without waiting for them to make a release

we could always fork the gem + upload to ruby gems ourselves? Or just build the gem file and leave it on some server somewhere :joy:

or maybe this would work? https://github.com/rdp/specific_install which gives you a command to install from git repo?

erasche:
Or, the only thing it looks like we're using it for is the repo URL? Since we're not encouraging the forking workflow, maybe we could just set this in `_config.yml` and then not need this extra gem anymore?

nsoranzo:
> or maybe this would work? https://github.com/rdp/specific_install which gives you a command to install from git repo?

This looks the more flexible solution to me.

erasche:
My personal preference is removing the gem. Maybe it isn't optimal to be relying on a gem where they can't find the time to release in 8-9 months? Especially if we cannot upgrade without some weird solutions, and it isn't critical and could be replaced with a `_config` var?

shiltemann:
I rewrote things a bit so that we don't need `jekyll-github-metadata` anymore, merged in master and updated all remaining links, and seems to work for me now.

shiltemann:
would be great if somebody else could test this too @hexylena @bgruening (@bebatut)?

bgruening:
rb-jekyll-github-metadata was released and conda packages have been just created.

I like this PR, it makes certain things even easier. Let me test this real quick with conda.

bgruening:
Works like a charm!

The gem installation takes some time in the conda env. I guess it would make sense to install and pin jekyll and some deps as well with conda. But this can be done later.

Gracias!

PR #1572, opened 2019-09-13
InterMine tutorial

danielabutano:
Added InterMine training material during 1st Elixir-UK Hackathon.
Thank you @nsoranzo for your help.
Please @yochannah, correct and update!


danielabutano:
Hi @nsoranzo. I have completed the intermine tutorial. It's ready to be published. Thank you

topics/galaxy-interface/tutorials/intermine/tutorial.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Create InterMine Interchange dataset
+
+We will use **Create InterMine Interchange Dataset** {% icon tool %} in order to generate an intermediate file which will be used to send the identifiers (e.g. gene identifiers) to InterMine. This file requires the identifier's type (e.g. `Gene`), the identifier (e.g `WBGene00007063`) and, optionally, the organims's name.
+
+> ### {% icon hands_on %} Hands-on: Generate InterMine file
+>
+> 1. {% tool [Create InterMine Interchange dateset](toolshed.g2.bx.psu.edu/repos/iuc/intermine_galaxy_exchange/galaxy_intermine_exchange/0.0.1) %} with the following parameters:
+>    - {% icon param-file %} *"Tabular file"*: select the `GenesLocatedOnChromosome4` dataset which contains some fly's genes
+>    - *"Feature Type Column"*: select a column from the input file which contains the identifer's type you are exporting to InterMine.
```
hexylena, line 114: Hi @danielabutano I could figure out what column should be selected here?

It's a very generic instruction, and in the GTN we usually prefer very specific ones (select col 1). Which column should be selected here?

danielabutano: Hi @hexylena. thanks for reviewing this PR.
In this case you don't have to select any column and (as specified in the following sentence)

> because the `GenesLocatedOnChromosome4` dataset does not containt the type we have to edit it, in the *"Feature Type"*

hexylena: Ah ok, this was not clear for me from my reading. Is there any chance you could make a screenshot of the interface? That would help me ensure there are no misunderstandings for the other parameters (if i misunderstood this one, maybe i misunderstand the other descriptions) 

Then I can then write it in the GTN's preferred format/style.

danielabutano: image

hexylena: thanks!

danielabutano: I try to explain in a different way. 
We need to provide Feature Type and Feature Identifier (Organism is not mandatory).
These values can be extraxted from the dataset provided (using the column selected in Feature Type column or Feature identifier column) or manually edited (in the Feature Type or Feature Identifier fields).
If you edit the value (in the Feature Type or Feature Identifier fields), the column selected is irrilevant.

hexylena: That makes sense. Does it function if both fields are specified (e.g. feature identifier column + feature identifier?) or what's the behaviour in that case?

danielabutano: If both fields are specified, the field edited wins (as Feature Type column and Feature type in the screenshot).
This behavior is a bit weird, I know, it would be much much more clear, if the user could select "Nothing selected" in the column and edit "Gene" in the field

hexylena: Maybe we could fix that in the tool, provide the user a conditional like https://github.com/galaxyproject/tools-iuc/blob/1f35303af979c16d9a3126dbc882a59f686ace5d/tools/tbl2gff3/tbl2gff3.xml#L48 that one? This way we can ensure no unintentional/bad states.

hexylena (approved):
I think this looks great for a start and it's been sitting around too long, let's get it in :)

topics/galaxy-interface/tutorials/intermine/tutorial.md
danielabutano: I agree with you that something like that might be more clear for the user

danielabutano:
Great! Thank you!

PR #1594, opened 2019-09-17
Add script to check contributors

hexylena:
Wrote this as a break between papers, it will check the git log of a given file for the contributors which could be missing.

There is commonly the case of a user using multiple emails, and so this script can read from a local map of emails → usernames.

We could just add these extra emails to the yaml file, but given my personal perspective on the issue, I think maybe it's fine just to have a local map since not so many people might be running this to find missing contributors. (E.g. I know if I have old emails addresses that were included as commits, I'm not sure I want those in a more public contributors.yaml file and I'm sure I'm not alone.)

@bebatut

### DIFF
--- a/bin/check-contributors.rb
+++ b/bin/check-contributors.rb
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
 require 'yaml'
-require 'pathname'
 
 fn = ARGV[0]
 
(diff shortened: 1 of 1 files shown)


bebatut:
Thanks @erasche 
How does it work? Do we need to run it manually? If yes, how?

erasche:
@bebatut yes you need to run it manually, it was a privacy tradeoff vs keeping a long list of alternate emails for that user publicly.

you can run with `ruby bin/check-contributors.rb path/to/tutorial.md` (or slides.html). If there are missing emails you can build a small (local) `private-contrib-map.yaml` with 

```
---
some-email: username
```

erasche:
Ok, wrote some documentation and did something more standard.

shiltemann:
thanks @hexylena !

PR #1597, opened 2019-09-17
10x SC: Only import Cell Barcodes once

blankenberg:


bebatut:
Thanks @blankenberg!
@mtekman could you check these changes? Thanks

PR #1598, opened 2019-09-18
re-arrange the steps in Pre-processing of 10X

hrhotz:
do the alignment first, before describing how to use MultiQC to look at “STAR log output”


bebatut:
Thanks @hrhotz 
@mtekman can you check the changes? Thanks

PR #1600, opened 2019-09-19
question about strandedness of the library

hrhotz:
Although "Forward" is the default in STARSolo tool, it is probably good to discuss the strandedness at one point in this tutorial


mtekman:
Good point. I will factor this in to this tutorial, or maybe add a slide deck so that it can be referenced in others.

mtekman:
hello is this issue now resolved?

Issue #1601, opened 2019-09-20
Pre-processing of 10X scRNA Datasets: number of "nNoMatch" is unusually high


hrhotz:
Hi

I was intrigued by the high number of "nNoMatch": 7491731 which corresponds to roughly 1% of valid barcodes (btw. running the tool locally gives me the same numbers). 
First I thought, this could be an artefact caused by the subsampling. Hence I run the complete data set. Similar result, "nNoMatch": 65256325 (valid barcodes ~1%).
I then tried the '3M-february-2018.txt' whitelist, and I got a much better result, "nNoMatch": 50037 (valid barcodes ~99%). The published cell ranger output has: "Valid Barcodes: 97.4%"

So I am wondering whether, a different whitelist should be usedin this example?

Regards, Hans-Rudolf

bgruening:
ping @mtekman

mtekman:
Wow yes. I assumed the 737K barcodes did not change between chemistries, and that the number of mismatches was just high by design.

Clearly that is wrong, and I will definitely use the more recent barcodes --- thanks!

mtekman:
This could be a good learning exercise for knowing whether or not you have right barcodes. I might keep the old ones in the tutorial as well for this reason.

mtekman:
hello is this issue now resolved?

hrhotz:
yes, it is all included in #1603

PR #1603, opened 2019-09-20
Fixes to 10x workflow

mtekman:
Added strandedness explanation, updated barcodes, zenodo record based on recommendations from @hrhotz 

This incorporates fixes mentioned:
- https://github.com/galaxyproject/training-material/pull/1600
- https://github.com/galaxyproject/training-material/issues/1601


hrhotz:
General Question: The data (i.e.: the fastq files) is provided under "https://creativecommons.org/licenses/by/4.0/" by 10X. As far as I know, the barcode files are part of the software, aren't-they? Is the software also distributed under  'CC BY 4.0' ?

mtekman:
As far as I understand, these files come with cell ranger and so are distributed with the cell ranger license which is MIT

https://github.com/10XGenomics/cellranger/blob/master/LICENSE

hrhotz:
I am not an expert on open source license, especially, I am always confused wrt the hierarchy between them: If the barcode files are distributed under "MIT", is it then allowed to re-distribute them under  'CC BY 4.0'  on zenodo?

mtekman:
I think it would be enough to serve both licenses in the Zenodo url, but we probably wouldn't need to mention it in the tutorial. Would that work?

nsoranzo:
@hrhotz You can mix MIT-licensed files with stuff under any other license, as long as you distribute the MIT license file with them.

hrhotz:
@nsoranzo Thank You for providing a solution!

mtekman:
@hrhotz good spot, will fix

mtekman:
@hrhotz so I updated the Zenodo link to include an MIT license, and added a link to CC. In the material I changed the cell numbers, and added a section about whitelist barcodes.

PR #1604, opened 2019-09-20
add emoji flags to translations

martenson:
not sure what flag can Arabic have

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -25,8 +25,8 @@ gitter_url: "https://gitter.im/Galaxy-Training-Network/Lobby"
 logo: "assets/images/GTN.png"
 small_logo: "assets/images/GTN-60px.png"
 help_url: https://help.galaxyproject.org/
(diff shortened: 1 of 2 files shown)


martenson:
A_short_introduction_to_Galaxy

bebatut:
At the beginning we wanted to use flags also. But we were not sure because many countries speak one language and how to choose.
I do not know what de should do

nsoranzo:
> At the beginning we wanted to use flags also. But we were not sure because many countries speak one language and how to choose.

Agreed, it's a can of worms :-1:

martenson:
I see it as more useful, but have no horse here. What about we include francais|اللغة العربية|日本語 next to the code?

lyoussar:
I found the flag attached for Arabic language is appropriate.
Did anyone started the translation to Arabic language?
arabic

martenson:
We have decided flags are not good enough representation and will implement native language name tags instead. Thanks for contributions.

PR #1605, opened 2019-09-21
Update variant analysis tutorials

wm75:
Different possibilities for filtering BAM reads get explained in
detail now in the exome-seq and somatic variants tutorials.
Dysfunctional usage of gemini annotate has been eliminated from the
somatic variants tutorial.

### DIFF
--- a/topics/variant-analysis/tutorials/exome-seq/workflows/workflow_exome_seq_from_premapped.ga
+++ b/topics/variant-analysis/tutorials/exome-seq/workflows/workflow_exome_seq_from_premapped.ga
@@ -3,7 +3,7 @@
     "tags": [],
     "format-version": "0.1",
     "name": "exome_seq_training_short_w_cached_ref",
-    "version": 3,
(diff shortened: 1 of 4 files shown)


topics/variant-analysis/tutorials/exome-seq/workflows/workflow_exome_seq_from_premapped.ga
```diff
@@ -114,207 +114,186 @@
             "type": "data_input"
         },
         "4": {
-            "tool_id": "toolshed.g2.bx.psu.edu/repos/devteam/bamtools_filter/bamFilter/2.4.1",
```
bgruening: anything from with bamFiler?

wm75: No, but the somatic variant tutorial uses BamFilter so I thought we could demonstrate another way of filtering here.

PR #1613, opened 2019-09-25
Ref-based RNA-seq: Fix the solution to extract up-regulated gen…

MiguelCos:
Hi,

We found a typo, please review!


bgruening:
@MiguelCos first contribution, that was fast! :)

PR #1618, opened 2019-10-01
Template: change slightly the display of supporting material

bebatut:
Screenshot 2019-10-01 at 15 07 29

- Put supporting material not as a list but as text
- Tweak a bit the css


nsoranzo:
What about a proper list instead of a single line?

assets/css/main.scss
```diff
@@ -594,36 +594,26 @@ div.highlight:hover .btn,div.highlight .btn:focus{
...
-    div {
-        display: flex;
-        flex-direction: row;
-        align-items: center;
-
-        div {
-            a.topic-icon {
-                padding: 0 .5rem 0 .5rem;
-            }
+div.supporting_materials {
```
erasche: just a comment on why I used `#`, I only ever expected there to be a single instance on any page / vaguely considered people might want to link people directly there but I don't think anyone is doing that? If this makes more sense as a class, cool, not a super css person :)

bebatut: For what I understand, class are usually recommended in css (over id, even for a single element).
I just added an `id` in the "Supporting Materials" icon to have a possible link to this part.

PR #1619, opened 2019-10-06
couple small fixes for 101

erasche:
- minimap appearance was updated and behaves differently, just remove.
- more detailed ucsc for second run since people complained
- tell people to relay out WF since automatic one has overlaps.


shiltemann:
thanks!

PR #1624, opened 2019-10-08
R-lesson: add (temporary) instructions for starting rstudio in Galaxy

shiltemann:
@bgruening oops, forgot to add this file to previous PR

### DIFF
--- a/snippets/launch_rstudio.md
+++ b/snippets/launch_rstudio.md
@@ -1,5 +1,7 @@
 >
->    > ### {% icon tip %} Tip: Launch RStudio
->    > 
+>    > ### {% icon tip %} Tip: Launch RStudio in Galaxy
+>    > 1. Open the Rstudio tool {% icon tool %} by clicking [here](https://usegalaxy.eu/?tool_id=interactive_tool_rstudio)
+>    > 2. Click Execute
+>    > <!-- TODO: update to tool panel search -->
 >    {: .tip}
 >


snippets/launch_rstudio.md
```diff
@@ -1,5 +1,7 @@
 >
->    > ### {% icon tip %} Tip: Launch RStudio
->    > 
+>    > ### {% icon tip %} Tip: Launch RStudio in Galaxy
+>    > 1. Open the Rstudio tool {% icon tool %} by clicking [here](https://usegalaxy.eu/?tool_id=interactive_tool_rstudio)
```
bgruening: keep in mind you can also always use live.usegalaxy.eu -- might be actually useful to explain that those fancy subdomains are still the same server and the histories are the same ...

yvanlebras: arrrrr ? ;) Live is amazing! Plenty of good things to you all eu team for your workshop!

erasche: @bgruening Unfortunately they're not logged in because we don't share the cookie. 

And we have >0 elixir users in the training, so not an option for us currently.

Should we start sharing cookies between subdomains? Firefox's "container" tabs feature is usually enough for me to manage multiple histories/sessions? Or we could encourgage galaxy to implement tab-local histories which would fix it completely

bgruening: Na, we should register the new domain with the ELIXIR people and wait for the share cookie patch that was recently PRed to Galaxy.

bgruening: We could backport this PR if you like...

PR #1626, opened 2019-10-15
Update considering dartaset format from spocc

yvanlebras:


abretaud:
Is this PR ready for review/merge @yvanlebras?

yvanlebras:
ok for me! There is an issue wiht the dockerized Wallace but content of the tuto is ok now normally. I just have to update the WAllace docker image ASAP in a proper way

PR #1627, opened 2019-10-16
add comments for python prerequisites in ansible galaxy installation

qiagu:
I spent some time to figure out the python requirements. The comment might help other people who have similar background to me.

### DIFF
--- a/topics/admin/tutorials/ansible-galaxy/tutorial.md
+++ b/topics/admin/tutorials/ansible-galaxy/tutorial.md
@@ -192,7 +192,8 @@ To proceed from here it is expected that:
 2. Your `ansible` version is `>=2.7`, you can check this by running `ansible --version`
 3. You have an [inventory file](../ansible/tutorial.html#inventory-file) with the VM or host specified where you will deploy Galaxy. We will refer to this group of hosts as "galaxyservers."
 4. Your VM has a public DNS name: this tutorial sets up SSL certificates from the start and as an integral part of the tutorial.
-5. In your inventory file, you have written the full DNS hostname that has been provided, and **not** `localhost`, as we will be requesting SSL certificates.
+5. Your VM has `python2.7` and `python-pip` installed.
+6. In your inventory file, you have written the full DNS hostname that has been provided, and **not** `localhost`, as we will be requesting SSL certificates.
 
 
 ## Requirements


erasche:
@nsoranzo python is missing on modern ubuntu machines. The [workarounds are kinda gross.](https://gist.github.com/gwillem/4ba393dceb55e5ae276a87300f6b8e6f#gistcomment-2096289). (I've linked to a specific one, but scroll through that gist for several variants.)

`python-pip` should be downloaded as part of `python-virtualenv` which is part of https://training.galaxyproject.org/training-material/topics/admin/tutorials/ansible-galaxy/tutorial.html#hands_on-hands-on-minimal-galaxy-playbook-1

```
$ aptitude show python-virtualenv
...
Depends: python, python-pip-whl (>= 8.1.1-2), python-pkg-resources
```

nsoranzo:
> @nsoranzo python is missing on modern ubuntu machines. The [workarounds are kinda gross.](https://gist.github.com/gwillem/4ba393dceb55e5ae276a87300f6b8e6f#gistcomment-2096289). (I've linked to a specific one, but scroll through that gist for several variants.)

@erasche `python` is a requirement of the `ansible` package, so it should be installed as part of the Ansible tutorial, no?

> `python-pip` should be downloaded as part of `python-virtualenv` which is part of https://training.galaxyproject.org/training-material/topics/admin/tutorials/ansible-galaxy/tutorial.html#hands_on-hands-on-minimal-galaxy-playbook-1
> 
> ```
> $ aptitude show python-virtualenv
> ...
> Depends: python, python-pip-whl (>= 8.1.1-2), python-pkg-resources
> ```

`python-pip-whl` != `python-pip` unfortunately. `python-pip-whl` is "the support package for the PEP 427 wheel version of the package, required for using pip inside a virtual environment", and contains only [a bunch of wheels](https://packages.ubuntu.com/bionic/all/python-pip-whl/filelist).
So, would it be a better solution to just add `python-pip` to the part of the tutorial you linked?

erasche:
> python is a requirement of the ansible package

This only happens when you run ansible on the VM or target machine. (And this specific issue is part of the reason we *do* that rather than even giving the option to run remotely.)

> python-pip-whl

Ah I misinterpreted that then. Huh, I really would have expected python-pip to be included in that tree somewhere. Looks like we do it in the dagobah provisioning playbook which is why it was missed the past two times I've given this training. We should definitely add that to the pre-tasks.

PR #1633, opened 2019-10-18
Use `geerlingguy.pip` role to install pip

nsoranzo:
Follow-up on https://github.com/galaxyproject/training-material/pull/1627 .


erasche:
Looks good, thanks for fixing that!

PR #1635, opened 2019-10-21
Update gromacs tutorial after changes to tools

simonbray:
Update after https://github.com/galaxycomputationalchemistry/galaxy-tools-compchem/pull/35.

Also a small correction to the protein-ligand docking tutorial added on Friday.


simonbray:
Whoops. :/

PR #1636, opened 2019-10-22
Fix data-library.yaml for ecology

abretaud:
Just some cleanup to the data-libray.yaml files for the ecology topic (found while trying to use some kind of https://github.com/usegalaxy-eu/shared-data for galaxy.genouest.org)

### DIFF
--- a/topics/proteomics/tutorials/mass-spectrometry-imaging-loading-exploring-data/data-library.yaml
+++ b/topics/proteomics/tutorials/mass-spectrometry-imaging-loading-exploring-data/data-library.yaml
@@ -13,10 +13,11 @@ items:
     - name: 'DOI: 10.5281/zenodo.1560645'
       description: latest
       items:
-      - url: https://zenodo.org/api/files/463f6b73-0c42-4be4-8b52-181a8e093abd/mouse_kidney_cut.ibd
-        src: url
-        ext: ibd
-        info: https://doi.org/10.5281/zenodo.1560645
+# Commented until composite dataset upload is possible (https://github.com/galaxyproject/galaxy/issues/8909)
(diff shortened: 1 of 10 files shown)


yvanlebras:
Thanks

abretaud:
There is still a problem with 2 other types of data:
- [ibd](https://github.com/galaxyproject/training-material/blob/master/topics/metabolomics/tutorials/msi-analyte-distribution/data-library.yaml#L16) which is part of a composite datatype, I wonder if there would be a way to represent a composite library dataset in data-library.yaml?
- [raw](https://github.com/galaxyproject/training-material/blob/master/topics/proteomics/tutorials/protein-id-oms/data-library.yaml#L32): maybe it should be changed to raw_pepxml or thermo.raw? I don't know these proteomic formats

bgruening:
> raw: maybe it should be changed to raw_pepxml or thermo.raw? I don't know these proteomic formats

It is thermo.raw ... just consulted my proteomics expert.

bgruening:
Let's comment this out for now. Can you create an issue for that? Thanks.

topics/proteomics/tutorials/mass-spectrometry-imaging-loading-exploring-data/data-library.yaml
```diff
@@ -13,10 +13,11 @@ items:
     - name: 'DOI: 10.5281/zenodo.1560645'
       description: latest
       items:
-      - url: https://zenodo.org/api/files/463f6b73-0c42-4be4-8b52-181a8e093abd/mouse_kidney_cut.ibd
-        src: url
-        ext: ibd
-        info: https://doi.org/10.5281/zenodo.1560645
+# Commented until composite dataset upload is possible (https://github.com/galaxyproject/galaxy/issues/8909)
```
bebatut: Can you need to indent it as the content? Travis error: `16:1      warning  comment not indented like content  (comments-indentation)`

PR #1643, opened 2019-10-27
Updates in UI of the 404 page.

EshitaNandy:
Some formatting and editing were done to make the page more responsive and understandable.


erasche:
Hi @EshitaNandy, I'm sorry, I don't think we can accept this PR as it is, removing the header and footer is not something we want to do here.

PR #1650, opened 2019-11-06
Add citation to every tutorial page

erasche:
This adds a box which helps users cite the tutorial + training material. the urls should be correct when deployed.

Hopefully it will make it easier for users to find the proper citations.

image

### DIFF
--- a/_layouts/tutorial_hands_on.html
+++ b/_layouts/tutorial_hands_on.html
@@ -225,9 +225,9 @@ layout: base
                     {% endif %}
 
                     <h1>Citing this Tutorial</h1>
-                    <p id="citation-text">
+                    <p>
                         <ol>
-                            <li>
+                            <li id="citation-text">
                             {%- include _includes/contributor-author-list-comma.html contributors=page.contributors -%}, {{ page.last_modified_at | date: "%Y" }} <b>{{ page.title }} (Galaxy Training Materials)</b>. <a href="{{ site.baseurl }}{{page.url}}">{{ site.baseurl }}{{page.url}}</a> Online; accessed TODAY
                             </li>
                             <li>
--- a/_layouts/tutorial_hands_on.html
+++ b/_layouts/tutorial_hands_on.html
@@ -98,7 +98,7 @@ layout: base
         {% endif %}
 
         <div class="contributors-line">By: {% include _includes/contributor-badge-list.html contributors=page.contributors %}</div>
-        
+
         <blockquote class="overview">
             <h3>Overview</h3>
             <strong>{% icon question %} Questions</strong>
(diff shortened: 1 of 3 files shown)


bebatut:
Nice!

Something that came up during some discussion yesterday, what about having contributors always listed by alphabetical order (and mention it)?

_layouts/tutorial_hands_on.html
```diff
@@ -224,6 +224,58 @@ <h1 id="bibliography">References</h1>
...
+        author = {B{\'{e}}r{\'{e}}nice Batut and Saskia Hiltemann and Andrea Bagnacani and Dannon Baker and Vivek Bhardwaj and Clemens Blank and Anthony Bretaudeau and Loraine Brillet-Gu{\'{e}}guen and Martin {\v{C}}ech and John Chilton and Dave Clements and Olivia Doppelt-Azeroual and Anika Erxleben and Mallory Ann Freeberg and Simon Gladman and Youri Hoogstrate and Hans-Rudolf Hotz and Torsten Houwaart and Pratik Jagtap and Delphine Larivi{\`{e}}re and Gildas Le Corguill{\'{e}} and Thomas Manke and Fabien Mareuil and Fidel Ram{\'{\i}}rez and Devon Ryan and Florian Christoph Sigloch and Nicola Soranzo and Joachim Wolff and Pavankumar Videm and Markus Wolfien and Aisanjiang Wubuli and Dilmurat Yusuf and James Taylor and Rolf Backofen and Anton Nekrutenko and Björn Grüning},
+        title = {Community-Driven Data Analysis Training for Biology},
+        journal = {Cell Systems}
+}</code>
+                    </pre></div></div>
+                    </p>
+                    </blockquote>
+
+
+<script type="text/javascript">
```
bebatut: Could the script there be in `_layouts/base.html`?

_layouts/tutorial_hands_on.html
```diff
@@ -224,6 +224,58 @@ <h1 id="bibliography">References</h1>
...
+    author = "{%- include _includes/contributor-author-list.html contributors=page.contributors -%}",
+    title = "{{ page.title }} (Galaxy Training Materials)",
+    year = "{{ page.last_modified_at | date: "%Y"}}",
+    month = "{{ page.last_modified_at | date: "%m"}}",
+    day = "{{ page.last_modified_at | date: "%d" }}"
+    url = "{% raw %}\url{{% endraw %}{{ site.baseurl }}{{page.url}}{% raw %}}{% endraw %}",
+    note = "[Online; accessed TODAY]"
+}
+@article{Batut_2018,
+        doi = {10.1016/j.cels.2018.05.012},
```
bebatut: Could you use the indentation as above?

_layouts/tutorial_hands_on.html
```diff
@@ -224,6 +224,58 @@ <h1 id="bibliography">References</h1>
...
                     {% endif %}
 
+                    <h1>Citing this Tutorial</h1>
+                    <p>
+                        <ol>
+                            <li id="citation-text">
+                            {%- include _includes/contributor-author-list-comma.html contributors=page.contributors -%}, {{ page.last_modified_at | date: "%Y" }} <b>{{ page.title }} (Galaxy Training Materials)</b>. <a href="{{ site.baseurl }}{{page.url}}">{{ site.baseurl }}{{page.url}}</a> Online; accessed TODAY
+                            </li>
+                            <li>
+                            Batut et al., 2018 <b>Community-Driven Data Analysis Training for Biology</b> Cell Systems <a href="https://doi.org/10.1016%2Fj.cels.2018.05.012">10.1016/j.cels.2018.05.012</a>
```
bebatut: Is it possible to use things as the citations in the tutorial?

_layouts/tutorial_hands_on.html
erasche: No, cannot in a `pre` block. It must be left aligned.

_layouts/tutorial_hands_on.html
erasche: No, it is not, unfortunately. We cannot easily call that code from within a template, or we would have to modifying the content generation pipeline to generate an extra bib file for every single tutorial, and it would need to be early enough in the pipeline that scholar could discover it, it really would not be easy unfortunately.

_layouts/tutorial_hands_on.html
erasche: It could be, but that snippet is only useful in conjunction with the citation blocks with those identifiers. It will generate a console error on any non-tutorial page, if loaded in the base, since those pages lack matching elements.

erasche:
> what about having contributors always listed by alphabetical order (and mention it)?

I'm currently fine with the ordering in recency of contribution. I heard a while back about the bias alphabetising last names can have so I don't feel motivated for it to change.

If we do change the sorting, I'd rather sort by # of characters changed in an article or some similar. That still misses key points of correcting some substantial error with a small diff, but, nothing is perfect.

See [What's in a surname? The effects of surname initials on academic success](http://web.stanford.edu/~leinav/pubs/JEP2006.pdf)

> found that you were more likely to win a Clarke Medal, a Nobel Prize and to achieve tenure as an economist if you were born in the first half of the alphabet. Interestingly, they performed the same test with tenure in psychology, and found that that relationship disappeared, and there is a very simple reason for that: authorship on papers in the field of economics are generally sorted alphabetically, meaning that people who have initials closer to the start of the alphabet are more likely to be first author and thus, more likely to open up new opportunities for them. However, this practice tends to be more of an exception than the rule in academia since, in the field of psychology (and biology), authorship order is generally granted based on merit and not on alphabetic placement. 
>
> *from http://drawcuriosity.com/2016/08/why-browns-do-better-than-smiths/*

shiltemann:
thanks @erasche!!

PR #1654, opened 2019-11-18
Broken box detection

erasche:
This PR implements a makefile target which checks that in every enabled tutorial:

- There are no blockquotes without classes (commonly caused with accidental line breaks in the middle of a hands on that isn't noticed)
- There are no blockquotes with unexpected/wrong classes (e.g. typos)

This is added as a mandatory travis test. To make it more pleasant for contributors, it attempts to locate the usage of the broken box and greps for context to help users to locate the issue.

In a pretend world where the matrix class didn't exist, this will show the relevant source code and files.

image

I've fixed many boxes in existing tutorials, and changed some usages to be more correct.

**Edit** maintainers were pinged due to changes to tutorials to bring them into compliance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,10 @@ check-framework:
 		ruby _plugins/jekyll-notranslate.rb
 .PHONY: check-framework
 
+check-broken-boxes: build ## List tutorials containing broken boxes
+	./bin/check-broken-boxes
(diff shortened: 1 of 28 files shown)


abretaud:
@yvanlebras this one is for you!

PR #1667, opened 2019-11-21
Fix wrong configuration of mapping quality filter in somatic variants tutorial

wm75:
Thanks @monfortm for reporting!


shiltemann:
thanks @wm75 and @monfortm !

PR #1668, opened 2019-11-21
Cleanup json ld for tess

erasche:
- Remove it from topics, just on tutorials, no more duplication
- Remove some of the extra fields that aren't being consumed or used (e.g. workflows are just Things, if we had a better class for them maybe it would be worth it, but since no one is going to consume then just useless a bit)
- Hardcode URL since we want going through the right site


shiltemann:
thanks!

PR #1676, opened 2019-11-26
Fix minor issues in the slides of mapping tutorial

anuprulez:


topics/sequence-analysis/tutorials/mapping/slides.html
```diff
@@ -964,7 +964,7 @@
 
 *“... there is no tool that outperforms all of the others in
 all the tests. Therefore, the end user should clearly
-specify his needs in order to choose the tool that
+specify [her/]his needs in order to choose the tool that
```
nsoranzo:
```suggestion
specify [their] needs in order to choose the tool that
```

anuprulez: @nsoranzo if we use `their` then we should also modify at the start of the sentence which reads as `the end user ...` and change to `the end users ...`

nsoranzo: @anuprulez No, it's a [singular their](https://en.wikipedia.org/wiki/Singular_they).

anuprulez: @nsoranzo thanks for your suggestion :)

shiltemann:
nicely spotted @anuprulez thanks!

PR #1679, opened 2019-11-27
Reference based RNA seq tutorial: some rephrasing and formatting

bebatut:
ping @nsoranzo, @mvdbeek and @mblue9


topics/sequence-analysis/tutorials/mapping/bam_explanation.md
```diff
@@ -1,4 +1,4 @@
-A BAM ([Binary Alignment Map](https://en.wikipedia.org/wiki/SAM_(file_format))) file is a compressed, binary file storing the sequences mapped to a reference sequence. 
+A BAM ([Binary Alignment Map](https://en.wikipedia.org/wiki/SAM_(file_format))) file is a compressed binary file storing the read sequences, whether they have been aligned to the reference sequence, and if so, the position on the reference sequence at which they have been aligned.
```
nsoranzo:
```suggestion
A BAM ([Binary Alignment Map](https://en.wikipedia.org/wiki/SAM_(file_format))) file is a compressed binary file storing the read sequences, whether they have been aligned to a reference sequence (e.g. a chromosome), and if so, the position on the reference sequence at which they have been aligned.
```

topics/transcriptomics/tutorials/ref-based/tutorial.md
```diff
@@ -1113,7 +1106,7 @@ This expression analysis is estimated from read counts and attempts are made to
 > We recommend to combine the count tables for different technical replicates (but not for biological replicates) before a differential expression analysis (see [DESeq2 documentation](http://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#collapsing-technical-replicates))
 {: .details}
 
-Multiple factors with several levels can then be incorporated in the analysis. After normalization we can compare the response of the expression of any gene to the presence of different levels of a factor in a statistically reliable way.
+Multiple factors with several levels can then be incorporated in the analysis representing known sources of variation (e.g. treatment, tissue type, gender, batches), with several levels, representing the conditions for one factor. After normalization we can compare the response of the expression of any gene to the presence of different levels of a factor in a statistically reliable way.
```
nsoranzo:
```suggestion
Multiple factors with several levels can then be incorporated in the analysis describing known sources of variation (e.g. treatment, tissue type, gender, batches), with several levels representing the conditions for one factor. After normalization we can compare the response of the expression of any gene to the presence of different levels of a factor in a statistically reliable way.
```

PR #1680, opened 2019-11-30
Fix some typo + add a temporary manual step to do to fix an issue

yvanlebras:


abretaud:
Anything I can do to help fix wallace? Otherwise I'll merge as is

yvanlebras:
Thank you Breutaudeau ;) hope all is fine for you! In fact there is a last issue with Wallace but with the tip mentionned on the PR it works fine! So for now it's mergeable and useable like this. Thank you so much!!!!

abretaud:
Ok, let's go! La bise

PR #1683, opened 2019-12-10
R lesson: add import box for newcomers

shiltemann:
one forgotten commit


erasche:
thanks for doing this!

PR #1685, opened 2019-12-10
For every topic a working Docker image

bedroesb:
Updated the docker image links and pushed a new docker image for every topic based on the latest version of GTN.


bedroesb:
The topic *computational chemistry* already had a docker image `docker_image: "scientificomputing/bridge"`  in the metadata.yaml file. This will now be replaced with an updated workflow central docker image.

bedroesb:
@bgruening Although I found out that not all tools are installed based on the workflows (I am working on an issue/PR about this), this will not affect this PR towards the links to the docker images itself.

simonbray:
Hi @bedroesb @bgruening thanks a lot for working on this. How can these images be updated when new tutorials are added?

bedroesb:
It would be nice to have a Jenkins Job that builds these images every month, when a new GTN release is released

bedroesb:
Until a few months ago I did it manually every month, but since I am working from home and don't want to put to much stress on my home pc, they are getting a bit out of sync...

nsoranzo:
A GitHub workflow could also be used.

bedroesb:
I already tried that but the builds take to long / or to big in size -> transcriptomics is more than 2h + around 20gb

PR #1689, opened 2019-12-11
fix typo

yvanlebras:


snippets/launch_rstudio.md
```diff
@@ -18,4 +18,4 @@
 > > 2. Create a new project
 > {: .tip}
 >
-{: .hands_on}
\ No newline at end of file
+{: .hands_on}
```
nsoranzo: This should not be here, can you run `make check-snippets` and commit the change that should remove the newline?

yvanlebras: oups. Sorry Nicola, don't understand... I made something wrong ?

shiltemann: newlines at the ends of snippets can sometimes mess up formatting of tutorials, so we run some automatic travis tests to check for it, which is why tests are failing now. If you run `make check-snippets` it will detect this and tell you the command to run to fix it:

```
perl -pi -e 'chomp if eof' snippets/*
```

(I did the fix now, but just in case you were wondering)

PR #1709, opened 2020-01-10
Fixing multiple workflows

bedroesb:
Multiple tools were not detected by ephermeris 'workflow-to-tools'. It turned out that this was caused by the lack of a "tool_shed_repository" key in the step of the workflow. I made a script that detects and fixes these problems where possible. I am still trying out the workflows for a manual check, but the results seem already promising.

- Added tool_shed_repository keys were possible in following workflows:
-> CTB workflow
-> Find Exons with the highest number of SNPs (or other features)
-> Database Handling Tutorial Workflow2 including Mycoplasma
-> Database Handling Tutorial Workflow
-> tails triple dimethyl OpenMS2.1
-> ProteinID_SG_PS_Tutorial_WF_datasetCollection
-> Peptide and Protein ID Tutorial
-> Secreted Proteins via GO annotation and WoLF PSORT for shCTSB paper ( `get_subontology_from` and `term_id_vs_term_name` not directly replacable with `onto_tk_term_id_vs_term_name` and `onto_tk_get_subontology_from`, help needed. Will be added to #1710 )
-> asdf
- fixed duplicate in workflow name 'Identification of the binding sites of the Estrogen receptor'
- ewas tour was empty and giving problems if installed on galaxy
- fixed duplicate in workflow name 'GTN - Sequence analyses - Mapping'

**NOTE: Not yet to be merged, still to be tested**

PR #1713, opened 2020-01-13
Editted CustomProDB tool version to tool ID

subinamehta:


shiltemann:
thanks @subinamehta :) I believe @bedroesb already made this change in https://github.com/galaxyproject/training-material/pull/1709/files

PR #1718, opened 2020-01-16
[Antibiotic Resistance Detection] use fewer input files

shiltemann:
at GCC this tutorial caused some issues when there were a lot of participants. To reduce the number of jobs generated, I cut the number of files in half, using only those that are explicitly mentioned in the tutorial (e.g. in questions or explanations)

ping @willemdek11 can you have look to see if this is ok?


willemdek11:
@shiltemann Yes looks good!

PR #1724, opened 2020-01-22
Workflow annotation

bedroesb:
To have a better link between workflow and tutorial, I placed the title of the tutorial in the annotation section of the workflow. This also serves as some extra information in many situations (helpful if people just scroll trough the workflows on a galaxy instance). This can be a confirmation that the workflow belongs to a specific tutorial.


topics/ecology/tutorials/ref-based-rad-seq/workflows/Galaxy-Workflow-Workflow_constructed_from_history__STACKS_RAD__population_genomics_with_reference_genome_.ga
```diff
@@ -1002,7 +1002,7 @@
           "type": "tabular"
         },
         {
-          "name": "out_fastpĥase",
+          "name": "out_fastp\u0125ase",
```
bgruening: uha, what is this ....

topics/ecology/tutorials/ref-based-rad-seq/workflows/workflow.ga
```diff
@@ -624,7 +624,7 @@
           "type": "tabular"
         },
         {
-          "name": "out_fastpĥase",
+          "name": "out_fastp\u0125ase",
```
bgruening: same here ...

bedroesb:
Nice catch, the non ascii characters are wrongly dumped by the json library in python. @hexylena already explained me that these are linked to the version of this tool. Maybe it would be a good idea to change to the newer version of this tool solving this?

bgruening:
Tool update sounds nice too me ... we just to ensure the tutorial works. But I guess we can ask @yvanlebras for help.

hexylena:
Do you want to update your linter to check for this as well? It's a nice feature and would be good if future workflows conformed.

bin/validate-workflow-tags.sh
```diff
@@ -31,7 +34,7 @@ do
                 echo "Checking tutorial $w for tags" 
                 if tester $w $topic;
                 then
-                    echo "ERROR: No 'tags' attribute found with its corresponding topic or workflow includes tool(s) from testtoolshed."
+                    echo "ERROR: No 'tags' attribute found with its corresponding topic or 'annotation' attribute with for example the title of the tutorial or workflow includes tool(s) from testtoolshed."
```
hexylena: you might consider:

1. in the tester script write to sys.stderr with the message that's more clear/detailed ("in wf X an annotation is missing, it should be Y")
2. Not writing an error in the bash portion and just setting the overall exit code

bedroesb: don't know 100% what you mean with  just setting the overall exit code, but I've added sys.stderr codes. Thanks for the suggestion.

bedroesb:
I included a mistake in one of the workflows on purpose to test the linting

bin/validate-workflow-tags.sh
hexylena: The changes you made look good :)

bedroesb:
I think this is good to go :+1:

hexylena:
Looks great! Thanks for the linting too

PR #1737, opened 2020-01-30
Update tutorial.md

erxleben:
Hi,
while preparing for a practical students course with that tutorial, I have seen some small issues which could be checked:

- The step "Hands-on: Classify the sequences into phylotypes"  was fast for me. Maybe to suggest to pick a coffee is not useful here, esp. if there is no coffee available in place ;) (line 414)
- The output of the step "Hands-on: Classify the sequences into phylotypes" looks different in the first line to me. This may be confusing for workshop participants if they do not get 100% the same
- line 462: input parameter is called "supply group or count table if you supplied OTU list"
- line 485: I get different results: pampa 2474, anguil 2202
- line 486: 922 instead 929
- line 648: the parameter shows only taxonomy or tabular. As output of Format MetaPhlAn2 output for Krona is tabular, tabular has to be chosen

Thanks!


topics/metagenomics/tutorials/general-tutorial/tutorial.md
```diff
@@ -82,7 +82,7 @@ has been sequenced using 454 GS FLX Titanium. For the tutorial, the original fas
 
 > ### {% icon hands_on %} Hands-on: Data upload
 >
-> 1. Import from [Zenodo](https://zenodo.org/record/815875) or from the data library (in "Analyses of metagenomics data") the files
+> 1. Import from [Zenodo](https://zenodo.org/record/815875) or from the data library (in GTN-Material -> "Analyses of metagenomics data") the files
```
lecorguille: `GTN-Material -> "Analyses of metagenomics data"`

Maybe it's just a detail but the path seems like that:
`GTN - Material -> Metagenomics -> Analyses of metagenomics data - The global picture`

bgruening: @erxleben ...?

topics/metagenomics/tutorials/general-tutorial/tutorial.md
```diff
@@ -82,7 +82,7 @@ has been sequenced using 454 GS FLX Titanium. For the tutorial, the original fas
 
 > ### {% icon hands_on %} Hands-on: Data upload
 >
-> 1. Import from [Zenodo](https://zenodo.org/record/815875) or from the data library (in "Analyses of metagenomics data") the files
+> 1. Import the following files from [Zenodo](https://zenodo.org/record/815875), or from the shared data library (in `GTN - Material -> Metagenomics -> Analyses of metagenomics data - The global picture`):
```
bebatut, line 85: we could even use Jekyll variables here (topic name, tutorial name) so that if we change them we do not need to update this line.
we could then add that to the planemo template
What do you think @shiltemann?

shiltemann: very good point :+1:, will update

shiltemann:
thanks @erxleben!

PR #1741, opened 2020-01-31
udate "Inspect AnnData" output

hrhotz:
unless I am doing something wrong, I get the ensembl identifiers and not the gene symbols at this step


hrhotz:
I guess, I have been wrong..... the output is correct and has to be like this in order to continue further down. More likely the problem is upstream: the “Variables index” should be set to gene_symbol in the 'Import AnnData' tool, shouldn't -it ?

PR #1743, opened 2020-02-04
import slides from Simon

hexylena:
@Slugger70 made some pretty ace slides, and they fit nicely with the ansible-galaxy training providing an overview for the week to come.


hexylena:
Cheers @nsoranzo

PR #1752, opened 2020-02-10
py3k/systemd the for connect-to-compute-cluster

hexylena:


topics/admin/tutorials/connect-to-compute-cluster/tutorial.md
```diff
@@ -297,20 +303,29 @@ At the top of the stack sits Galaxy. Galaxy must now be configured to use the cl
...
->
->    ```xml
->    <destinations default="slurm">
->        <destination id="slurm" runner="slurm"/>
->        <destination id="local" runner="local"/>
->    </destinations>
+> 3. Next, we need to configure the Slurm job runner. First, we instruct Galaxy's job handlers to load the Slurm job runner plugin, and set the Slurm job submission parameters. A job runner plugin definition must have the `id`, `type`, and `load` attributes. Then we add a basic destination with no parameters, Galaxy will do the equivalent of submitting a job as `sbatch /path/to/job_script.sh`. Note that we also need to set a default destination now that more than one destination is defined. In a `<destination>` tag, the `id` attribute is a unique identifier for that destination and the `runner` attribute must match the `id` of defined plugin:
```
nsoranzo:
```suggestion
> 3. Next, we need to configure the Slurm job runner. First, we instruct Galaxy's job handlers to load the Slurm job runner plugin, and set the Slurm job submission parameters. A job runner plugin definition must have the `id`, `type`, and `load` attributes. Then we add a basic destination with no parameters, Galaxy will do the equivalent of submitting a job as `sbatch /path/to/job_script.sh`. Note that we also need to set a default destination now that more than one destination is defined. In a `<destination>` tag, the `id` attribute is a unique identifier for that destination and the `runner` attribute must match the `id` of a defined plugin:
```

PR #1753, opened 2020-02-10
De novo transcriptome reconstruction - some fix

lecorguille:


topics/transcriptomics/tutorials/de-novo/tutorial.md
```diff
@@ -118,6 +118,9 @@ For quality control, we use similar tools as described in [NGS-QC tutorial]({{si
 >    > {: .solution }
 >    {: .question}
 > ![Before and after trimming comparison](../../images/BeforeAndAfterTrimming.png)
+>
+> 4. **Trimmomatic** {% icon tool %}: Run `Trimmomatic` on the remaining forward/reverse read pairs with the same parameters.
```
abretaud: Hmm why do you want to rerun trimmomatic?

lecorguille: Because we ask them to `Run HISAT2 on the remaining forward/reverse read pairs with the same parameters.` so I feel that we should also run Trimmomatic, isn't?

lecorguille:
What do you thing about introducing a little bit of Dataset Collection and [Group tags](https://galaxyproject.github.io/training-material/topics/galaxy-data-manipulation/tutorials/group-tags/tutorial.html) in this Tutorial ?

I don't want to denature the tutorial.

And and about Group tags, I don't know the policy about this kind of new features.
Do we consider the Group Tags (19.01) as enough mature? Because it will imply that the tutorial will no longer run on Galaxy < 19.01

bgruening:
Imho, group tags are mature enough and collections as well. Can we link users to the group tag tutorial and include this as a recommented dependency maybe?

topics/transcriptomics/tutorials/de-novo/tutorial.md
abretaud: Ah yes, got it

PR #1760, opened 2020-02-13
fixes to Clustering 3K PBMCs with Scanpy

hrhotz:
small improvements in the text in order to make it clearer which tool to use


shiltemann:
thanks @hrhotz !

PR #1770, opened 2020-02-14
Ephemeris tutorial improvements

nsoranzo:
- Add ephemeris installation instructions
- Explain how to get the API key of an admin
-  Test only one tool. Use `shed-tools test`
- Other small fixes.

### DIFF
--- a/topics/admin/tutorials/tool-management/tutorial.md
+++ b/topics/admin/tutorials/tool-management/tutorial.md
@@ -49,13 +49,13 @@ To run this tutorial, you will need to [install Ephemeris](https://ephemeris.rea
 
 > ### {% icon tip %} Installing Ephemeris in a Python virtual environment
 >
-> - Install `virtualenv` if it is not already available. On Ubuntu this can be done with `sudo apt install virtualenv`
-> - Create a virtual environment just for ephemeris, activate it and install ephemeris inside it:
->   ```console
->   virtualenv -p python3 ephemeris_venv
->   . ephemeris_venv/bin/activate
->   pip install ephemeris
->   ```
+> 1. Install `virtualenv` if it is not already available. On Ubuntu this can be done with `sudo apt install virtualenv`
+> 2. Create a virtual environment just for ephemeris, activate it and install ephemeris inside it:
+>    ```console
+>    virtualenv -p python3 ephemeris_venv
+>    . ephemeris_venv/bin/activate
+>    pip install ephemeris
+>    ```
 {: .tip}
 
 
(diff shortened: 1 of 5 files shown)


topics/admin/tutorials/tool-management/tutorial.md
```diff
@@ -36,15 +36,26 @@ This tutorial will introduce you to one of Galaxy's associated projects - [Ephem
...
 
-To run this tutorial, first you will need to [install Ephemeris](https://ephemeris.readthedocs.io/en/latest/installation.html). We recommend installing it in a virtualenv.
+To run this tutorial, you will need to [install Ephemeris](https://ephemeris.readthedocs.io/en/latest/installation.html). You would normally install it on your workstation, but during training courses we recommend to install it on the same virtual machine used for the Galaxy server.
+
+> ### {% icon tip %} Installing Ephemeris in a Python virtual environment
+>
+> - Install `virtualenv` if it is not already available. On Ubuntu this can be done with `sudo apt install virtualenv`
+> - Create a virtual environment just for ephemeris, activate it and install ephemeris inside it:
+>   ```console
+>   virtualenv -p python3 ephemeris_venv
```
hexylena: any chance this could be: 

```suggestion
>   python3 -m virtualenv -p python3 ephemeris_venv
```

nsoranzo: What's the advantage? It's longer and not how it's explained in most docs I've encountered.

hexylena:
The virtualenv command didn't work for me like that, on a couple of test
machines. it requested me to install the python-virtualenv package which
shouldn't be necessary.

fre. 14. feb. 2020, 20:42 skrev Nicola Soranzo <notifications@github.com>:

nsoranzo:
> The virtualenv command didn't work for me like that, on a couple of test machines. it requested me to install the python-virtualenv package which shouldn't be necessary.

In fact I've added instructions to also install the `virtualenv` package on the line above. Galaxy itself uses the `virtualenv` command to create its own virtualenv in `scripts/common_startup.sh`, and if that's not present on the system it downloads and runs it from a tmp directory.

Your suggested solution `python3 -m virtualenv -p python3 ` works on the VMs because the `python3-virtualenv` package is installed, but it's not clear why that's the case on the VMs given it's required only by `pipsi` and `tox`, neither of which seems required for the tutorials.

topics/admin/tutorials/tool-management/tutorial.md
```diff
@@ -125,50 +146,51 @@ There are additionally two ways to install tools, we'll show both:
 >
 {: .hands_on}
 
-This provides an easy way to do one-off installation of tools, but is less convenient if you want to install many tools. For that, you can install from a yaml file:
+This provides an easy way to do a one-off installation of a tool, but is not very convenient if you want to install many tools.
+For that, you can install from a YAML file:
 
 > ### {% icon hands_on %} Hands-on: Installing tools from a tool list
 >
-> 1. (optional) Use the `tail` command to watch the installation proceed
```
nsoranzo: @martenson I suppose you meant to watch the Galaxy logs here.

hexylena: looks like it

topics/admin/tutorials/tool-management/tutorial.md
```diff
@@ -197,15 +219,15 @@ Sometimes your users might request that you install all of the same tools as the
 >    {: .question}
 {: .hands_on}
 
-We will not install the tools from that server as the EU Galaxy server has more tools than most other known Galaxies, but it is useful as an example of how you can use Ephemeris to help you mirror another Galaxy instance to meet user needs.
+We will not install all the tools from the EU Galaxy server as that server has likely more tools than any other Galaxy instance, but it is useful as an example of how you can use Ephemeris to facilitate the mirroring of another Galaxy instance.
```
hexylena:
```suggestion
We will not install all the tools from the EU Galaxy server as that server likely has more tools than any other Galaxy instance, but it is useful as an example of how you can use Ephemeris to facilitate the mirroring of another Galaxy instance.
```

hexylena:
> Your suggested solution python3 -m virtualenv -p python3 works on the VMs because the python3-virtualenv package is installed, but it's not clear why that's the case on the VMs given it's required only by pipsi and tox, neither of which seems required for the tutorials.

The package was in the original bootstrapping, I just added the `3`, not knowing who or what might be using it. So the solution here looks like removing python{3,}-virtualenv, and just installing `virtualenv`, or?

PR #1776, opened 2020-02-24
Pulsar to py3k

hexylena:
:us::fireworks::us::fireworks:

### DIFF
--- a/topics/admin/tutorials/connect-to-compute-cluster/tutorial.md
+++ b/topics/admin/tutorials/connect-to-compute-cluster/tutorial.md
@@ -284,7 +284,7 @@ Moving one level further up the stack, we find DRMAA Python. This is a Galaxy fr
 
 # Galaxy and Slurm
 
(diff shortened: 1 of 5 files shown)


hexylena:
https://github.com/galaxyproject/ansible-pulsar/pull/6 needs to be merged / tagged 1.0.0 (breaking changes)

hexylena:
This almost works, pending the issues with requests currently being debugged.

Slugger70:
Are we not using the pip install so as to get the latest changes Marius was working on?

hexylena:
> Are we not using the pip install so as to get the latest changes Marius was working on?

no, per some discussion it was removed from the role, figuring that we could just use the proper release procedure, but I guess we'll need to use a workaround here.

natefoo:
We can pip install from git+https github, right?

hexylena:
Should be. If not i get to back out my changes in ansible-pulsra.

hexylena:
```
pulsar_package_name: git+https://github.com/galaxyproject/pulsar@commit#egg=pulsar-app
```

PR #1777, opened 2020-02-24
Make the creation of an ansible.cfg with `inventory` set non-optional

natefoo:
So that `-i hosts` does not need to be invoked in every Ansible tutorial.

I did leave `-i hosts` in a few tutorials because they either precede this training (ansible) or are more self-contained (monitoring, jenkins).


hexylena:
Awesome! So much less typing :raised_hands:

PR #1715, opened 2020-01-15
Workflow names + tags + linting

bedroesb:
When I installed all the workflows of GTN I saw that there was no consistency in the workflow naming and that tags helped a lot in organizing the huge list of workflows.

- improved tool install in docker scripts (retry if tool install fails)
- standardized naming in workflows
- tagging of workflows with their topic
- linting for workflows without topic tag
- linting for workflows using tools from the testtoolshed #1663 
- disabled `CTB Unittest workflow` workflow because it contained testtools -> linting failed as expected
- updated contributing tutorial to add the topic tag to the new workflow

PR #1780, opened 2020-02-25
[WIP] Improved workflow linting with gxformat2.

jmchilton:
gxformat2 has much more extensive validation of workflow structure then just checking json parses. Does some linting of best practices including the ones added here as of 0.11.0.


bgruening:
@bedroesb is this something you could look at? Thanks a lot @bedroesb and @jmchilton!

jmchilton:
Small subset of the Travis logs...

```
Checking ./topics/imaging/tutorials/hela-screen-analysis/workflows/analyze_screen.ga
.. WARNING: Workflow contained output without a label
.. WARNING: Missing expected training topic (imaging) as workflow tag.
.. WARNING: Missing workflow documentation (annotation or doc element)
.. WARNING: Workflow contained output without a label
Checking ./topics/imaging/tutorials/hela-screen-analysis/workflows/feature_extraction.ga
.. WARNING: Workflow contained output without a label
Checking ./topics/imaging/tutorials/imaging-introduction/workflows/Galaxy-Workflow-GTN_Tutorial__Introduction_to_Imaging.ga
.. WARNING: Workflow contained output without a label
Checking ./topics/introduction/tutorials/galaxy-intro-101-everyone/workflows/main_workflow.ga
.. WARNING: Workflow contained no outputs
```

It seems like it is catching the tag and annotation things as expected?

Lots of warning about workflows without outputs or outputs without labels. I mean - I hope @mvdbeek would agree these are real problems that should be linted. But... I guess y'all will not want to go through all the workflows and identifier and label outputs.

Do we need arguments to ignore these errors on an exclude list or something? I hate to add features for transient issues like this that will hopefully go down over time but that is a big list of workflows...

mvdbeek:
> I mean - I hope @mvdbeek would agree these are real problems that should be linted.

They 100% are, but I also think that training material workflows aren't production workflows. From that point of view I think it does make sense to be able to ignore these.

bedroesb:
Normally I added to every workflow an annotation and tag in  #1724 and #1715

bedroesb:
Ok nice work! I was wondering if it also checks whether there are tools used from the testtoolshed?

jmchilton:
> if it also checks whether there are tools used from the testtoolshed

It should - there is a test case for that (https://github.com/galaxyproject/gxformat2/blob/master/tests/test_lint.py#L396) but I haven't tested it in the wild.

bedroesb:
How can an output have no label when the workflow is downloaded from the Galaxy instance? Is it an extra step the user has to make in the workflow editor or how does this come? Would it be possible to automatically label the outputs and change the contribution tutorials to make people aware?

shiltemann:
closing this for now since our linting has changed a lot since this PR was opened 3,5 years ago, but created an issue to see if it might still be useful to use gxformat2 to augment our existing linting

PR #1784, opened 2020-02-26
Add workaround for CERTIFICATE_VERIFY_FAILED error

nsoranzo:


topics/admin/tutorials/tool-management/tutorial.md
```diff
@@ -146,6 +146,20 @@ There are two ways to install tools, depending on how you specify the tools to i
 >
 {: .hands_on}
 
+> ### {% icon tip %} Certificate issues
+>
+> If your Galaxy instance is served via the HTTPS protocol (as it should!), ephemeris will use the [requests](https://requests.readthedocs.io) Python library to encrypt the communication with Galaxy. Therefore, if your Galaxy uses a self-signed SSL certificate, `shed-tools` may fail with a `CERTIFICATE_VERIFY_FAILED` error.
```
hexylena:
```suggestion
> If your Galaxy instance is served via the HTTPS protocol (as it should be!), ephemeris will use the [requests](https://requests.readthedocs.io) Python library to encrypt the communication with Galaxy. Therefore, if your Galaxy uses a self-signed SSL certificate, `shed-tools` may fail with a `CERTIFICATE_VERIFY_FAILED` error.
```

natefoo:
:+1: 

Would it be worth mentioning that other than for the purposes of training, this is not something you should ever do on a real system?

nsoranzo:
> Would it be worth mentioning that other than for the purposes of training, this is not something you should ever do on a real system?

@natefoo It does say "if your Galaxy uses a self-signed SSL certificate...", but feel free to amend as you see fit!

natefoo:
True - so it should already be fairly obvious. Although in the case of training, they're not self-signed, they're just signed by an untrusted (and untrustworthy) CA.

hexylena:
Yeah, thought about that same distinction (self vs untrusted) and decided majority won't be bothered and merged as-it-was.

PR #1827, opened 2020-03-04
Switch to monitoring group where it should apply

selten:


selten:
I think we need to revert the last 2 lines that were changed, telegraf should run on the same machine as the galaxy servers.

hexylena:
Ok, can do.

nsoranzo:
> I think we need to revert the last 2 lines that were changed, telegraf should run on the same machine as the galaxy servers @hexylena.

Will be fixed by https://github.com/galaxyproject/training-material/pull/1828

PR #1858, opened 2020-03-19
Covid-19 analyses topic -> Virology

bedroesb:
This PR serves as a starting point for the Covid-19 analyses/ workflows found on https://covid19.galaxyproject.org/ .

I am planning to add everything that is on https://covid19.galaxyproject.org/ but for the text itself I will need the help of other people. I know that @lynnlangit  already proposed to help.

This topic will include 6 tutorials. @hexylena and @shiltemann how do you think we can work on this the best way? 

@bgruening This is only the backbone, will add more as soon as possible.

### DIFF
--- a/metadata/virology.yaml
+++ b/metadata/virology.yaml
@@ -1,0 +1,1 @@
+../topics/virology/metadata.yaml
--- a/topics/virology/README.md
+++ b/topics/virology/README.md
@@ -1,0 +1,5 @@
+
+Virology
+==========
+
+Please refer to the [CONTRIBUTING.md](../../CONTRIBUTING.md) before adding or updating any material
--- a/topics/virology/docker/Dockerfile
+++ b/topics/virology/docker/Dockerfile
@@ -1,0 +1,26 @@
+
+# Galaxy - Virology
+#
+# to build the docker image, go to root of training repo and
+#    docker build -t virology -f topics/virology/docker/Dockerfile .
+#
+# to run image:
+#    docker run -p "8080:80" -t virology
+#    use -d to automatically dowload the datalibraries in the container
+
+FROM bgruening/galaxy-stable:19.01
+
+MAINTAINER Galaxy Training Material
+
+ENV GALAXY_CONFIG_BRAND "GTN: Virology"
+
+# copy the tutorials directory for your topic
+ADD topics/virology/tutorials/ /tutorials/
+
+# install everything for tutorials
+ADD bin/docker-install-tutorials.sh /setup-tutorials.sh
+ADD bin/mergeyaml.py /mergeyaml.py
+ADD bin/data_libarary_download.sh /data_libarary_download.sh
+RUN /setup-tutorials.sh
+
+ENTRYPOINT ["/data_libarary_download.sh"]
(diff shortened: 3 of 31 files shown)


nsoranzo:
Could this be part of https://github.com/virtual-biohackathons/covid-19-bh20/wiki (and of https://github.com/virtual-biohackathons/covid-19-bh20/wiki/Workflows#galaxyprojectsars-cov-2 in particular) ?

bgruening:
ping @wm75

bgruening:
Thanks @bedroesb awesome start! Hopefully, we have this done for the hackathon @nsoranzo!

nsoranzo:
Cool, in case we will find something else to work on!

wm75:
Hmm, I like the idea, but as much as I can see the point of having these tutorials bundled under one topic, such a structure seems to collide with how the GTN material is organized in general.
If you look at the other topic, they are all analyses techniques, not biological research areas.

Following that structure the different parts of https://covid19.galaxyproject.org/ should rather live in different existing topics (like Assembly and Variant Analysis. One could put a covid-19 label on all individual tutorials to make them more findable.

I understand that this may seem less attractive at first, but I think not just duplicating the info available at https://covid19.galaxyproject.org/, but presenting the same concepts in restructured form could also be seen as a reason why you'd want to have this in the GTN in the first place.

Just my personal view. Maybe @shiltemann @hexylena @bebatut disagree with me?

hexylena:
I agree wholeheartedly with @wm75. I wrote my own response but his is more clear.

I understand that y'all are doing this because y'all want visibility; the ability to link to the gxy/training/virology and have everything in one place, but you've got that on numerous other places already and where you do want to link to the training materials you can re-use something like is on https://usegalaxy-eu.github.io/index-rna.html#training for the interim.

I would rather see a solution like tagging things covid19 and fixing the search (preferrably TeSS integration, otherwise LunrJS et al) to allow searching across topics properly. In the future when we do curricula (as part of gallantries), we can use that as well.

bedroesb:
This PR was made specifically to have this discussion. I heard that there was interest in `in depth` tutorials for the covid-19 analyses, but did not yet see any movement around it here on GTN. Happy to see now what the visions are.

hexylena:
@bedroesb great that you want to contribute these, super exciting. Thanks for being open to alternate structures :)

frederikcoppens:
What is best from the user point of view ? The visibility should serve that.

I agree fixing the search is the more scalable option, but this might not be easily done in the way you want / need. I'm not up-to-date with the trajectories in TeSS, not sure this is a viable option at this point (development effort will now go to other things in those teams, quite sure of that). 

A topic is indeed somewhat unconventional, which kind of fits the times, but might be easiest to make them findable as a set. Yes, we'll probably need to refactor over time to a more scalable approach.

shiltemann:
This PR: https://github.com/galaxyproject/training-material/pull/1943 should provide search functionality across topics, and supports sharing an overview of all covid19-tagged tutorials by url, e.g. `https://training.galaxyproject.org/search?query=covid19`

hexylena:
Looks like it is live @shiltemann https://training.galaxyproject.org/search?query=covid19 :)

shiltemann:
closing this since the search option should provide the desired functionality, thanks all for the discussion!

PR #1860, opened 2020-03-23
Climate panoply tutorial remove enable false for review

annefou:
The training is ready to be reviewed. Especially nice to get feedback from non climate specialists because panoply is a tool that could be used by anyone interested in visualizing netCDF data on maps.


bgruening:
@yvanlebras do you have time for this one?

PR #1876, opened 2020-04-28
ATAC-seq: add a question to check output of Cut step

shiltemann:
I added a question after the `Cut` step to have people check if their output is a valid BED file. There is a second Cut tool with almost identical title that produces the wrong output, this might help catch that in time.


lldelisle:
Indeed, many thanks

PR #1887, opened 2020-05-06
Add admin training compatibility notice

hexylena:
To address the mistaken belief in that

> This is focused for paid distros. Centos 7/8 builds do not work due to package requirements and availability

[from the tutorial feedback](https://github.com/galaxyproject/training-material/issues/989#issuecomment-624070315)

### DIFF
--- a/snippets/admin-testing.md
+++ b/snippets/admin-testing.md
@@ -1,0 +1,9 @@
+> ### {% icon tip %} Operating System Compatability
+>
+> These Ansible roles and training materials were last tested on Centos 7 and Ubuntu 18.04, but will probably work on other RHEL and Debian variants.
+>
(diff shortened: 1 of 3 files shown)


nsoranzo:
For CentOS 7 the "Galaxy Installation with Ansible" tutorial doesn't work out of the box because yum doesn't support Python3 at all and neither does dnf on CentOS 7 due to https://github.com/ansible/ansible/issues/67083 .

So I had to use:
```ini
[defaults]
interpreter_python = /usr/bin/python
```
in `ansible.cfg`, and install both `python-psycopg2` and `python3-psycopg2` in the playbook.

bgruening:
@nsoranzo do you want to put this in as a note in the text?

PR #1896, opened 2020-05-27
Update some statistics workflows for workflow testing

cat-bro:
- Add workflow_outputs for machine_learning.ga
- Add labels to workflow_outputs for classification_LSVC.ga and regression_GradientBoosting.ga
- Copy and update workflow tests for these 3 workflows from workflow-testing and add them here

The workflows load and run as they did before.  Any alterations have been for the purpose of testing the outputs with planemo.

@bgruening this follows from https://github.com/usegalaxy-eu/workflow-testing/pull/52

### DIFF
--- a/topics/statistics/tutorials/classification_regression/workflows/classification_LSVC-test.yml
+++ b/topics/statistics/tutorials/classification_regression/workflows/classification_LSVC-test.yml
@@ -1,0 +1,28 @@
+---
+- doc: Test Linear SVC classifier workflow
+  job:
+    breast-w_train:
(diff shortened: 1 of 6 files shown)


bgruening:
Thanks a bunch!

bgruening:
@cat-bro if you have any ideas where the files should be stored let us know. The workflow needs to be here and up-to-date. The testing files need to be on the workflow-testing repo, they are not needed here strictly. No idea at the moment how we should structure this and keep everything in sync.

PR #1898, opened 2020-05-28
update to fontawesome 5

shiltemann:


topics/contributing/tutorials/create-new-tutorial-content/tutorial.md
```diff
@@ -688,5 +688,82 @@ A bibliography will automatically be appended to the end of your tutorial (scrol
...
 {: .tip}
 
+
+# Icons
+
+We use the following icons:
+
+Icon ID                  | Icon
+---                      | ---
+comment                  | {% icon   comment %}
```
bebatut, line 698: We could maybe generate this table automatically with a for loop. It would avoid having to update it if we add, remove, change some icons in the config file. What do you think?

shiltemann: it was meant mostly as a temporary table for testing purposes that @hexylena  made for me (using a for loop indeed). ..we could keep a table in the tutorial if handy (though most icons are used just by the template, and not necessarily useful for tutorial creators)

hexylena:
It was my first idea as well, but I had trouble to use a Jekyll tag with a
variable. It just would say unknown tag, and would not use the variables
value. I've seen some things online but it did not look promising so I gave
up.

Also because if we kept it, then it needs more explanation for users,
rather than this random list but what each should be used for. And then we
have to update the config structure to handle icon + descriptions and I
decided it was too much for today :)

tor. 28. mai 2020, 20:40 skrev Saskia Hiltemann <notifications@github.com>:

topics/contributing/tutorials/create-new-tutorial-content/tutorial.md
bebatut: ah ok 😄

PR #1901, opened 2020-05-28
Add missing index

hexylena:


simonbray:
Thanks Helena!

hexylena:
Thanks for the review @simonbray

PR #1906, opened 2020-05-29
2020 overhaul of architecture slides.

jmchilton:
- Bring in plantuml sources for UML diagram SVGs and some small utils for building them. These things didn't migrate over when these slides became "the source for truth" instead of Galaxy's docs.
- Expand coverage of related projects.
  - Add slides for important newer projects - gxadmin, ephemeris, EU architecture, training-material.
  - Organize that section into user facing projects, plugin developer projects, and admin projects.
  - Include Björn's big ecosystem slide at the end of that section to try to place things in some context.
  - Add some logos, other small changes.
- Add big new section on project management for the code.
  - Cover pull requests, branches, security, code of conduct.
  - important things for people wanting to understand and contribute to the code.
- Add a new section describing important files and directories in the repository.
  - Use PlantUML mindmaps to describe these things visually in their hierarchies.
- Small tweaks to principles section - mention VueJS, ES6, uWSGI, GUnicorn.
- Substantial overhaul of walking through Galaxy communication and request handling.
  - In client-server diagram drop legacy content, updated client pieces for VueJS and Vuex.
  - Replace two uWSGI diagrams with a more full sequence diagram walking from TCP/IP request to Galaxy managers.
  - Add class diagram to compliment object diagram describing Galaxy web application framework pieces.
  - Add slide about why we don't use Django (a common question).
- Overhaul client architecture section.
  - Drop now incorrect information about less & gulp.
  - Rework client dependencies to reflect these are fetched npm instead of bower and no longer ship with Galaxy.
  - Try to update and clarify some points around webpack.
  - Update Makefile targets - fixing them and mentioning newer development targets (linting, formatting, testing, watching)
- Small tweaks to style, etc..

PR #1907, opened 2020-06-01
Add architecture slides on Python packages for Galaxy

jmchilton:
@nsoranzo started a review on #1906 so I spun this out. This PR extends that one with more information on Galaxy packages and Python files.


jmchilton:
Merged as part of a different PR.

PR #1914, opened 2020-06-12
Gitter sidecar

shiltemann:
@hexylena previously enabled embedding the Gitter sidecar to quickly access the admin Gitter channel from the admin tutorials. This allows the same to be specified on the tutorial level.

The sidecar will open the tutorial-level Gitter channel if defined, otherwise the topic-level Gitter channel, and defaults to the GTN lobby if neither are defined

Since @nomadscientist recently had the great idea to create a dedicated [single-cell Gitter channel](https://gitter.im/Galaxy-Training-Network/galaxy-single-cell), I have specified this channel for the various single-cell tutorials (let me know if I've missed any)

PR #1915, opened 2020-06-12
Add cheminformatics gitter

simonbray:
Depends on https://github.com/galaxyproject/training-material/pull/1914


shiltemann:
Awesome, thanks @simonbray! 

If this Gitter channel applies to every tutorial in the topic, you can also just set it in the topic-level metadata.yaml file, e.g. as was done [here for the admin topic](https://github.com/galaxyproject/training-material/blob/master/topics/admin/metadata.yaml#L22), and it will be applied to all tutorials automatially

simonbray:
Oh, that makes more sense, thanks @shiltemann.

PR #1900, opened 2020-05-28
Make translations more obvious

hexylena:
Fixes https://github.com/galaxyproject/training-material/issues/1878

image

Also switches to native names for languages as is more preferred. I don't super love this translation icon but that should be fixed with #1898 and the newer icon https://fontawesome.com/icons/language?style=solid

PR #1916, opened 2020-06-12
Fix "And more!" link

nsoranzo:
Follow-up on https://github.com/galaxyproject/training-material/pull/1900/files/053bc100d5acf0492fdd5d599ec7187b3bca5c72#r438757998


shiltemann:
thanks @nsoranzo!

PR #1926, opened 2020-06-18
Add download links for datasets

anuprulez:
The links for datasets should be downloadable and should not show them as raw files. That's why I think the datasets do not get automatically uploaded to the data library. The links in "Basics of machine learning" tutorial has downloadable links mentioned in the yaml file 

Thanks!

ping @bgruening

### DIFF
--- a/topics/statistics/tutorials/regression_machinelearning/data-library.yaml
+++ b/topics/statistics/tutorials/regression_machinelearning/data-library.yaml
@@ -6,23 +6,22 @@ destination:
   synopsis: Galaxy Training Network Material. See https://training.galaxyproject.org
 items:
 - name: Statistics and machine learning
-  description: Statistical Analyses for omics data and machine learning using Galaxy
-    tools
+  description: Statistical Analyses for omics data and machine learning using Galaxy tools.
   items:
   - name: Regression in Machine Learning
     items:
     - name: 'DOI: 10.5281/zenodo.2545213'
       description: latest
       items:
-      - url: https://zenodo.org/record/2545213/files/test_rows.csv?download=1
+      - url: https://zenodo.org/record/2545213/files/test_rows.csv
         src: url
         ext: csv
-        info: https://zenodo.org/record/2545213
-      - url: https://zenodo.org/record/2545213/files/test_rows_labels.csv?download=1
+        info: http://doi.org/10.5281/zenodo.2545213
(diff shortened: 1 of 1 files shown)


topics/statistics/tutorials/regression_machinelearning/data-library.yaml
```diff
@@ -11,18 +11,18 @@ items:
...
       description: latest
       items:
-      - url: https://zenodo.org/api/files/f1431fd7-a9dc-426d-ac78-3eaaa61a8d29/test_rows.csv
+      - url: https://zenodo.org/record/2545213/files/test_rows.csv?download=1
         src: url
         ext: csv
-        info: https://zenodo.org/record/2545213#.XEWTJ9-YVa0
-      - url: https://zenodo.org/api/files/f1431fd7-a9dc-426d-ac78-3eaaa61a8d29/test_rows_labels.csv
+        info: https://zenodo.org/record/2545213
+      - url: https://zenodo.org/record/2545213/files/test_rows_labels.csv?download=1
```
bgruening, line 21: the download=1 can also be removed I think

bgruening:
Cool, thanks are the others trainings also correct?

anuprulez:
I am not sure, maybe not .. We have this webinar trial run call now. Once it is over, I will check for other tutorials as well.

topics/statistics/tutorials/regression_machinelearning/data-library.yaml
anuprulez: if download=1 is removed, the file is shown just like that ... that's the error .. the link should automatically download the file

hexylena: We generally do not include it in other datasets. This supports both the usecases of :

- the user to viewing the dataset online without downloading a file and needing to open
- it to be loaded into galaxy.

There are other examples of datasets in libraries that lack the download=1, so, if it is failing, then it is likely due to some other reason.

bgruening:
@anuprulez can you please check the others. If we merge tm this will be deployed by monday.

anuprulez:
@bgruening  I have copied the format from https://github.com/galaxyproject/training-material/blob/master/topics/assembly/tutorials/debruijn-graph-assembly/data-library.yaml and updated the URLs. I cannot test if it works. I checked the links are working and the description is correct.

PR #1927, opened 2020-06-18
Update tutorial.md

erxleben:
some additional comments:
- Fig 1 has no description of the y axis
- line 84-86: explanation of dataset content would be useful after the data upload not before
- line 188: the tool "plot actuals vs predicted curves..." is found in the statistics section. Maybe give this as a hint (most of the tools are in the ML section)
- line 226: how?
- line 253: be careful! There is a different version of the tool to be found in "Test Tools" which has different parameters. They should use the tool in the ML section
- line 317: from where does this comes from?
- Key points: a "." is missing at the end of the second key point.


shiltemann:
thanks @erxleben! I will merge this, perhaps you would like to move the remaining comments you list to an issue so we don't lose track of them? thanks for the feedback!

khanteymoori:
@erxleben for the  "plot actuals vs predicted curves..."  hint, we have planned to delete all the duplicate tools.

PR #1935, opened 2020-06-23
Enchancements to Machine learning: regression tutorial

pvanheus:
* Fix reference to figures 7 and 8 (should be 8 and 9)
* Add hint to check that you are using Gradient Boosting Regressor, not Gradient Boosting Classifier
* Point out use of Estimator attributes tool to see attributes of estimator dataset.
* Highlight when to use `test_rows_labels` and when to use `test_rows_labels_without_header`.


anuprulez:
This looks good to me. Thanks @pvanheus

anuprulez:
If an estimator is used in combination with "Search CV" tool, then the hyperparameter optimisation table shows all the values of parameters of the estimator. Moreover, apart from the "Estimator attributes" tool, I am not sure about any other tool @khanteymoori

PR #1944, opened 2020-06-29
Contributor pages

hexylena:
Add contributor pages documenting their contributions to the training materials and providing more of a "training profile" page.

image

In the future we can consider expanding this to involve trainings they're offering or so.


shiltemann:
This is really cool, thanks @hexylena! 

I just added an option for people to add some bio text to the CONTRIBUTORS file, which will be shown at the top instead of the disclaimer, if present. Hopefully this will also be a place where those who have contributed to the framework can describe their contributions

PR #1950, opened 2020-07-06
Add GTN Stats Page

shiltemann:
@hexylena and I have started a page showing some general stats about this repo (available at `/stats`)

image

It uses [chart.js](https://www.chartjs.org/) for the graphs

Please let us know if you have suggestions of more stats or graphs to add here, we will expand this page over time :)

(Should hopefully make @tnabtaf's life a teensie bit easier for the monthly newsletter too? any requests from your side?)


tnabtaf:
Thank you

hrhotz:
I don't know, whether it is feasible, easy to implement, but what about some info wrt to
 - new tutorials have been added to those topics
 - updates have been made in these topics
 - etc.

hexylena:
It should be possible to detect "recently updated", which was on our radar. Completely new, is more difficult.

tnabtaf:
Some thoughts on additions: Are page visits counted at all?  

But let me stress, it is great as is, and I would approve it if I were a reviewer!

bgruening:
@tnabtaf no this is not easily possible. But the GalaxyProject is tracking GTN in google stats. So you have the page visits there.

bebatut:
I will open an issue with possible extension for the stat page #1954

PR #1959, opened 2020-07-10
Speed up jekyll serve

shiltemann:
The `jekyll-scholar` plugin was resulting in very long build times; this disables the plugin by default. It led to a ~3x speedup for me for the initial build,  and might also fix the problem with re-build times in the `--incremental` increasing over time (not sure tho)

This means citations/the bibliography will no longer be rendered when you do `make serve`, but `make serve-full` will enable you to still build the full site. A hint to this effect is shown in the tutorials in place of the real citations. `--incremental` is now also enabled on by default

we could also leave make serve as it is now, and add a make-quick target? There are also other things we could disable, e.g. all the contributor pages, but that doesn't take all that much time

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,10 +52,11 @@ install: clean create-env ## install dependencies
 .PHONY: install
 
 serve-full: ## run a local server (You can specify PORT=, HOST=, and FLAGS= to set the port, host or to pass additional flags)
+	@echo "Tip: to serve in incremental mode (faster rebuilds), use the command: make serve FLAGS=--incremental" && echo "" && \
 	$(ACTIVATE_ENV) && \
 		mv Gemfile Gemfile.backup || true && \
 		mv Gemfile.lock Gemfile.lock.backup || true && \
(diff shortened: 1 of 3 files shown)


nsoranzo:
I like the idea of call the new target `serve-quick` and keep the `serve` for the full thing.

hexylena:
```
$ make serve-quick ACTIVATE_ENV=pwd
This will build the website with citations and other content disabled, and incremental on by default. To run the full preview (slower), use make serve-full

/home/hxr/arbeit/galaxy/training-material
mv: cannot stat 'Gemfile': No such file or directory
mv: cannot stat 'Gemfile.lock': No such file or directory
Configuration file: /home/hxr/arbeit/galaxy/training-material/_config.yml
Configuration file: /home/hxr/arbeit/galaxy/training-material/_config-dev.yml
            Source: /home/hxr/arbeit/galaxy/training-material
       Destination: /home/hxr/arbeit/galaxy/training-material/_site/training-material
 Incremental build: enabled
      Generating...
       Jekyll Feed: Generating feed for posts
                    done in 12.972 seconds.
```

:heart_eyes: :heart_eyes: :heart_eyes:

PR #1961, opened 2020-07-10
Galaxy Architecture - Document Plugins

jmchilton:
Builds on #1907. Add more information on backend plugins.

- Re-contextualize models and managers as the top and bottom of the backend core stack, with plugins in the middle.
- Categorize different kinds of Galaxy plugins.
- Add information about datatype files.
- Add class diagrams for toolbox and tools, tool classes.
- Add class diagrams for workflow modules.
- Describe plugin_config.py and contextualize within an example (Galaxy's container resolvers)


topics/dev/images/Makefile
```diff
@@ -1,9 +1,11 @@
+MINDMAPS := $(wildcard *.mindmap.yml)
 INPUTS := $(wildcard *.plantuml.txt)
 OUTPUTS := $(INPUTS:.txt=.svg)
 
-all: plantuml.jar plantuml_options.txt $(OUTPUTS)
+all: plantuml.jar $(MINDMAPS) $(OUTPUTS)
 
 $(OUTPUTS): $(INPUTS)
```
hexylena, line 7:
```suggestion
$(OUTPUTS): $(INPUTS) $(MINDMAPS)
```

jmchilton:
I rebased and pushed another huge diff in there - that wasn't nice to do 🙈. Sorry.

I doubt these SVGs are very large - but they are generated artifacts and that makes me want to minimize the number of times they are commits they are regenerated inside of.

topics/dev/images/Makefile
jmchilton: Put this in the latest rebase, thanks for the suggestion!

PR #1963, opened 2020-07-12
Enhancements for Galaxy 101 for everyone

wm75:
This brings only one major fix: the output of datamash now retains the
header line, which makes it much more interpretable.
The rest are minor enhancements and typo fixes.


wm75:
@jennaj @Delphine-L I just went through the tutorial using the EU server and kept notes for the BCC training sessions:

1. The tutorial demonstrates use of the tools search bar with a screenshot that shows `remove beginning` typed in.

    With various configurations of the search on usegalaxy.* instances, I could reliably find the tool only when typing `beginning`.
    - `remove beginning` works on main and au, but not eu (has problems with multiple words in general)
    - `remove` works on au and eu, but not main
    Just something to keep in mind if users should work on different servers

2. In the "Extract Workflow" section the tutorial talks about clicking asterisks for toggling dataset visibility.
    With the latest WF editor UI on main, these asterisks are now checkboxes.
    On eu things still look as described (haven't checked au), but in the mid-term the text and the accompanying screenshot should be updated, and we should point this out to participants.

wm75:
@annefou Figures 1-3 lack an attribution

annefou:
@wm75 you are right! 

For the iris dataset, I think it is from https://www.datacamp.com/community/tutorials/machine-learning-in-r but the original pictures are available at:
     - https://nn.wikipedia.org/wiki/Fil:Iris_versicolor_3.jpg
     - https://commons.wikimedia.org/wiki/File:IMG_7911-Iris_virginica.jpg
- The diamon grading scale is from : https://4cs.gia.edu/en-us/blog/gia-diamond-grading-scales/​

Does it help?

PR #1978, opened 2020-07-16
Correct a few typos and add a tip in the ansible-galaxy tutorial

natefoo:


topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -101,7 +103,7 @@ The [static configuration setup](https://github.com/galaxyproject/ansible-galaxy
 3. Any templates are copied over
 4. The `galaxy.yml` (or `.ini`) is deployed
 
-The setup for deploying templates and configuration files is a little bit non-standard by Ansible standards. Here you are expected to provide your own templates and static config files, and then describe them as a list of files and where they should be deployed to.
+The setup for deploying extra Galaxy configuration templates and files is a little bit non-standard by Ansible standards. Here you are expected to provide your own templates and static config files, and then describe them as a list of files and where they should be deployed to.
```
nsoranzo, line 106:
```suggestion
The setup for deploying extra Galaxy configuration templates and files is a little bit non-standard by Ansible best practices. Here you are expected to provide your own templates and static config files, and then describe them as a list of files and where they should be deployed to.
```

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -117,8 +119,8 @@ galaxy_config_files:
     dest: "{{ galaxy_config.galaxy.datatypes_config_file }}"
   - src: files/galaxy/config/dependency_resolvers_conf.xml
     dest: "{{ galaxy_config.galaxy.dependency_resolvers_config_file }}"
-  - src: files/galaxy/config/disposable_email_blacklist.conf
-    dest: "{{ galaxy_config.galaxy.blacklist_file }}"
+  - src: files/galaxy/config/disposable_email_blocklist.conf
+    dest: "{{ galaxy_config.galaxy.blocklist_file }}"
```
hexylena, line 123: xref https://github.com/galaxyproject/galaxy/issues/9965

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -133,20 +135,23 @@ galaxy_config:
 ```
 {% endraw %}
 
-So the references in `galaxy_config_files` to `galaxy_config` are done to ensure that the setting for e.g. "location of the blacklist file" is the same between where we have configured Galaxy to looking for it, and where the file has been deployed, without requiring us to make variables changes in numerous places.
+So the references in `galaxy_config_files` to `galaxy_config` are done to ensure that the setting for e.g. "location of the datatypes config file" is the same between where we have configured Galaxy to looking for it, and where the file has been deployed, without requiring us to make variables changes in numerous places.
+
+> ### {% icon tip %} Define once, reference many times
+> Using practices like those shown above helps to avoid problems caused when paths are defined differently in multiple places. The datatypes config file will be copied to the same path as Galaxy is configured to find it in, because that path is only defined in one place. Everything else is a reference to the original definition!
```
hexylena, line 141:
```suggestion
> Using practices like those shown above helps to avoid problems caused when paths are defined differently in multiple places. The datatypes config file will be copied to the same path as Galaxy is configured to find it in, because that path is only defined in one place. Everything else is a reference to the original definition! If you ever need to update that definition, everything else will be updated accordingly.
```

PR #1989, opened 2020-07-22
New tool tag / GTN webhook

hexylena:
In support of https://github.com/galaxyproject/galaxy/pull/10024, we can now annotate training materials with the precise tool ID and optionally version.

We will add this to planemo so most people won't care about it, but for the older tutorials we can migrate them or encourage authors to, in order to take advantage of the webhook.

Please see https://github.com/galaxyproject/galaxy/pull/10024 for more information on what the tool tag will be used for.

I have converted a couple of training materials, but we should focus on doing the rest of the introductory ones before merging.

### DIFF
--- a/_plugins/jekyll-tool-tag.rb
+++ b/_plugins/jekyll-tool-tag.rb
@@ -7,20 +7,21 @@ module Jekyll
     end
 
     def render(context)
-      parts = @text.split('|')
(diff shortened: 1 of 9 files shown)


hexylena:
We've taken the opportunity to indicate if just a tool ID is specified:

image

Or if a tool version is specified as well:

image

shiltemann:
awesome!

PR #1990, opened 2020-07-22
Add .dockstore.yml

coverbeck:
The .dockstore.yml works in conjunction with GitHub apps to notify [Dockstore](https://dockstore.org) when a Galaxy workflow has been updated in GitHub.

This .dockstore.yml adds the workflows listed in dockstore/dockstore#3186, suggested by @MoHeydarian .

I have also discussed this with @jmchilton . If the PR gets merged, he will do the work as a Dockstore user to publish these workflows on Dockstore.

Thanks!

### DIFF
--- a/.dockstore.yml
+++ b/.dockstore.yml
@@ -1,0 +1,17 @@
+version: 1.2
+workflows:
+  - name: intro-short
+    subclass: GXFORMAT2
+    primaryDescriptorPath: topics/introduction/tutorials/galaxy-intro-short/workflows/Galaxy-Workflow-galaxy-intro-short.ga
+  - name: intro-everyone
+    subclass: GXFORMAT2
+    primaryDescriptorPath: topics/introduction/tutorials/galaxy-intro-101-everyone/workflows/main_workflow.ga
+  - name: de-novo
+    subclass: GXFORMAT2
+    primaryDescriptorPath: topics/transcriptomics/tutorials/full-de-novo/workflows/main_workflow.ga
+  - name: dip
+    subclass: GXFORMAT2
+    primaryDescriptorPath: topics/variant-analysis/tutorials/dip/workflows/diploid.ga
+  - name: genome-annotation
+    subclass: GXFORMAT2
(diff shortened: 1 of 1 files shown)


hexylena:
Hi all. We're happy to merge this if it helps. 

For the future, should this list be auto-generated at some point in the future from all the workflows? Is this just an initial test with a few hand-selected ones?

jmchilton:
@hexylena Thanks so much for being open to this, awesome!

I'm working Planemo commands for initializing this file automatically by scanning for Galaxy workflows and test files - https://github.com/galaxyproject/planemo/pull/1028. I hadn't considered auto-updating over time - I guess I should put some thought into that.

hexylena:
Oh no worries, we can make it part of the GTN linting - ensure if there's a workflow that the index matches and otherwise ask them to run the command. But it being in planemo is the important bit for me :)

.dockstore.yml
```diff
@@ -0,0 +1,17 @@
+version: 1.2
+workflows:
+  - name: intro-short
+    subclass: GXFORMAT2
```
jmchilton, line 4: @coverbeck Can we make this Galaxy also as we're fixing that descriptor bug? GXFORMAT2 isn't appropriate for these workflows in this file either.

coverbeck: @jmchilton Can we discuss at our CoFest meeting this afternoon? We felt this one made more sense than the TRS case, but sounds like not.

jmchilton: Sounds good we can discuss it then, but these files are literally the older JSON files and not the newer format 2 YAML files. We can talk about it then.

coverbeck: That sounds pretty compelling.

denis-yuen: I think at CoFest, I discussed the reverse actually. GALAXY can cover both the newer format and the older format. TRS doesn't distinguish between, say WDL draft-2 and v1 or CWL 1.0 and CWL v1.1. It is just WDL and CWL. 
TRS looks like it matches this expectation currently. 

We use GXFORMAT2 in a bunch of places internally which was my mistake. But we might be able to change the external representation for now without much pain?

coverbeck: dockstore/dockstore#3686

PR #1992, opened 2020-07-23
test for statistics clustering workflow

cat-bro:
Workflow outputs for the workflow and a test.   Passes on .au and .eu


shiltemann:
awesome!

PR #2003, opened 2020-07-24
Quality control: update tutorial to include FASTQE

mblue9:
Update Jul 2021: Have tried to incorporate FASTQE into current Quality Control tutorial adding in one new dataset. Am done and happy to revert anything or make other changes.

Jul 2020: Here's a start for a FASTQE tutorial, adapting the current Galaxy Quality Control tutorial.
- Definitely needs more emojis!
- Waiting on this tool update to add in screenshot of output https://github.com/galaxyproject/tools-iuc/pull/3141
- How do I get this tut to show up in the Sequence Analysis topic list locally eg local equivalent of here https://galaxyproject.github.io/training-material/topics/sequence-analysis/ (sorry I forget)
- Can add in the MultiQC like original tut later if FASTQE gets that working
- Can add in GC later if FASTQE gets that working


hexylena:
Hi @mblue9 

> How do I get this tut to show up in the Sequence Analysis topic list locally eg local equivalent of here galaxyproject.github.io/training-material/topics/sequence-analysis (sorry I forget)

It should do so automatically.

Given the large overlap with the other QC tutorial, might it make sense to integrate FASTQE directly into that?

mblue9:
>It should do so automatically. (If you're serving with --incremental you might need to touch the topic index, touch topics/sequence-analysis/index.md)

Ok I'll try that, thanks.

>Given the large overlap with the other QC tutorial, might it make sense to integrate FASTQE directly into that?

It could do if people were ok with that. Although I was thinking of changing the dataset to better show the effect of fastqe & then would need to see if that dataset makes sense for the rest of the points shown in the other QC tut. Also there isn't a paired-end version of the dataset I was thinking of swapping in afaik (the snake one in the history I shared) so might end up needing to include 2 different datasets if want to show the qc of paired data. But I could try it & see. Or any comments welcome. Will try to work on this next week.

hexylena:
I'd be ok with it! It's a much more visual explanation than just the ascii (I always see students eyes glazing over) and if it isn't in the default one, I worry that people don't realise there's this awesome new tool that makes the topic potentially easier for students?

I think we have discussed previously replacing the dataset because the quality is too good already? and it would be more interesting to show one that has more rubbish. So this could be a good time to do that too?

shiltemann:
Hi @mblue9 I agree with @hexylena, it could be nice to just add it to the QC tutorial, we can always indicate it is possible to skip that section if people are short on time orso. 

Indeed, I have been thinking of replacing the dataset anyway, just to have a more pronounced before/after effect of the trimming step, so you wanted a new one for this anyway, two birds with one stone maybe? :+1: 

thanks for adding this!

mblue9:
Cool thanks for the input @hexylena @shiltemann! I've started trying to incorporate FASTQE into the existing QC tut, still need to update some screenshots & check it makes sense.

topics/sequence-analysis/tutorials/quality-control/tutorial.md
```diff
@@ -131,29 +127,253 @@ Phred Quality Score | Probability of incorrect base call | Base call accuracy
...
+>
+> What is the lowest mean score in this dataset? 
+>
+> > ### {% icon solution %} Solution
+> > The lowest score in this dataset is 😿 13.
+> {: .solution }
+{: .question}
+
+
+## Improve quality with Cutadapt
```
shiltemann, line 197: I love that you added a dataset that shows a clearer improvement before and after QC! 

Would it maybe make sense to show FastQC before Cutadapt as well? and then after cutadapt participants can observe the before/after improvements with both tools?

mblue9: Thanks @shiltemann ! I wasn't expecting a review during GCC I just wanted to finish this off (finally). 

>Would it maybe make sense to show FastQC before Cutadapt as well? and then after cutadapt participants can observe the before/after improvements with both tools?

It could do. I was afraid to make it too long but I'll see how I could fit it in, maybe people can just skip bits if they don't want to do/see both.

shiltemann: yeah fair point, maybe we can just amend the FastQC section to also do the before & after, and give users a choice of tool orso, not sure. But yes, things that can wait until after GCC (or maybe a CoFest activity?). Thanks for finishing it!

mblue9: Ok I've added it as a CoFest activity :)
Just noting here, I was considering adding in fastp while I was updating this tut as it's used in other tutorials. Thought it might make sense to show it in this QC tutorial as an optional step. But unlike FastQC it didn't detect the adapter, not sure why so I've that out for the moment.

shiltemann (approved):
Thanks @mblue9! I like this update, it is still easy for student/teachers to show only FastQC, or only FastQE, and adds a bit of fun to the tutorial. I also really like the fact that we can see improvement in the quality reports after trimming with this additional dataset.

I look forward to teaching this in my course on Friday :)

mblue9:
Thanks @shiltemann! ☺️ 
Let me know how you get on.

shiltemann (approved):

PR #2008, opened 2020-07-26
101 tutorial: add note about error

mblue9:
They may get "Invalid column choice" error & if so they need to specify the column.


hexylena:
Is there a specific format that's required? "3" or "c3" or so?

mblue9:
Yes I've added a line for that now 
``` 
If you have to type the column number, you need to type just the number e.g. `4` (not `Column 4` or anything else).
```

hexylena:
Cool, thanks @mblue9

PR #2007, opened 2020-07-25
Add workflow test to M. tuberculosis variant analysis tutorial

pvanheus:
This is a continuation of #2004.

PR #2010, opened 2020-07-28
Fix broken boxes & make checking mandatory

hexylena:
Much of what was caught wasn't a 'true' breakage, just a completely unnecessary use of blockquotes to wrap images or other text.

But a few real problems were found!

Will conflict with #2007


shiltemann:
thanks @hexylena! And thanks for the review @melpetera :+1:

PR #2021, opened 2020-08-03
update maxquant workflow to use histogram version to 1.0.4, add test

cat-bro:
Replace final step dna_filtering/histogram with histogram/histogram.  dna_filtering and histogram versions 1.0.3 are identical tools.  histogram version 1.0.4 has been updated to use python3.


topics/proteomics/tutorials/maxquant-label-free/workflows/maxquant_label_free_tests.yml
```diff
@@ -0,0 +1,11 @@
+- doc: Test for maxquant_label_free.ga
+  job: maxquant_label_free_job1.yml
```
bgruening, line 2: @cat-bro what do you think about having in general everything in one file for those simple tests?

cat-bro: @bgruening that probably makes more sense.  I've altered this and one from last week to be single files with `-test.yml` naming convention (though planemo will accept a few variations)

bgruening: If you think its worth to enforce this on the planemo site, let's do this. We need to flush out what we recommend and what we enforce I think and write more documentation.

This might also be interesting for you: https://github.com/galaxyproject/planemo/pull/1029
Many more ways to specify complex inputs :)

PR #2028, opened 2020-08-12
Fix to inputs 10x pre-processing tutorial

mtekman:
The new STARsolo does not use repeats, since the new version (https://github.com/galaxyproject/tools-iuc/pull/3151)

This training makes use of the `snippets/build_paired_list_collection.md` which is conditional on  https://github.com/galaxyproject/training-material/pull/2027


bgruening:
@mtekman does the workflow needs an update?

mtekman:
@bgruening Good point, yes it does -- on my TODO

mtekman:
(force pushed to incorporate merge from #2027)

Issue #1459, opened 2019-07-04
Tutorials missing workflows


shiltemann:
This is a list of tutorials that are missing a workflow, if anybody would like to run this tutorial and extract the workflow from their history afterwards and upload it here, this would be a great help for us!

- [ ] **topic:** Introduction, **tutorial:** galaxy-intro-101 [link](https://training.galaxyproject.org/training-material/topics/introduction/tutorials/galaxy-intro-101/tutorial.html)
- [ ] **topic:** Introduction, **tutorial:** galaxy-intro-peaks2genes [link](https://training.galaxyproject.org/training-material/topics/introduction/tutorials/galaxy-intro-peaks2genes/tutorial.html)
- [ ] **topic:** Genome Annotation, **tutorial:** genome-annotation [link](https://training.galaxyproject.org/training-material/topics/genome-annotation/tutorials/genome-annotation/tutorial.html)
- [ ] **topic:** Ecology, **tutorial:** regionalGAM [link](https://training.galaxyproject.org/training-material/topics/ecology/tutorials/regionalGAM/tutorial.html)
- [ ] **topic:** Ecology, **tutorial:** species-distribution-modeling [link](https://training.galaxyproject.org/training-material/topics/ecology/tutorials/species-distribution-modeling/tutorial.html)
- [ ] **topic:** Statistics and machine learning  **tutorial:** machinelearning  [link](https://training.galaxyproject.org/training-material/topics/statistics/tutorials/machinelearning/tutorial.html) 
- [ ] **topic:** Transcriptomics, **tutorial:** rna-seq-viz-with-volcanoplot [link](https://training.galaxyproject.org/training-material/topics/transcriptomics/tutorials/rna-seq-viz-with-volcanoplot/tutorial.html)
- [ ] **topic:** Variant Analysis, **tutorial** non-dip [link](https://training.galaxyproject.org/training-material/topics/variant-analysis/tutorials/non-dip/tutorial.html)
- [ ] **topic:** Variant Analysis, **tutorial** somatic-variants [link](https://training.galaxyproject.org/training-material/topics/variant-analysis/tutorials/somatic-variants/tutorial.html)

And if you run into any other issues or questions when doing the tutorial, please add a fix or file an issue for it :)

chrisbarnettster:
Trying out the Statistics and machine learning  **tutorial**.

erasche:
Updated with fresh data, ca 75% of tutorials have workflows so that's great! :)

[...]

PR #2047, opened 2020-10-05
TAL1 tutorial updates

malloryfreeberg:
# Summary of changes

## General
- Added a Galaxy workflow (#1459).
- Added workflow test yaml, with inputs and outputs.
- Added subdirectory for images for this tutorial.
- Fixed typos.
- Changed formatting to make it consistent within the tutorial.
- Updated text to improve clarify.
- Changed formatting of mouse protein and gene names to be consistent with international standards.
- Updated images where Galaxy interface was sufficiently changed.
- Removed "enable: true".

### Step 2
- Removed second question since "Kmer Content" heading no longer appears in the tool output.

### Step 6
- Added parameters to pool control and treatment samples and updated format of input files to "single-end BAM", since these are replicate *single-end* samples. This will affect downstream steps as pooling the replicates (the approach in this PR) is different than considering the replicates paired-end reads (the existing approach).

### Step 7
- Removed the cut columns step. The existing tool doesn't cut the columns correctly, and the resulting files isn't recognised by Trackster or other tools. The narrow peaks files produced from MACS2 callpeaks suffice for subsequent steps that take peak information.
- Updated much of the Trackster instructions and images as there have been many changes to Trackster.
- Removed an unanswered question.

### Step 10
- Added new image since heatmaps appear different after changing Step 6.


topics/epigenetics/tutorials/tal1-binding-site-identification/tutorial.md
```diff
@@ -6,22 +6,22 @@ redirect_from:
 title: "Identification of the binding sites of the T-cell acute lymphocytic leukemia protein 1 (TAL1)"
 zenodo_link: "https://doi.org/10.5281/zenodo.197100"
 edam_ontology: "topic_3169"
-enable: false
+enable: true
```
hexylena, line 9:
```suggestion
```

is seems there's a small bug in our linting for whenever this value is 'true', so, we can just remove this attribute and it will have the same effect.

hexylena:
Hi @malloryfreeberg! This looks great at first view. Maybe one of our @galaxyproject/training-epigenetics maintainers has time to review this?

On the technical/gtn side:

1. Would you mind adding an index.md to the workflows directory? Check one of the other workflows directories for an example. This should fix one of the linting errors.
2. Additionally if you have a Galaxy history on hand where you ran this, would you be willing to write a test case for this workflow? Then we can make sure it stays functional. There is some documentation on how to do this here: https://training.galaxyproject.org/training-material/topics/contributing/tutorials/create-new-tutorial-technical/tutorial.html#testing-the-workflow-recommended

topics/epigenetics/tutorials/tal1-binding-site-identification/tutorial.md
malloryfreeberg: I was trying to read through the Travis logs and couldn't figure out what the bug was. Thanks for finding it!

malloryfreeberg:
Hi @hexylena - 
1. I have added the `index.md` file, and the linting errors are now fixed.
1. I have added a `-test.yml` file which includes all inputs and tests for 3 output files. I confirmed the test yml works by running `planemo test`, although the test takes about 30m to complete on https://usegalaxy.org/ 

Let me know if there is anything else I can do!

hexylena (approved):
This looks great, thanks @malloryfreeberg!! 

I've pushed some additional changes:

- Use of bibliography file for references
- Used the tool references, so GTN-in-Galaxy users can take advantage of https://github.com/galaxyproject/galaxy/pull/10024
- Replaced a couple blocks with snippets.

But overall this looks great.

PR #2054, opened 2020-10-10
Somatic variants tutorial updates

malloryfreeberg:
# Summary of changes

This was a really fun tutorial, and really nice to follow. Kudos to @wm75 for a very high-quality tutorial!

## Changes

- Added a Galaxy workflow (#1459).
- Added workflow test yaml.
- Fixed a typo and indentation issue.
- Added tool references, so GTN-in-Galaxy users can take advantage of galaxyproject/galaxy#10024.

## Notes

- In the `VarScan somatic` step, the instructions say to use the outputs of CalMD as the input files; however, there was an additional filtering step done after the CalMD step. So, the text has been updated to clarify that the outputs of the *filtered* CalMD outputs should be used here.
- In the workflow yaml, I originally marked the dbsnp input file as `filetype: vcf` as indicated by the tutorial. However, running the planemo test resulted in the file marked as failed with the error message `"Failed to run job final job state is [error]."` in the `tool_test_output.html` output file. I therefore set the dbsnp file to `filetype: vcf_bgzip` in the yaml, thinking maybe there was an issue with the file being so large (>400MB). This resulted in successful import of the file during the planemo test, although I'm not sure if it will have any effect on the integrity of the final results.

Workflow tested by running:
```
planemo test --galaxy_url "https://usegalaxy.eu/" --galaxy_user_key "$GALAXY_USER_KEY" \
 --no_shed_install --engine external_galaxy somatic_variants_tutorial_workflow.ga
```


wm75 (approved):
@malloryfreeberg A great contribution! Thanks for the careful fixes and the test workflow; all your changes and additions make perfect sense!

The explanation for the vcf <-> vcf_bgzip issue is this: the dbSNP file is large so storing and downloading it in compressed form makes a lot of sense. However, the version of gemini_annotate that was the latest when I wrote the tutorial couldn't use vcf_bgzip, but only vcf. Declaring the format as vcf during data import makes Galaxy decompress the file on import so it becomes usable with the tool. Since then I've updated gemini_annotate to work also with vcf_bgzip so the file format does not matter anymore (https://github.com/galaxyproject/tools-iuc/pull/2819).
Apparently, planemo is stricter than Galaxy itself and doesn't allow the implicit conversion, but with the current tool version this won't impact the results.

Issue #2078, opened 2020-10-20
Fix (or remove) broken links, minor typos in tutorials


ic4f:
- [x] [Rule Based Uploader](https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/upload-rules/tutorial.html):
	- Hands-on: Downloading Data from ENA: downloaded directly **here**
- [x] [R basics in Galaxy](https://training.galaxyproject.org/training-material/topics/introduction/tutorials/r-basics/tutorial.html): 
	- Learning more about lists: **nice tutorial**
- [x] [NGS data logistics](https://training.galaxyproject.org/training-material/topics/introduction/tutorials/galaxy-intro-ngs-data-managment/tutorial.html):
	- [x] From “Understanding and improving high-throughput sequencing data production and analysis”, Ph.D. dissertation by **Martin Kircher**
	- [x] From Ph.D. dissertation of **Martin Kircher**.
	- [x] What are base qualities? - last paragraph: missing closing paren (after Wikipedia)

hexylena:
The first one is a 500 internal server error on their end. But we should definitely move this dataset to zenodo!

ic4f:
@hexylena I notified them (ENA) using their contact form. 
(I don't have all the dataset info to fill out the Zenodo form though.)

hexylena:
@ic4f it doesn't take that much information for a zenodo entry, for what it's worth. e.g. https://zenodo.org/record/3603221

- GTN Tutorial: Name
- Short description like "data for this tuto"
- The files
- CC4.0
- there's an option for associating it with a collection, just find the galaxy/gtn one
- I don't think there's anything more?

ic4f:
@hexylena Seems simple enough, will do!

hexylena:
Just to check: were you claiming this issue as in you want to work on it? Or just reporting the issue for someone else to?

ic4f:
Not claiming the issue, just adding the dataset to zenodo to check it off.

ic4f:
@hexylena apparently those datasets are on zenodo (although with overwritten sequence identifiers), and those URLs are in the tutorial - right below the paragraph with the broken link. More importantly, the link is fixed ^.

PR #2094, opened 2020-10-23
Fix ENA broken link in tutorial.

ic4f:
Fix part of #2078.
As simple as that.


hexylena (approved):
love it! super simple :)

PR #2095, opened 2020-10-23
Medium level assembly slides

abretaud:
I had a session this week where I presented these new slides for assembly.
It's mostly a remix of [de bruijn](https://training.galaxyproject.org/training-material/topics/assembly/tutorials/debruijn-graph-assembly/slides.html) and [unicycler](https://training.galaxyproject.org/training-material/topics/assembly/tutorials/unicycler-assembly/slides.html) slides, with a few more slides here and there.
The goal was to have medium level slides on debruijn/OLC graphs, after the [general intro](https://training.galaxyproject.org/training-material/topics/assembly/tutorials/general-introduction/slides.html), without digging too much into the details. I think it worked better for students than last years where I used the [de bruijn](https://training.galaxyproject.org/training-material/topics/assembly/tutorials/debruijn-graph-assembly/slides.html) slides.
Any comments welcome of course!

By the way is there any plan to be able to sort things in the topic page? Ideally, the third line should be first on this page to reflect a kind of learning path:
image


abretaud:
(hum, to be clear: the existing slides are fine, it's just that with the public I had + the available time, I needed something a bit simpler)

shiltemann:
thanks @abretaud! Yeah, maybe we should support having more than one topic intro slide deck (for different levels, audiences, etc indeed) and displaying those all at the top. I will look into it

topics/assembly/slides/introduction.html
```diff
@@ -0,0 +1,376 @@
...
+```
+
+...then they may overlap within the genome.
+
+---
+
+## Directed graphs
+
+We build a *directed graph* where directed *edges* connect *nodes* representing overlapping reads:
+![](../images/dag.png)
```
shiltemann, line 71: could you add alt text to the images?

shiltemann (commented):
looks good @abretaud!

Would you like to add a `level:` metadata tag to indicate these are medium (intermediate) level slides?

Speaker notes would also be nice (And if you want them to be suitable for automatic video creation, check out the instructions [here](https://training.galaxyproject.org/training-material/topics/contributing/tutorials/slides-with-video/tutorial.html))

shiltemann:
btw, the topic-level slides are always shown at the top, since it is assumed those will provide a general introduction into to the topic. 

So if the slides from the third line are more of a general/first intro, would it maybe make sense to use these as the topic intro slide deck, and put these new medium-level slides in a tutorial folder (even if it doesn't have any hands-on) as say "A closer look at Genome Assembly Algorithms"? You'd get the ordering you want I think at least'? Just an idea..

abretaud:
Sorry for the delay :D
So I tweaked the titles to get a better ordering, what do you think?
image

hexylena (approved):

PR #2105, opened 2020-10-29
Move UCSC steps into tips, and use data from zenodo

hexylena:
I love the galaxy 101 tutorial, much prefer it to some of the others, but the UCSC steps cause *so* much pain with users. We tell them about how uniform all galaxy tools are... and then show them the one type of tool that doesn't behave like 99% of the tools they will use.

This simplifies that and should make the tutorial much easier.


hexylena:
Thanks so much @malloryfreeberg for depositing the data in zenodo already, that made this change super easy :smile:

shiltemann (approved):

PR #2118, opened 2020-11-06
Data library update

hexylena:
I'm not sure what to do about this. We stopped generating these for some reason? Which I now forget.

But it means that the shared-data library doesn't have any of these datasets which is really unfortunate. We were doing a tutorial today and I noticed that some of the tutos were missing.

I've added a script here to update every data library, maybe we can run it on cron or so?

Also there are numerous issues :( 

- [ ] planemo can't handle it when there's a `---` in the tutorial, which is perfectly valid markdown. It needs to do proper yaml streams and only process the first.
  - any one willing to volunteer for this? it needs some refactor (split document with `---` and pick first section, rather than the current regex which breaks.)
- [ ] the share/datatypes doesn't seem to work, judging by the number of 'fastq.gz' that became 'gz' instead of 'fastqsanger.gz' per the file.
- [ ] topics/computational-chemistry/tutorials/htmd-analysis/data-library.yaml has a zip file in their zenodo which means it's useless for us. (Their bib file also has some serious issues we should've caught.)
- [ ] many datatypes are missing

### DIFF
--- a/topics/ecology/tutorials/genetic-map-rad-seq/data-library.yaml
+++ b/topics/ecology/tutorials/genetic-map-rad-seq/data-library.yaml
@@ -15,89 +15,89 @@ items:
       items:
       - url: https://zenodo.org/api/files/af88bbbe-43a7-4c6b-8e12-3477cd3106ad/female
         src: url
-        ext: fasta
+        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
         info: https://doi.org/10.5281/zenodo.1219888
       - url: https://zenodo.org/api/files/af88bbbe-43a7-4c6b-8e12-3477cd3106ad/male
         src: url
-        ext: fasta
+        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
         info: https://doi.org/10.5281/zenodo.1219888
       - url: https://zenodo.org/api/files/af88bbbe-43a7-4c6b-8e12-3477cd3106ad/progeny_1
         src: url
-        ext: fasta
+        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
         info: https://doi.org/10.5281/zenodo.1219888
       - url: https://zenodo.org/api/files/af88bbbe-43a7-4c6b-8e12-3477cd3106ad/progeny_10
         src: url
-        ext: fasta
+        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
         info: https://doi.org/10.5281/zenodo.1219888
       - url: https://zenodo.org/api/files/af88bbbe-43a7-4c6b-8e12-3477cd3106ad/progeny_11
         src: url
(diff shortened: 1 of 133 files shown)


topics/ecology/tutorials/genetic-map-rad-seq/data-library.yaml
```diff
@@ -15,89 +15,89 @@ items:
       items:
       - url: https://zenodo.org/api/files/af88bbbe-43a7-4c6b-8e12-3477cd3106ad/female
         src: url
-        ext: fasta
+        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
```
bgruening, line 18: that looks wrong

topics/ecology/tutorials/regionalGAM/data-library.yaml
```diff
@@ -10,9 +10,24 @@ items:
   items:
   - name: Regional GAM
     items:
-    - name: 'DOI: 10.5281/zenodo.1324204'
+    - name: 'DOI: 10.5281/zenodo.1324204#.W2BmRn7fNE4'
```
bgruening, line 13: this should not be needed imho

topics/epigenetics/tutorials/atac-seq/data-library.yaml
```diff
@@ -17,13 +17,13 @@ items:
       items:
       - url: https://zenodo.org/api/files/b205aa3d-919b-4fbb-a8dd-3ae01d5daa91/ENCFF933NTR.bed.gz
         src: url
-        ext: bed
+        ext: gz
         info: https://zenodo.org/record/3862793
       - url: https://zenodo.org/api/files/b205aa3d-919b-4fbb-a8dd-3ae01d5daa91/SRR891268_chr22_enriched_R1.fastq.gz
         src: url
-        ext: fastqsanger.gz
```
bgruening, line 24: fastq.gz should be problably be always fastqsanger.gz

topics/visualisation/tutorials/jbrowse/data-library.yaml
```diff
@@ -19,15 +19,15 @@ items:
...
+        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
         info: https://doi.org/10.5281/zenodo.3591856
       - url: https://zenodo.org/api/files/0f156ca2-fb89-4d6f-8b80-02e51e5f5339/dna%20sequencing.bam
         src: url
         ext: bam
         info: https://doi.org/10.5281/zenodo.3591856
       - url: https://zenodo.org/api/files/0f156ca2-fb89-4d6f-8b80-02e51e5f5339/dna%20sequencing%20coverage.bw
         src: url
-        ext: bigwig
+        ext: '# Please add a Galaxy datatype or update the shared/datatypes.yaml file'
```
bgruening, line 30: that looks odd

PR #2153, opened 2020-11-25
Replace Travis with GH Actions!

hexylena:
I'm moving the build/deploy over to github since travis is broken and not sure if galaxyproject buys any more credits for it. it's actually super awesome.

- linting workflows/snippets is done in ~40 seconds.
- any linting we write that doesn't need ruby, can be done in <40 seconds.
- any linting that needs the compiled site can start after ~3minutes.
- html proofing done in ~5 minutes.

the configuration is SO much easier to read, too. So we can get feedback to the user super quickly for big issues if we lint for them. e.g. I'm updating the broken boxes detection so it can run without a built copy of the site, and then that will be able to run in ~40 seconds, too.

Other benefits:

- gh knows which are PRs vs branches
- I've currently set it up to only test PRs, we can open PRs from branches without the 2x builds.
- deploys can happen from master and be significantly faster.

it's looking very positive


shiltemann (approved):

shiltemann:
whoo! thanks @hexylena :tada: and so much faster too, I'm excited

PR #2157, opened 2020-11-27
Cut columns from a table -> Cut columns from a table (cut)

yvanlebras:
This is something where for sure trainees will make mistake... at least with this tool and the concatenate one.... Maybe only keep one "version" of each tools ? ITM be sure we can make the same operations, it seems to me this is not the case so "we" have to work on it carefully

### DIFF
--- a/topics/ecology/tutorials/PAMPA-toolsuite-tutorial/tutorial.md
+++ b/topics/ecology/tutorials/PAMPA-toolsuite-tutorial/tutorial.md
@@ -440,7 +440,7 @@ nomenclature for the "observation.unit" fields, namely:
 >    >
 >    {: .question}
 >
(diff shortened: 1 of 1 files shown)


shiltemann (approved):

shiltemann:
thanks @yvanlebras 

I agree, some of these tool names are very similar and confusing for new users, but not so easy to solve. That was part of the reason we made the GTN webhook, then when you click the toolname in the tutorial it automatically opens the correct tool, which could be a nice feature to show new users!

yvanlebras:
> 
> 
> thanks @yvanlebras
> 
> I agree, some of these tool names are very similar and confusing for new users, but not so easy to solve. That was part of the reason we made the GTN webhook, then when you click the toolname in the tutorial it automatically opens the correct tool, which could be a nice feature to show new users!

+100 an issue here was that the tutorial creator made the mistake, so propagating the wrong tool name and version ;) And this is due to the original issue of such very similar tools with so similar names... Thank you !!

PR #2165, opened 2020-12-01
Ansible tutorial misc. edits

### LINKED ISSUE #2166: Ansible tutorial: replace geerlingguy.git with alternative that is not already installed on VMs
The Ansible tutorial expects the user to have a failed playbook run [when installing a package](https://training.galaxyproject.org/training-material/topics/admin/tutorials/ansible/tutorial.html#question-question-2). However, this doesn't happen because Git is already installed. 
Solution: we need to use a different module as an example.

(This was discussed prior to and during admin training at BCC 2020.)

ic4f:
Fix formatting, punctuation, consistency; update structure of sample cvmfs role; fix more broken links (same issue as in ed434d7c95f3a4fc8a3c3a9161d2201459f40f6d); address misc. issues from BCC 2020.

Fixes #2166.


topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -607,7 +607,7 @@ Now that you've built a small role, you can imagine that building real roles tha
 >    > > ```
 >    > > - downloading role 'git', owned by geerlingguy
 >    > > - downloading role from https://github.com/geerlingguy/ansible-role-git/archive/3.0.0.tar.gz
->    > > - extracting geerlingguy.git to /tmp/tmp.Jm681gks8d/roles/geerlingguy.git
+>    > > - extracting geerlingguy.git to /home/xhr/ansible/intro/roles/geerlingguy.git
```
hexylena, line 610: This should probably be ubuntu for the user, right?

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -171,7 +173,7 @@ And an inline style.
 {% raw %}- package name={{ package_name }} state={{ package_state }}{% endraw %}
 ```
 
-Some groups prefer one style or another. You can mix both of these but you probably shouldn't. The inline style does not require quoting of templated values.
+Some groups prefer one style or another. You can mix both of these but you probably shouldn't. The YAML style requires that the templated value is quoted (if the value after the colon starts with a "{"). The inline style does not require quoting of templated values.
```
hexylena, line 176:
```suggestion
Some groups prefer one style or another. You can mix both of these but you probably shouldn't. The YAML style requires that the templated value is quoted (if the value after the colon starts with a `{`). The inline style does not require quoting of templated values.
```

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -102,9 +103,11 @@ We can look at the [ansible-cvmfs role](https://github.com/galaxyproject/ansible
 │   ├── main.yml
 │   └── ...
 ├── templates
-│   └── stratum1_squid.conf.j2
+│   ├── stratum1_squid.conf.j2
+│   └── ...
 └── vars
     ├── debian.yml
+    ├── main.yml
     └── redhat.yml
```
nsoranzo, line 111:
```suggestion
    ├── ...
    ├── main.yml
    └── ...
```

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -147,9 +149,9 @@ A `tasks/main.yml` file calls multiple Ansible modules to accomplish its goal. A
...
+The first invokes the [`yum`](https://docs.ansible.com/ansible/2.9/modules/yum_module.html) module with the arguments `name: cvmfs, state: ...`. This will use yum to install the package named `cvmfs`. The state parameter uses a [Jinja2](http://jinja.pocoo.org/) template to evaluate the value of the variable `cvmfs_upgrade_client`. We can [grep through](https://github.com/galaxyproject/ansible-cvmfs/search?q=cvmfs_upgrade_client&unscoped_q=cvmfs_upgrade_client) the repository and see that `defaults/main.yml` sets that to `false` by default. We can override this if we need, we'll come back to that later. The first task also has a `when` condition to ensure it only runs on RHEL family machines, so RedHat or CentOS. It is better to use the OS-generic [`package`](https://docs.ansible.com/ansible/2.9/modules/package_module.html) module, if possible.
```
nsoranzo, line 152: The "Install CernVM-FS client (yum)" task above has been substantially changed in the `ansible-cvmfs` role and doesn't use `cvmfs_upgrade_client` directly any more. I'd suggest to replace that part of the example with:

```yaml
- name: Download cvmfs_preload utility when desired
  get_url:
    url: https://cvmrepo.web.cern.ch/cvmrepo/preload/cvmfs_preload
    dest: "{{ cvmfs_preload_path }}/cvmfs_preload"
    owner: root
    group: root
    mode: 755
  when: cvmfs_preload_install | bool
```

from https://github.com/galaxyproject/ansible-cvmfs/blob/master/tasks/client.yml , which still use a standard module ([get_url](https://docs.ansible.com/ansible/2.9/modules/get_url_module.html)), Jinja2 with a greppable variable, and a `when` condition.

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -171,7 +173,7 @@ And an inline style.
 {% raw %}- package name={{ package_name }} state={{ package_state }}{% endraw %}
 ```
 
-Some groups prefer one style or another. You can mix both of these but you probably shouldn't. The inline style does not require quoting of templated values.
+Some groups prefer one style or another. You can mix both of these but you probably shouldn't. The YAML style requires that the templated value is quoted (if the value after the colon starts with a `{`). The inline style does not require quoting of templated values.
```
nsoranzo, line 176:
```suggestion
Some groups prefer one style or another. You can mix both of these but you probably shouldn't. In the YAML style the templated value needs to be quoted if the value after the colon starts with a `{` (see https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#when-to-quote-variables-a-yaml-gotcha ). The inline style does not require quoting of templated values.
```

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -301,10 +303,10 @@ The above introduction was certainly not enough for you to feel confident in Ans
...
->    > You can usually find a module that will represent most commands you will run at a shell. Usually by searching the internet for "ansible do-some-action" e.g. "ansible copy file to server" or "ansible restart service." If you cannot find a module that does it, there is the [`command`](http://docs.ansible.com/ansible/2.9/command_module.html) module, but this should be avoided if possible. Expect to have a browser session with 10-30 different Ansible module documentation tabs if you work with Ansible regularly, no one remembers what arguments are available to every module.
+>    > You can usually find a module for most commands you will run in a shell. Usually by searching the internet for "ansible do-some-action" (e.g. "ansible copy file to server" or "ansible restart service)." If you cannot find a module that does it, there is the [`command`](http://docs.ansible.com/ansible/2.9/modules/command_module.html) module, but this should be avoided if possible. Expect to have a browser session with 10-30 different Ansible module documentation tabs if you work with Ansible regularly, no one remembers what arguments are available to every module.
```
nsoranzo, line 309:
```suggestion
>    > You can usually find a module for most commands you will run in a shell, e.g. by searching the internet for "ansible copy file to server" or "ansible restart service". If you cannot find a module that does it, there is the [`command`](http://docs.ansible.com/ansible/2.9/modules/command_module.html) module, but this should be avoided if possible. Expect to have a browser session with 10-30 different Ansible module documentation tabs if you work with Ansible regularly, no one remembers what arguments are available to every module.
```

topics/admin/tutorials/ansible/tutorial.md
ic4f: I wasn't sure we wanted to expose that username - I don't remember if we used it for attendees last time, or just trainers, so I kept `xhr`. So yes, let's use `ubuntu` then. I'll change it for all other occurrences too.

topics/admin/tutorials/ansible/tutorial.md
ic4f: I suggest keeping the distribution names here: they are mentioned as an example in the text describing the `var` folder.

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -102,9 +103,11 @@ We can look at the [ansible-cvmfs role](https://github.com/galaxyproject/ansible
 │   ├── main.yml
 │   └── ...
 ├── templates
-│   └── stratum1_squid.conf.j2
+│   ├── stratum1_squid.conf.j2
+│   └── ...
```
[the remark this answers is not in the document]
ic4f: We don't mention the `tests` folder, so I don't think it's relevant in the context of this tutorial. (That scheme doesn't have to reflect the entire structure of the cvmfs role, I think?)

topics/admin/tutorials/ansible/tutorial.md
ic4f: I'll make the change

topics/admin/tutorials/ansible/tutorial.md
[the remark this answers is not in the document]
nsoranzo: Fair enough, I've updated the suggestion.

topics/admin/tutorials/ansible/tutorial.md
nsoranzo: The 2 files have been renamed in the role. I don't think we need to mention them explicitly any way, since the `vars` description below doesn't rely on having separate files to define the variables.

topics/admin/tutorials/ansible/tutorial.md
ic4f: done.

topics/admin/tutorials/ansible/tutorial.md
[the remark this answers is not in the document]
ic4f: Thanks!

topics/admin/tutorials/ansible/tutorial.md
ic4f: @nsoranzo I've updated the example; could you check the wording, please?

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -147,9 +153,9 @@ A `tasks/main.yml` file calls multiple Ansible modules to accomplish its goal. A
...
+The first invokes the [`get_url`](https://docs.ansible.com/ansible/2.9/modules/get_url_module.html) module with the arguments `url: ..., dest: ..., owner: root, group: root, mode: 755`. This will download a file from `url` to the location specified in `dest`, change user and group ownership to `owner` and `group`, respectively, and assign permissions specified in `mode`. The `dest` parameter uses a [Jinja2](http://jinja.pocoo.org/) template to evaluate the value of the variable `cvmfs_preload_path`. We can [grep through](https://github.com/galaxyproject/ansible-cvmfs/search?q=cvmfs_preload_path) the repository and see that `defaults/main.yml` sets that to `/usr/bin` by default. We can override this if we need, we'll come back to that later. The first task also has a `when` condition to ensure it only runs when the `cvmfs_preload_install` variable is set.
```
nsoranzo, line 156:
```suggestion
The first example invokes the [`get_url`](https://docs.ansible.com/ansible/2.9/modules/get_url_module.html) module with 5 arguments (`url`, `dest`, `owner`, `group` and `mode`). This will download a file from the location indicated in `url` to the directory specified by `dest`, change user and group ownerships to `root`, and change file permissions to `755` (as assigned by the `mode` argument). The `dest` parameter uses a [Jinja2](http://jinja.pocoo.org/) template to evaluate the value of the variable `cvmfs_preload_path`. We can [grep through](https://github.com/galaxyproject/ansible-cvmfs/search?q=cvmfs_preload_path) the repository and see that `defaults/main.yml` sets that to `/usr/bin` by default. We can override this if we need, we'll come back to that later. The first task also has a `when` condition to ensure it only runs when the `cvmfs_preload_install` variable is set.
```

ic4f: Thanks!

nsoranzo (dismissed):

ic4f:
@hexylena I've made the change we [discussed on gitter](https://gitter.im/dagobah-training/Instructors?at=5fc5d165d61cdb6a209f5ec4) - I've tried to integrated relevant bits from both, your comments and the previous version. I've removed the hands-on altogether, I think there's no point in keeping it as such. Can you take a look?
@nsoranzo I believe my new commit automatically "dismissed" your approval. There's at least one more edit coming.

ic4f:
04541b0 fixes #2166.

ic4f:
@hexylena @nsoranzo This is ready for review, no more pending edits for now.

topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -69,14 +69,28 @@ Looking at each of these briefly:
 
 ```ini
 [webservers]
-web_a
-web_b
+web_1.host.fqdn.or.ip
+web_2.host.fqdn.or.ip
```
hexylena, line 73: maybe we show a hostname example below, maybe just use an IP here. [ip ranges for docs](https://tools.ietf.org/html/rfc5737)

hexylena (dismissed):
looks fine to me! thanks for this update @ic4f

topics/admin/tutorials/ansible/tutorial.md
ic4f: Done. And yes, I think it's better this way.

ic4f:
New commit automatically dismissed approvals - merging blocked.

hexylena (approved):

PR #2176, opened 2020-12-03
Add MRSA assembly with Nanopore & Illumina versions

hexylena:
@bazante1 @miaomiaozhou88 here it is :)

I've added a track for the staramr results directly in JBrowse so students don't have to compare back and forth.

image

*todo*

- [x] resolve duplication from amr tutorial
- [x] nanopore snippet?
- [x] 'amr' should find both tutos.
- [x] Need to reduce the size of the bam datasets.
- [x] Reviewed by Carla (AVANS)


gallardoalba:
Zenodo links don't seem to work.

hexylena:
Ahh yeah, good catch. Thanks @gallardoalba. Converted to draft pending more comments from my colleagues.

hexylena:
Ok this has been reviewed sufficiently internally and ready for external review.

**Background**

This is what we want to use for our course this year. Students will be running the public data + some "mystery" RIVM data through this tutorial and figuring out which one is MRSA and which is S. aur. (I think)

This tutorial layout is currently *suboptimal* but this is what's committed to being taught this year. Next year we commit to reorganising this as follows:

- RIVM data will be public so everyone can follow along
- We will remove these two tutorials
- Instead we will have two different tutorials:
  - one will do three different assemblies: Nanopore, Illumina, Hybrid and will look in more detail at their outputs 
  - One will do the (identical) second half of both tutorials, and analyse the resulting 'best' assembly / map both sequencing datasets / do the staramr on that.

shiltemann (approved):
this looks great! thanks a bunch @bazante1 @miaomiaozhou88 @hexylena!! And thanks for the review @gallardoalba!

PR #2186, opened 2020-12-07
[Small Patch] Format tutorial for paper

mtekman:
Fix to regex for non backwards relative paths

    python bin/format_tutorial_for_article.py -t topics/transcriptomics/tutorials/scrna-raceid

  Before it would error out an image entry such as:

    'Histograms of raw and filtered data\n'

Also, the `check_exists` function modified to allow boolean return. This can be used to check whether or not the tutorial path needs to  be appended to the image path (depending on whether or not the image path is relative)

### DIFF
--- a/bin/format_tutorial_for_article.py
+++ b/bin/format_tutorial_for_article.py
@@ -325,10 +325,13 @@ def copy_images(images, images_dp, tuto_dp):
     :param images_dp:
     '''
     for img in images:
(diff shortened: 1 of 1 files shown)


shiltemann (approved):
looks good to me, not sure if @bebatut wants to look at this as well?

thanks @mtekman!

mtekman:
thanks @shiltemann ! There's actually no need to merge it, as I understand this is still alpha code and my edits likely do not follow the best python practices -- the PR is more of a general feedback

bebatut:
Thanks @mtekman

PR #2200, opened 2020-12-10
Add google translate links to slide buttons.

blankenberg:


hexylena:
This doesn't seem to work at all under firefox which sucks, but fine in chrome. We could probably de-dupe, but we'll be re-organise the introduction slides soon anyway and can deal with that then.

hexylena (approved):

blankenberg:
Aww, that's too bad about firefox. Definitely agree that deduping and generalizing would be great. Thanks for the quick review, edits, and merge!

PR #2205, opened 2020-12-17
Populate robots.txt file when served from a fork

shiltemann:
To allow users to build the website on their own fork, without search engines indexing it

also removes google analytics since we switched to Plausible

ping @hexylena please check the changes to the github actions workflows

### DIFF
--- a/.github/workflows/deploy-s3.yml
+++ b/.github/workflows/deploy-s3.yml
@@ -49,6 +49,7 @@ jobs:
           JEKYLL_ENV=production bundle exec jekyll build --strict_front_matter -d _site/training-material
         env:
           SOURCE_TAG: ${{ steps.branch_name.outputs.SOURCE_TAG }}
+          GTN_FORK: ${{ secrets.GTN_FORK }}
 
       - name: Deploy 🚀
         run: |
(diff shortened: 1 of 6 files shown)


hexylena (approved):

.github/workflows/deploy-s3.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           JEKYLL_ENV=production bundle exec jekyll build --strict_front_matter -d _site/training-material
         env:
           SOURCE_TAG: ${{ steps.branch_name.outputs.SOURCE_TAG }}
+          GTN_FORK: ${{ secrets.GTN_FORK }}
```
nsoranzo, line 52: I think this could be `${{ github.repository_owner }}` instead?

.github/workflows/deploy.yml
```diff
@@ -38,6 +38,8 @@ jobs:
           export PATH=/home/runner/.local/bin:$PATH
           make annotate ACTIVATE_ENV=pwd
           JEKYLL_ENV=production bundle exec jekyll build --strict_front_matter -d _site/training-material
+        env:
+          GTN_FORK: ${{ secrets.GTN_FORK }}
```
nsoranzo, line 42: As above

.github/workflows/deploy-s3.yml
shiltemann: ah, that would be cleaner, thanks

bebatut:
Could we keep Google analytics a bit to see the evolution over time?

PR #2209, opened 2020-12-17
Update visualization circos training

gallardoalba:
A comment block has been included to inform that it is necessary to use Circos version 0.69.8+galaxy4 for this tutorial since some options cannot be modified in more recent versions. In particular, from version 0.69.8+galaxy5 to 0.69.8+galaxy7 is not possible to select the cytogenetic bands and the option Convert bands from BED format to circos karyotype band format is not available.
In addition, the option Fill Band has been renamed as Band Transparency.


hexylena:
Hey @gallardoalba, I think I'd rather make this tutorial compatible with the `7` version properly, rather than downgrading to the older version of the tool which has numerous known bugs. Is that OK with you?

gallardoalba:
Sure @hexylena, but then it will be necessary to update some parts of the tutorial.
Regards

hexylena:
Yeah, I can care for this. Thanks!

PR #2216, opened 2020-12-31
Annotation with maker: short version + update

abretaud:
Let's finish 2020 with a PR on the annotation tutorials!
- added the new tool buttons + levels to Maker and Prokka tutorial
- updated the Maker tutorial (new tool versions, updated workflow, various little changes)
- made a short version of this Maker tutorial, removing the training steps that are soooo long and usually confuse the students when they need to run 3 times the same tools


abretaud:
The test is still failing on a link but I don't understand why, any clue?

hexylena:
```
<a href="/training-material/topics/genome-annotation/tutorials/annotation-with-maker/tutorial.html to understand how they were trained">longer version of tutorial</a>
```

looks like `]` in the wrong place, and the url got a bit more text than expected.

abretaud:
too simple :D thanks!

hexylena (approved):

PR #2152, opened 2020-11-24
Update admin slides for 2021 GAT + videos

hexylena:
xref https://github.com/galaxyproject/admin-training/issues/104

PR #2220, opened 2021-01-04
add wf to build videos

hexylena:
edit: probably needs #2152


hexylena:
This is ready to go, I think. The commit prior to the tip shows it building all of the videos (minus upload because AWS keys not available on branches).

The last commit switches it to be a cron job that'll run every two days and rebuild only the changed videos to keep costs down.

Not currently implemented is a cache for the voice snippets, but, this is maybe not necessary given the architecture.

shiltemann (approved):

PR #2226, opened 2021-01-06
Fix another yml

abretaud:
Just as #2225 with another yml (tested now, no new similar pr hopefully)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,8 @@ jobs:
           fetch-depth: 1
       - name: Lint 🔍👁👁
         uses: ibiqlik/action-yamllint@v3
+        with:
+          file_or_dir: topics/
 
   build-site:
     needs: setup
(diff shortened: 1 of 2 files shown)


hexylena:
How are you testing? Is there something we should add to the linting?

abretaud:
I'm testing with https://github.com/galaxyproject/training-material/blob/master/bin/mergeyaml.py
I could add a GA step to run it, sure
(which I use to update data libs on my instance using something derived from https://github.com/usegalaxy-eu/shared-data/)

hexylena:
That'd be ace! Cheers!

hexylena:
We can probably make the linter be less strict, for what it's worth? Or, if it's easier, if `bin/merge-yaml` crashes when running, then, we could just use that as the test?

abretaud:
Yes, I hoped for a cleaner output with yaml-lint, but it's not doing what I want apparently... I will switch to the merge script yep

abretaud:
Ready, you might want to squash my mess ;)

hexylena (approved):

hexylena:
thanks for adding this!

Issue #2223, opened 2021-01-05
Wording on tutorial for Galaxy deploy with ansible


profgiuseppe:
In the tutorial about using ansible to deploy galaxy (https://training.galaxyproject.org/training-material/topics/admin/tutorials/ansible-galaxy/tutorial.html) I found a pair of confusing/conflicting statements. The galaxy_layout variable states that 

> all of the code, configuration, and data folders will live beneath galaxy_root

but then we have the file_path variable stating that it is 

> a place to store data

Looking to an installation I made few months ago, it appears that the first description is not accurate (e.g. under galaxy_root I get the code and configuration, under file_path I get the data files).

This may be clarified by either referring to the second variable ("unless otherwise specified via file_path") or by specifying the kind of data under galaxy_root (if there are different types of data).

This will also help in understanding what parts of galaxy must be on a shared path in case of cluster installation (e.g. if only file_path or also other parts of the directory tree).

PR #2229, opened 2021-01-06
Fix ansible galaxy issues

hexylena:
Ansible broke something in 2.10.4 for us, and now we have to document a workaround explicitly when it used to be just done implicitly.

Additionally this addresses the issue from #2223

### DIFF
--- a/topics/admin/tutorials/ansible/tutorial.md
+++ b/topics/admin/tutorials/ansible/tutorial.md
@@ -272,14 +272,14 @@ The above introduction was certainly not enough for you to feel confident in Ans
 >
 >    1. We will call our group `my_hosts`
 >
->    2. Create [an inventory file](https://docs.ansible.com/ansible/2.9/user_guide/intro_inventory.html) with the group `my_hosts` and `localhost ansible_connection=local`, which tells ansible to not use SSH, and just use the local connection.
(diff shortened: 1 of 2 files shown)


topics/admin/tutorials/ansible/tutorial.md
```diff
@@ -272,14 +272,14 @@ The above introduction was certainly not enough for you to feel confident in Ans
 >
 >    1. We will call our group `my_hosts`
 >
->    2. Create [an inventory file](https://docs.ansible.com/ansible/2.9/user_guide/intro_inventory.html) with the group `my_hosts` and `localhost ansible_connection=local`, which tells ansible to not use SSH, and just use the local connection.
+>    2. Create [an inventory file](https://docs.ansible.com/ansible/2.9/user_guide/intro_inventory.html) with the group `my_hosts` and `localhost ansible_connection=local`, which tells ansible to not use SSH, and just use the local connection. Additionally, you should explicitly set the `ansible_user` to your username on the server. Ansible has changed it's behaviour over time regarding whether or not `ansible_user` is defined, and it's most effective to define it explicitly even when it sometimes can be inferred.
```
nsoranzo, line 275:
```suggestion
>    2. Create [an inventory file](https://docs.ansible.com/ansible/2.9/user_guide/intro_inventory.html) with the group `my_hosts` and `localhost ansible_connection=local`, which tells Ansible to not use SSH, and just use the local connection. Additionally, you should explicitly set the `ansible_user` variable to the username to use when connecting to the server. Ansible has changed its behaviour over time regarding whether or not `ansible_user` is defined, and it is most effective to define it explicitly even when it can sometimes be inferred.
```

nsoranzo (approved):

PR #2233, opened 2021-01-08
Use "singularity run" for verifying installation

torfinnnome:


hexylena (approved):
Ahh that makes sense.

PR #2236, opened 2021-01-08
Using galaxy_user variable also for miniconda become_user

profgiuseppe:
If for some reason the value of galaxy_user.name gets changed, the playbook will fail when the uchida.miniconda role tries to become an fixed (e.g. "galaxy") user. The change fixes that. The required variable is added immediatly after the writing of the playbook, so it should not be an issue.


nsoranzo (approved):

nsoranzo:
I also have this in my setup, thanks @profgiuseppe !

PR #2241, opened 2021-01-12
Minor updates to tool management tutorial for admin training 2021

cat-bro:
The installed tool for the "Hands-on: Installing a single tool" step has been changed from bwa to pilon.  The tested tool has been changed from bamtools_filter to pilon.  The reason for this change is to avoid the error 'NoneType has no attribute view' which comes from running tests for some tools where pysam is not also installed.  pilon was chosen because it doesn't take up much more conda space than bwa and it installs is about the same amount of time.

There is a new tip for splitting the screen with byobu, because at some point the video tutor will split their byobu screen and instruct the students to do the same.

### DIFF
--- a/topics/admin/tutorials/tool-management/tutorial.md
+++ b/topics/admin/tutorials/tool-management/tutorial.md
@@ -102,7 +102,7 @@ Ephemeris can take care of this process. Let's practice this on a real worfklow.
 >    >
 >    > > ### {% icon solution %} Solution
 >    > > ```console
->    > > workflow-to-tools -w mapping.ga -o workflow_tools.yml -l "mapping tools"
+>    > > workflow-to-tools -w mapping.ga -o workflow_tools.yml -l Mapping
 >    > > ```
 >    > {: .solution }
 >    {: .question}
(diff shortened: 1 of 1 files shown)


hexylena (approved):
@cat-bro I've tested everything here with the singularity setup, and it all works with no changes, so, that's exciting. These changes look great!

cat-bro:
@hexylena awesome!  Will the students have the singularity setup when they take this tutorial?

hexylena:
Aye, I'll put it to the start of tuesday or so.

cat-bro:
What changes will I need to make to the VM I am using so that I'm using the same setup as the students will have?

hexylena:
Just follow https://training.galaxyproject.org/training-material/topics/admin/tutorials/singularity/tutorial.html, you can skip the tool installation of course.

Next year we should figure out some more global way to setup ephemeris, there are *three* tutorials which need tools (singularity, ephemeris, pulsar) and we twice write the instructions to do it manually because we either don't want to assume shed-tools is there (or we want to do it after we setup singularity for the job configuration), and even if it is there it'd be too complicated to be like "remember that API key from yetserday? and where you put your venv?"

PR #2247, opened 2021-01-13
YAML array version of the tool_data_table_config_path doesn't work

Slugger70:


hexylena (approved):
ugh, yes, how did I forget this.

PR #2248, opened 2021-01-13
add raw tags around a yaml block in singularity tutorial

cat-bro:
the singularity_go_path variable disappears and reads as `singularity_go_path: ""`


nsoranzo (approved):

hexylena:
thanks @cat-bro, good catch. We should check every tuto for these issues.

PR #2260, opened 2021-01-15
Add subtopics for transcriptomics

shiltemann:
There are a LOT of tutorials in the transcriptomics topic, it would be nice to create some subtopics. I had a first go at it, but am not an expert, so please let me know what would be good groupings, @mblue9, @bebatut, and others?

Currently I made the following subtopics:
- Introduction (with intro slides and ref-based tutorial)
- End-to-End (@mblue9's 3 tutorials that form a unit, please let me know what a better name would be here?)
- Single Cell (@mtekman, @nomadscientist let me know if there is some logical ordering to these that might be nice to enforce?)
- Visualisation
- Anything else ends up in "Other"

Thoughts? Any subtopics we should add/remove/rename? Tutorials that need to move to a different subtopic?
 
image

### DIFF
--- a/_includes/tutorial_list.html
+++ b/_includes/tutorial_list.html
@@ -18,6 +18,7 @@
   </thead>
   <tbody class="list">
 
+  {% assign topic_material = topic_material | sort: 'priority'%}
   {% for material in topic_material %}
     {% if include.sub and include.sub.id == material.subtopic %}
       {% assign show = true %}
@@ -31,7 +32,7 @@
       {% if material.enable == nil or material.enable == true %}
         <tr>
           <td class="tutorial_title">
-          {{ material.title }}
+          {{ material.title}}
           <div class="">
           {% if material.level %}
             {% include _includes/difficulty-indicator.html level=material.level %}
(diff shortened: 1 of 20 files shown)


nsoranzo:
Maybe "De novo transcriptome reconstruction with RNA-Seq" can go into Introduction, since the ref-based is there as well.

mtekman:
Thanks @shiltemann !

I'd recommend the single-cell order to be:

1. Pre-processing of single-cell RNA Data
1. Understanding Barcodes
1. Plates, Batches, and Barcodes
1. Single-cell quality control with scater
3. Downstream Single-cell RNA analysis with RaceID
2. Pre-processing of 10X single-cell RNA Data
3. Clustering 3K PBMCs with ScanPy

and the Blockclust tutorial is possibly maybe not single-cell (I am not sure)? 

thoughts @bebatut @nomadscientist

shiltemann:
thanks @mtekman! (yep, the blockclust tutorial was an accident on my part, fixed now)

hexylena (approved):

PR #2263, opened 2021-01-18
Update BioBlend slides for video

nsoranzo:
xref. https://github.com/galaxyproject/admin-training/issues/105

### DIFF
--- a/topics/dev/tutorials/bioblend-api/slides.html
+++ b/topics/dev/tutorials/bioblend-api/slides.html
@@ -1,26 +1,23 @@
 ---
 layout: tutorial_slides
 logo: "GTN"
+video: true
 
 title: "Scripting Galaxy using the API and BioBlend"
 questions:
-  - "What is a RESTful API?"
-  - "How to interact with Galaxy in a script?"
-  - "Why and when should I use BioBlend?"
-objectives:
-  - "Describe the principles of RESTful APIs"
-  - "Select the most efficient way to work with Galaxy"
-  - "Use the API and BioBlend to interact with Galaxy"
-requirements:
-  -
-    title: "IPython"
-    type: "external"
-    link: "https://ipython.org"
+  - What is a REST API?
+  - How to interact with Galaxy programmatically?
+  - Why and when should I use BioBlend?
+#requirements:
+#  -
+#    title: "IPython"
+#    type: "external"
+#    link: "https://ipython.org"
 time_estimation: "2h"
 key_points:
-  - "The API allows you to use Galaxy's capabilities programmatically"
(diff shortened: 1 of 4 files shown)


topics/dev/tutorials/bioblend-api/slides.html
```diff
@@ -1,26 +1,23 @@
...
-    type: "external"
-    link: "https://ipython.org"
+  - What is a REST API?
+  - How to interact with Galaxy programmatically?
+  - Why and when should I use BioBlend?
+#requirements:
+#  -
+#    title: "IPython"
+#    type: "external"
+#    link: "https://ipython.org"
```
nsoranzo, line 15: I had to comment out the requirements here because the corresponding slide was without audio/subtitles, therefore messing up the audio sync with all following slides.

hexylena: Thanks for letting me know! I'll get that fixed so we can uncomment this again, I'm surprised nothing else triggered that bug!

topics/dev/tutorials/bioblend-api/slides.html
```diff
@@ -263,18 +336,32 @@
...
+  - Need to extract the entity ID for further processing
 - No explicit modeling of Galaxy entities and their relationships
 - Complex operations still need many function calls
   - Need for higher-level functionality
 
+???
+
+- Although easier to use than the Galaxy API, BioBlend has some limitations.
+- First of all, it's only available for Python.
+- There are alternative libraries for Java and PHP, but they are less complete.
```
nsoranzo, line 360: A weirdness that I noticed when testing with Mozilla TTS: anything after "PHP" is cut out.

hexylena: Yeahhh, I've had this a ton, and @eancelet experienced this too, everything after `Jupyter,` was cut. Removing the `,` fixed it in Moz TTS but, it works fine either way in AWS.

The mozilla tts has a lot of bugs but it's primary purpose isn't producing the final output, just producing *something* so contributors have an idea of what the final video will look like without having to setup AWS credentials and potentially pay money

hexylena (approved):
This looks good, thanks for submitting this!

nsoranzo:
Another small issue I noticed is that if there is a `-` in the middle of a speaker note line, the `bin/ari.sh` crashes. Easy to work around, but a bit difficult to find what the problem is.

hexylena:
Ahh, had not encountered that in any of the other slides yet. (Feel free to just ping me when it crashes without debugging first, my bad code after all.)

I'll see if I can track that down at some point.

hexylena (approved):

nsoranzo:
Thanks for the fixes!

PR #2265, opened 2021-01-18
sars-cov-2 assembly: minor fixes

abretaud:
Just 3 tiny fixes for the sars-cov-2 assembly tutorial


wm75 (approved):
Good catches. Thank you!

PR #2279, opened 2021-01-21
Update galaxy_ansible

hexylena:
@mvdbeek patched container issues in https://github.com/galaxyproject/galaxy/pull/11187 but in turn this required that we set some variables from the start. I've updated the role accordingly and now we need to update this training.

side note: this is what is so miserable about admin training. Singularity tutorial → bug fix in galaxy → update singularity tuto → update ansible-galaxy role → update ansible-galaxy tutorial → ??? maybe other things broken now? → retest from an empty VM to be sure.


mvdbeek (approved):

hexylena:
thanks!

PR #2293, opened 2021-01-26
Rephrase title of UI topic

shiltemann:
In our last GTN meeting, @nomadscientist and others remarked that "Interface" is a very tech/dev term, and probably not very clear to most users. Here I've rephrased the topic title to be more accessible.

### DIFF
--- a/topics/galaxy-interface/metadata.yaml
+++ b/topics/galaxy-interface/metadata.yaml
@@ -1,7 +1,7 @@
 ---
 name: galaxy-interface
 type: "basics"
-title: "User Interface and Data Manipulation"
+title: "Using Galaxy and Managing your Data"
 summary: "A collection of microtutorials explaining various features of the Galaxy user interface and manipulating data within Galaxy."
 docker_image:
 requirements:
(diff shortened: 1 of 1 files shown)


topics/galaxy-interface/metadata.yaml
```diff
@@ -11,7 +11,7 @@ subtopics:
     title: "Uploading Data"
     description: "These tutorials will help you get data into Galaxy, before you analyse it."
   - id: interface
-    title: "Galaxy Interface"
+    title: "Galaxy Features"
     description: "These tutorials will cover the UI and important features"
```
nsoranzo, line 15: UI is even less accessible than interface! Maybe "Learn about the most useful Galaxy features" or something like that?

hexylena: yes, that was the point of this pr? removing the "UI" from the homepage.

The description can definitely be better (but no one reads those since they're on the next page)

nomadscientist: "These tutorials will cover how to navigate Galaxy efficiently" or something?

PR #2310, opened 2021-01-29
GxIT: clarify wildcard certificate

abretaud:
*Trying to* clarify at least! Hope it's better like this


hexylena (approved):

hexylena:
yeah this is better mayube

PR #2312, opened 2021-01-29
Create a  page with video slides  per topic

shiltemann:
Now that we are getting more and more video slides, it would be nice to have a page per topics showing all the videos, instead of a single huge page.

Also added a TOC for because some topics (such as admin) have a lot of video slides

image

ping @hexylena


yvanlebras (dismissed):
Not so many change for ecology ;) But really nice feature !!

shiltemann:
> Not so many change for ecology ;) But really nice feature !!

yeahh, sorry for the spam, I had to touch every topic for this :P

hexylena (dismissed):

hexylena (approved):

PR #2331, opened 2021-02-09
[Update] Tool-links to some scrna materials

mtekman:
This adds the latest tool versions as tool links to some of the training materials:

* scrna-umis
* scrna-raceid
* scrna-scater-qc
* scrna-preprocesing

The scanpy 3k stuff was left because that requires some manual verification on the tool versions, and will be addressed in a separate PR


hexylena (approved):
Perfect, thank you so much!

PR #2332, opened 2021-02-09
Update tutorial.md

nomadscientist:
Inspect and Manipulate, as well as the Loom Import, are both older tool versions so as to not clash with the other tools.
Also, fingers crossed I did this right...


hexylena (approved):
This is fantastic @nomadscientist! I'm so happy to see this being used, and, it can ensure that we can accurately tell which servers support these.

PR #2350, opened 2021-02-12
Update JBrowse option

ic4f:
This is what I see on usegalaxy.org. I assume this is the equivalent (updated) option?


hexylena (approved):
yep

PR #2353, opened 2021-02-12
Update M. tuberculosis variant analysis tutorial

pvanheus:
* Mention need to run `sed` on output of `snippy`
* Add example where sequence quality is good but mapping is poor
* Fix problem identified by @gmauro with data library


bgruening (approved):

bgruening:
Will try to deploy the data library again.

PR #2356, opened 2021-02-13
Fix last hands on section in M. tuberculosis variant analysis tutorial

pvanheus:
There was a missing `{: .hands_on}` at the close of the last hands on section, this fixes it


hexylena (approved):
eek! should've been caught by our broken box detection, thanks so much @pvanheus!

PR #2358, opened 2021-02-13
Add work towards extra M. tb variant analysis workflows - WIP

pvanheus:
This PR contains two new workflows for testing the last parts of the M. tuberculosis variant analysis [tutorial](https://training.galaxyproject.org/training-material/topics/variant-analysis/tutorials/tb-variant-analysis/tutorial.html). Neither of them work right now, but I'm starting a PR to bookmark this work and in case others want to take a look.

The workflow that uses kraken2 currently only works against usegalaxy.eu because the "Standard" databases get their date of production embedded in the name.


shiltemann:
@pvanheus going through old PRs, what should we do with this one? is it still relevant? close or keep open?

PR #2360, opened 2021-02-14
small changes to circos tutorial

nekrut:
- Added a warning about two Cut tools
- Added a figures with axes


hexylena:
This looks ok but I'd like to check further when I get home.

It doesn't hurt to note the cut tools, but we've stopped noting that as much, now that we have the GTN in galaxy webhook which is highly recommended to avoid all of these issues and the "tool search" problem as well. If you haven't tried it, please do and let me know what you think!! https://docs.galaxyproject.org/en/master/_static/20.09-gtn-in-galaxy.mp4

hexylena (approved):
lgtm, thanks @nekrut

nekrut:
> This looks ok but I'd like to check further when I get home.
> 
> It doesn't hurt to note the cut tools, but we've stopped noting that as much, now that we have the GTN in galaxy webhook which is highly recommended to avoid all of these issues and the "tool search" problem as well. If you haven't tried it, please do and let me know what you think!! https://docs.galaxyproject.org/en/master/_static/20.09-gtn-in-galaxy.mp4

Damn this is awesome!

PR #2374, opened 2021-02-16
Formatting fixes to the scanpy tutorial

mtekman:
AnnData 0.7.4 no longer works for Import or Manipulate, though it did at most two weeks ago.

@mvdbeek  is patching it as we speak https://github.com/galaxyproject/tools-iuc/pull/3483

Surprisingly no links need to be updated, as all the relevant versions are correctly linked in the source material.... 
This is just a formatting fix PR then

### DIFF
--- a/topics/transcriptomics/tutorials/scrna-scanpy-pbmc3k/tutorial.md
+++ b/topics/transcriptomics/tutorials/scrna-scanpy-pbmc3k/tutorial.md
@@ -141,9 +141,9 @@ This format is used by [Scanpy](https://scanpy.readthedocs.io/en/stable/index.ht
 >            - {% icon param-file %} *"Barcodes"*: `barcodes.tsv`
 >            - *"Variables index"*: `gene_symbols`
 >            - *"Make the variable index unique by appending '-1', '-2'?"*: `Yes`
->
+> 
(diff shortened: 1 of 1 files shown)


hexylena (approved):

hexylena:
So, no tool update? If so this is fine to go.

mtekman:
Thanks -- and yeah it's strange that the user ran into an issue [this morning](https://gtnsmrgsbord.slack.com/archives/C01FQ9RLLF2/p1613457316041600), since the correct version of the tool was correctly specified...

No idea what the actual issue was

hexylena:
it can be they're not using the webhook. We should encourage that more :/

mtekman:
Oh for the tool update Marius is patching it - once it's patched and tested with the workflow, we can update the links again

Edit: Aaaah, that might have been it!

mvdbeek:
Yeah, would be good to know how many people are using the webhook.

mtekman:
If it's not that many, then maybe it would be good to disable the latest Anndata for tomorrow....

hexylena:
Unfortunately no stats on webhook usage. I can PR something to add a tracking campaign for those visits though, it would be nice to have.

PR #2385, opened 2021-02-18
fix filter tabular replacement expression

foellmelanie:


foellmelanie:
mh, I had only started the workflow, looking in the GTN there are two filter tabular steps with replacement expression
: 1; probably both should be changed to \1 @subinamehta ?

hexylena (approved):

PR #2387, opened 2021-02-18
More metaproteomics select format fixes.

jmchilton:


hexylena (approved):

hexylena:
We should come back and use a triple backtick with `sql` at some point in the future but this looks good for now :)

PR #2271, opened 2021-01-19
Introduce auto generated screenshots in rule-based uploader

OlegZharkov:
Replace screenshots by images produced by selenium in https://github.com/galaxyproject/ci-testing
The corresponding selenium tests are marked in galaxyproject/galaxy#11171 and galaxyproject/galaxy#10911

Please deploy locally and assure that all screenshots are correct.

PR #2393, opened 2021-02-22
Improved language for the compose parameter tool...

jmchilton:
... the workflow parameters tutorial.

The word repeat doesn't appear so I guess we should avoid it. I was thinking about including this screenshot also but maybe that is overkill / would go stale too quickly?

Screen Shot 2021-02-22 at 1 30 53 PM


hexylena (approved):
This definitely looks like an improvement. I think I'd be fine to include it here since it uses the workflow connection feature which probably is helped by a visual. (And we're nearly there to automated screenshots with #2271, so, even less worried about it going out of date.)

shiltemann:
thanks @jmchilton!

PR #2411, opened 2021-02-25
Add snippet plugin

hexylena:
This replaces the include command (it's honestly copied and pasted code from the include command of jekyll)

- it strips any markdown document header that might be in the .md
- renders to html
- removes newlines to prevent breaking the inclusion in markdown documents
- seems to work?

We no longer need to check snippets, and now snippets can include yaml metadata at the top, permitting some advancements we're planning.


hexylena:
Sorry for the spam everyone need to do a large refactoring. If this sort of notification bothers you, we're considering removing codeowners (due to my perceived lack of usefulness for folks) in #2410 so comment there if you have opinions.

shiltemann (approved):

PR #2417, opened 2021-02-26
Rephrase reference to "bits"

ic4f:
"Boolean bits" may be confusing to the reader: a Boolean is a data type, whereas a bit is a unit of information. So, "Boolean" cannot be used as a qualifier for "bits". If "...bits (0, 1)" is not clear enough, another option is "...bits (0, 1) representing Boolean values (true, false)"; but I would suggest just "bits (0, 1)" - for that's what the text is describing.


ic4f:
This particular phrase appears to be a quote from a [paper](https://chagall.med.cornell.edu/RNASEQcourse/Intro2RNAseq.pdf); however, I do think this edit will make our tutorial a (tiny) bit more accurate.

shiltemann (approved):

PR #2423, opened 2021-03-01
New toolfactory tutorial for dev

fubar2:
This tutorial is designed to help fill the gap between scripts working on the command line with small test data samples and real Galaxy tools. Scripts developed in interactive environments are one important use case. It is ideal for programmers and informaticians new to Galaxy. It is not designed for experienced or full time Galaxy developers who will be using planemo on the command line already and do not need a code generator.

The target audience is developers/informaticians who are dab hands at command line scripting in bash or some other interpreted language, are relatively new to Galaxy and who have completed the `Tool development and integration into Galaxy` tutorial material and perhaps the interactive environments tutorial. 

The reader is warned at the outset to check if the tutorial might be useful for their situation; and that scripting skills are essential to make much use of the ToolFactory and thus of the tutorial itself. No effort is made to teach the basics of tools or of command line scripting - this is a dev tutorial, but for a fairly small market segment - including script writing researchers who don't think of themselves as developers.

The ToolFactory itself has had little testing since undergoing a major rewrite to use galaxyxml to generate the new tool and the scripts and training material have had no testing other than by the author.


hexylena:
Hi @fubar2, please un-delete the Gemfile, that will probably fix the error.

hexylena:
The lockfile too please. They get moved by the makefile and people like to do git commit -a, and then they get wiped out. Just be careful when staging your commits.

fubar2:
> Hi @fubar2, please un-delete the Gemfile, that will probably fix the error.

Yep, that appears to have fixed it - no idea why it vanished - thanks!

fubar2:
Lint. Including a complaint about a welcome.html in the docker folder not having valid links. Linters are a condign punishment for life long pedantic bastards like me.

hexylena:
@fubar2  if linters are failing for urls that should be correct in your opinion, don't comment them out -- I'd rather fix the linter than disable a useful/correct url.

fubar2:
>I'd rather fix the linter than disable a useful/correct url.
Agreed - but I couldn't see any obvious way to notify that htmlchecker to ignore that particular html file so I commented the offending links out temporarily.

The style sheet matters a lot obviously so it would be much better to have a way to exclude html files from topic subdirectories  called "Docker" - this won't be the last one I suspect!

If not done yet, I'll post an issue?

hexylena:
This PR is coming from your master branch which usually means we will be prevented from pushing commits to it, even if you've checked the box. In the future please make PRs from separate branches if you can!

fubar2:
I can close and reopen on a branch later this morning....

fubar2:
PR updated to a branch in #2428 so closing this.

PR #2437, opened 2021-03-16
Add citations to scRNA galaxy papers

mtekman:
Add 2020 singlecell paper, and the 2019 paper.
   Mention the singlecell portal in relevant materials

### DIFF
--- a/topics/transcriptomics/metadata.yaml
+++ b/topics/transcriptomics/metadata.yaml
@@ -101,3 +101,9 @@ references:
     title: "BlockClust: efficient clustering and classification of non-coding RNAs from short read RNA-seq profiles"
     link: "doi:10.1093/bioinformatics/btu270"
     summary: ""
+  -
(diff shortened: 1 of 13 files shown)


hexylena (commented):
Congrats on the paper!

hexylena (approved):

hexylena:
the build error is an ongoing issue we're solving, not worried about it but it won't be deployed until we get that solved, sorry.

mtekman:
> the build error is an ongoing issue we're solving, not worried about it but it won't be deployed until we get that solved, sorry.

Unrelated to this PR: 
- There is a workshop tomorrow that I was going amend some of the version links for some of the scanpy tools. I take it that this would not be possible? At the moment things just appear to work as they are with the current versions, so it's not at all a pressing issue

hexylena:
It's since fixed, Saskia found a workaround and builds are succeeding again
so just PR any changes you want! 🎉

ons. 17. mar. 2021, 12:45 skrev Mehmet Tekman ***@***.***>:

mtekman:
Fast and awesome as always, thanks @shiltemann @hexylena !

PR #2449, opened 2021-03-18
fix a few errors in speaker notes

annefou:
I just noticed listening to the video that there were a few errors in speaker notes (for text-to-speech video) so this PR will hopefully fix them.


hexylena (approved):
looks good!

PR #2453, opened 2021-03-23
Update tutorial.md

gmauro:
use a more general syntax


hexylena (approved):
thanks for thinking to update this!

Issue #2455, opened 2021-03-24
The "Search Tutorials" function on the webpage doesn't seem to find some tools


AnnaSyme:
Issue: A bug in the search function? or a suggested enhancement to the search function

Detail: On the webpage https://training.galaxyproject.org/training-material/  if I want to search the tutorials for a particular tool (e.g. Blast), there are no results found. However this is a tool that is used in some tutorials (e.g. https://training.galaxyproject.org/archive/2019-06-01/topics/genome-annotation/tutorials/genome-annotation/tutorial.html). 

Suggestion: I don't think this is necessarily a bug as some other tool searches seem to work. But perhaps there's a way to enhance the search function?

hexylena:
Currently it only searches the metadata and not the whole tutorial. I've got a branch where it's opened up to completely full text search but it isn't merged yet. We can definitely resurrect that work and fix this :)

PR #2458, opened 2021-03-24
Add LunrJS for Full Text Search

hexylena:
fixes https://github.com/galaxyproject/training-material/issues/2455 It does do some things better, e.g. searching for prokka is now found within tutorials.

# The Good

New | Old
--- | ---
image | image

Or 'nanopore'

New | Old
--- | ---
image | image

Those hits are scored, so maybe we can convey that to the users to help them know 'gxadmin tutorial isn't representative'. Or maybe just toss anything under 3. (or maybe 3 isn't a good threshold. But I bet we can find counterexamples for every threshold we might filter at.)

```
$ node bin/lunr-search.js nanopore
14.74   topics/metagenomics/tutorials/nanopore-16S-metagenomics/tutorial.md     nanopor
11.25   topics/metagenomics/tutorials/plasmid-metagenomics-nanopore/tutorial.md nanopor
6.423   topics/assembly/tutorials/chloroplast-assembly/tutorial.md      nanopor
5.034   topics/assembly/tutorials/assembly-with-preprocessing/tutorial.md       nanopor
4.192   topics/assembly/tutorials/unicycler-assembly/tutorial.md        nanopor
2.102   topics/admin/tutorials/gxadmin/tutorial.md      nanopor
1.829   topics/contributing/tutorials/create-new-tutorial-content/tutorial.md   nanopor
```

Or for jbrowse

```
$ node bin/lunr-search.js jbrowse
12.902999999999999      topics/visualisation/tutorials/jbrowse/tutorial.md
10.277000000000001      topics/genome-annotation/tutorials/annotation-with-prokka/tutorial.md
6.470000000000001       topics/assembly/tutorials/chloroplast-assembly/tutorial.md
6.191000000000001       topics/variant-analysis/tutorials/microbial-variants/tutorial.md
5.476   topics/genome-annotation/tutorials/annotation-with-maker/tutorial.md
5.476   topics/genome-annotation/tutorials/annotation-with-maker-short/tutorial.md
4.134   topics/variant-analysis/tutorials/tb-variant-analysis/tutorial.md
4.051   topics/sequence-analysis/tutorials/mapping/tutorial.md
2.883   topics/assembly/tutorials/ecoli_comparison/tutorial.md
```
# The Bad

The indexer is much fancier, but, it struggles with terms like 'sars-cov-2' where it finds anything with a '2' in it. 

image

# The Ugly

Partial matches fail outright, unlike the current one.
image

Because you need to use wildcards.

image

We could add these by default? Not sure how much that might screw up some results. You can also use `~` to accept N bad matches? https://lunrjs.com/docs/lunr.Index.html#~queryBuilder I'm not sure if it's too much power for users and we should just force wildcard usage.

image

or searching specific fields:

image

Oh also it requires nodejs in the build process :( so I guess restricted to deployments. I found a ruby plugin, but the ruby plugin did 'ExecJS' to run the lunr code :joy::sob:


shiltemann:
This is cool, could we also have a search box like this specifically for FAQs? I think that was high on @jennaj's wishlist. Would be great if there is a search across all different FAQ folders :)

hexylena:
yeah, it'd be quite easy to build a second index for the FAQs! good idea

hexylena:
discussion with @yvanlebras @bebatut , it would be nice to be able to find "all tutorials for tool X" so we can link from galaxy to help in the GTN.

> Like this idea because allow to have something directly linked to the tool

It sounds fantastic. How specific do we want to search? Just the tool name? Or would that have too many conflicts?

.github/workflows/monthly-release.yml
```diff
@@ -21,6 +21,9 @@ jobs:
       - uses: actions/setup-ruby@v1
         with:
           ruby-version: "2.5"
+      / uses: actions/setup-node@v2
```
hexylena, line 24:
```suggestion
      - uses: actions/setup-node@v2
```

shiltemann (approved):

shiltemann:
Thanks @hexylena!

PR #2469, opened 2021-03-30
[WIP] update ATAC-seq tutorial

lldelisle:
- I updated the screen shots.
- I changed the tool to do the histogram of paired-end fragment size
- I put MACS2 first
- I used MACS2 usual parameters (extension 200bp).
- I used the same for Genrich and -> catastrophe no more peak...

I am wondering if I should remove Genrich.


lldelisle:
I removed Genrich after discussing with @heylf and @hrhotz .
I tried to update everything...
I still did not serve it to check it was correct...

lldelisle:
@heylf @hrhotz Ready for review...

hrhotz (dismissed):
just a few comments for the text: 

line 511: [...] and the more reads you have, the less peaks you get

line 524: This means in order to have the read start site reflecting the centre of where Tn5 bound, [...] in the footprint, [...]

line 527: [...](100bp in each [...]

line 531: [...]set the extension size in MACS2, it will only consider one read of the pair while here we would like to use both the information from both.

hrhotz (approved):

heylf:
@lldelisle everything is fine for me!

PR #2486, opened 2021-04-06
Epigenetics speakernotes update

lldelisle:
Hi,
I filled all the speaker notes.
I have 2 questions:
- At the end I have a slides with "Related tutorial" which is empty. How to fill it?
- Is there a way to preview the video?
Thanks


topics/epigenetics/slides/introduction-atac-seq.html
```diff
@@ -0,0 +1,154 @@
...
+
+## Overview
+
+![](../images/atac-seq/ATACWF.svg)
+
+- This is an overview of ATAC-Seq data analysis.
+
+???
+
+- This is an overview of the workflow which is used in the tutorial.
```
hexylena, line 154: do you want to describe it at all? Or is it sufficient you think?

topics/epigenetics/slides/introduction-atac-seq.html
```diff
@@ -0,0 +1,154 @@
...
+  - What are the quality parameters to check for each dataset?
+  - How to analyse ATAC-Seq data?
+objectives:
+  - Understand ATAC-Seq
+  - Quality Parameters for ATAC-Seq
+  - Understand Peak calling for ATAC-Seq
+requirements:
+time_estimation: "30min"
+contributors:
+  - heylf
```
hexylena, line 22: I think add yourself here as well :)

hexylena (commented):
This file *must* be named introduction.html or it will not get discovered correctly.

Is it not appropriate for a topic-level introduction? I think we need to move it into being a normal slide deck then, rather than a topic level introduction.

hexylena:
> At the end I have a slides with "Related tutorial" which is empty. How to fill it?

This might be a bug on the GTN end, I'll look.

> Is there a way to preview the video?

Yes! after renaming it (either to intro, or a folder like topic/epi../tutorials/atac-intro/slides.html) you can run the following:

```
docker run -d -p 5002:5002 synesthesiam/mozillatts
./bin/ari-quick.sh topics/transcriptomics/tutorials/scrna-intro/slides.html
```

and this will compile them with the MozillaTTS option (it has some known bugs, so if you get really weird audio sometimes, it's normal. When we run it in production with AWS it's a lot better.)

lldelisle:
Hi,
I think there was a misunderstanding. This is not a general introduction to epigenetics for sure.
I need to check with @heylf but I think these are the slides corresponding to the ATAC-seq tutorial... I think his idea was to have multiple ATAC-seq tutorial and this would be the intro to the ATAC-seq tutorialS.

hexylena:
Ahhh ok. Then we can do that another way, by using subtopics in epigenetics!

They look like this: https://training.galaxyproject.org/training-material/topics/admin/ there are different sections as you can see.

They are defined in the topic level metadata https://github.com/galaxyproject/training-material/blob/main/topics/admin/metadata.yaml#L8

And then you just write 'subtopic: <topicid>' in each slide deck/tutorial that should go in the subtopic. I imagine we could have an ATAC-Seq, but I'm not sure about the others. You can also use the keyword 'priority: <0-N>' to specify the ordering within a subtopic.

shiltemann:
@lldelisle ah ok, if they are slides that are meant to go together with the atac-seq tutorial, you can move them to `topics/epigenetics/tutorials/atac-seq/slides.html` ..if you want something more complicated like having a different slide deck in the future to go with the existing tutorial, and have this be a general ATAC-Seq intro, you can just make a new tutorial folder with only slides.html inside it (no hands-on) (and then you can follow Helena's suggestion to group and order them if needed)

(And let us know if you need some help with it)

hexylena (approved):

PR #2489, opened 2021-04-09
Fix Alevin version, link answer key

pinin4fjords:
From @nomadscientist 's [answer key](https://humancellatlas.usegalaxy.eu/u/wendi.bacon.training/h/pre-processing-with-alevin---part-1---answer-key) I can see that Alevin 1.3.0 is the version that should be used, and I'm having trouble with the currently cited version, so I've edited that. I've also linked the answer key for the benefit of students.


nomadscientist (approved):
Thanks for the fix, great catch!

hexylena (approved):

Issue #2472, opened 2021-04-01
Most tours broken currently due to UI changes


hexylena:
I found this issue while searching for [tours I could use in Selenium tests](https://github.com/galaxyproject/galaxy/issues/11413).

The TL;DR version is that all of the tours are broken; 30 of the 33 tours use a CSS .class selector (.fa.fa-upload) that is no longer used on the upload button.  The other three tours failed waiting for #btn-reset to become clickable; note that most of the other 30 would have failed for this reason as well if they hadn't already failed trying to find the upload button.

Having said that, those two problems can likely be fixed with find/grep/awk scripts and I would like to investigate how much effort it would be to get the tours working again.  I am willing to bet that it is less effort than crafting an entire new set of Selenium tests from scratch and the tours provide some nice real world use cases to test against.  Once the tours are part of the selenium test regime it should be much easier to keep them updated as we will know as soon as one of them breaks.

_Originally posted by @ksuderman in https://github.com/galaxyproject/training-material/issues/840#issuecomment-811322997_

hexylena:
@ksuderman moved this into it's own issue. thanks for adding this bug report! We really need more stable UI identifiers for these buttons :(

PR #2490, opened 2021-04-09
[WIP] Updating Tours

ksuderman:
This is a first pass at updating the tours to be in sync with the latest Galaxy UI (21.01) and latest tool versions available on the tool shed.  This PR does not modify any existing tours but adds several scripts that address two of the primary issues. See #2472.

1. Element selectors use CSS styles that are no longer used on the elements they are supposed to select.  For example, the upload button no longer uses the *.fa-fa.upload* style for the icon.  However, most elements have an identifier (ie. HTML id attribute) and the CSS style selectors are not required.
2. Updating tours to use the latest versions of tool  requires finding the latest available version and manually editing the tour YAML file.

The first problem can be solved with simple string search/replace operations.  The second problem is slightly more complicated, and this PR proposes generating [Jinja2](https://jinja.palletsprojects.com/en/2.11.x/) templates with parameterized tool versions to address the difficulties maintaining tours in the future.

1. `fix_tours.py`
   Performs the string replacesment for CSS style selectors and generates a [Jinja2](https://jinja.palletsprojects.com/en/2.11.x/) template with tool ids replaced with variables defined in an external `vars.yaml` file.  Tool ids and revisions are also saved to a `tools.yaml`file that can be used to install the tools.  The `fix_tours.py` script traverses the entire `topics` directory structure looking for `tour.yaml` files and the generated files are written to a separate location so they can be easily collected during development and testing.  In production the generated files would likely be written to the same directory as the original  `tour.yaml` file.
2. `update_tour.py`
   Updates the `tools.yaml` and `vars.yaml` files with the latest verions available from the tool shed.
3. `render_tour.py`
   Generates a `tour.yaml` file from the Jinja template and `vars.yaml` file.

# Usage

### fix_tours.py

The `fix_tours.py` script takes one parameters, the name of the output directory:

```
mkdir /tmp/tours
python bin/fix_tours.py /tmp/tours
```

Each tour will be written to its own directory with the tour id as the directory name.  For example:

```
mapping
├── tools.yaml
├── tour.yaml.j2
└── vars.yaml
```

### update_tour.py

The `update_tour.py` script takes one parameter; the path to the directory containing the tour files generated by `fix_tours.py` script and updates the  versions in the `tools.yaml` and `vars.yaml` files with the latest tool shed versions.  The existing files are overwritten.

```
python bin/update_tour.py /tmp/tours/mapping
```

### render_tour.py

The `render_tour.py` script takes one parameter; the path to the directory containing the tour files generated by `fix_tours.py` script and generates the `tour.yaml` file from the Jinja template and `vars.yaml` file.

```
python bin/render_tour.py /tmp/tours/mapping
```

### Utility Scripts

1. `bin/generate_tours.sh` a Bash script that runs all three of the above Python scripts to process all existing tour files.
2. `bin/copy_tours.sh` copies the generated tours to a local Galaxy instance, or any other directory.

# Next Steps

1. Identify remaining issues that can be fixed automatically in the `fix_tours.py` script.
2. Integrate tours as Selenium tests

# Caveats and Known Issues

Only a few of the generated tours have been tested, but several issues have been identified.

1. The process for creating new histories has changed slightly (there seems to be one less thing to click).  The fix can likely be automated and won't be an issue for tours generated after the history panel change.
2. Tool searches are not working. The text is being entered in the search field but the search doesn't take place. This may be another incorrect element selector.
3. The tour dialog frequently covers the UI element it is describing. 
4. It is not known if updating a tool version breaks a tour due to changes in parameters in the tool's XML configuration that may change the UI.

# Possible Future Work

1. It may be desireable to parameterize the element selectors into the `vars.yaml` files as well.
2. Modify the tour file generator browser extensions so they do not include CSS style selectors if an element contains an #id.
3. Have the tour generator generate the Jinja template, and other files directly
4. Modify Galaxy to load the tour from the Jinja template rather than preprocessing them before installing into Galaxy.

### DIFF
--- a/bin/fix_spaces.py
+++ b/bin/fix_spaces.py
@@ -1,0 +1,22 @@
+import os
+import sys
+
+'''
+At least one tour (iwtomics) uses incorrect indentation that causes yamllint
+to complain.
+'''
+
+def fix(path):
+    with open(path) as f:
+        for line in f.readlines():
+            if len(line) > 4 and line.startswith('  '):
+                print(line[2:], end='')
+            else:
+                print(line, end='')
+
+
+if __name__ == '__main__':
+    if len(sys.argv) != 2:
+        print('No filename provided.')
+        sys.exit(1)
+    fix(sys.argv[1])
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ vendor
 .jekyll-metadata
 .jekyll-cache
 plantuml.jar
+.idea
+.venv
--- a/bin/copy_tours.sh
+++ b/bin/copy_tours.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+set -eu
+
+# Copy tour files to a local Galaxy installation. Tours will be renamed using
+# the name of parent directory.
+
+if [[ $# != 2 ]] ; then
+  echo "USAGE: $0 /tours/directory /galaxy/tours/directory"
+  exit 1
+fi
+
+for tour in $(find $1 -name tour.yaml) ; do
+  dir=$(dirname $tour)
+  tour_name=$(basename $dir)
+  echo "cp $tour $2/$tour_name.yaml"
+  cp $tour $2/$tour_name.yaml
+done
--- a/bin/fix_tours.py
+++ b/bin/fix_tours.py
@@ -9,29 +9,30 @@ A program to update outdated Galaxy Tours to conform to the latest version
 (21.01 as of this writing) of the Galaxy UI.
 
 This program generates a Jinja2 template and externalizes tool IDs to a 
-separate vars.yaml file so tool versions can be updated automatically (see the
-update-tour.py program) without editing the tour file itself.
+separate vars.yaml file so tool versions can be updated automatically without 
+editing the tour file itself. See the update_tour.py program for details on
+updating tours. 
 
 A list of the tools used in the tour are  written to a tools.yaml file so 
 admins can easily see and install the tools needed by a particular tour. The
-tools.yaml file is also updated by the update-tour.py program.
+tools.yaml file is also updated by the update_tour.py program.
 
-Use the render-tour.py program to generate the tour YAML file that will be 
+Use the render_tour.py program to generate the tour YAML file that will be 
 loaded by Galaxy.
 
-NOTE: Since YAML is a PITA to roundtrip cleanly and easily we don't use a YAML
-parser to read the original tour.yaml file, but simple read it line by line and
-do string replacements.  The problem is Block Literals, i.e. multi-line strings
-specified with >- or |-. Once those strings are stored in dictionaries the style
-of input is lost and the YAML processor will pick what it thinks is best when
-it outputs the string.
+NOTE: Since YAML is a PITA to round trip cleanly and easily we don't use a YAML
+parser to read the original tour.yaml file. Instead we simply read it line by 
+line and do string replacements.  The problem is Block Literals, i.e. multi-line 
+strings specified with >- or |-. Once those strings are stored in dictionaries 
+the style of input is lost and the YAML processor will pick what it thinks is 
+best, which is almost always not what we want.
 '''
 
 DEFAULT_TOOLSHED = 'https://toolshed.g2.bx.psu.edu'
 shed = toolshed.ToolShedInstance(DEFAULT_TOOLSHED)
 
 # The strings that we do simple search/replace on.
-# TODO These values should be parameterized.
+# TODO These values should be parameterized, or at least loaded from a config.
 replacements = [
     (' .fa.fa-upload', ''),
     ('#history-options-button', '#history-new-button'),
@@ -86,9 +87,11 @@ def parse_tour(tour_file, directory):
         for line in f.readlines():
             line = do_replacement(line)
             if line.startswith("steps:"):
+                # Add tags immediately before the steps element.  None of the
+                # existing tours have tags so this is safe.
                 lines.append("tags:\n")
-                lines.append(f'  - "{ tag }"\n')
-                lines.append('  - auto\n')
+                # lines.append(f'  - "{ tag }"\n')
+                lines.append('  - Auto\n')
                 lines.append(line)
             elif line.find("/tool_runner?tool_id=") > 0:
                 start = line.index("tool_id=") + 8
--- a/bin/fix_tours.py
+++ b/bin/fix_tours.py
@@ -1,0 +1,142 @@
+import os
+import sys
+import yaml
+from bioblend import toolshed
+from urllib.parse import unquote as decode
+
+'''
+A program to update outdated Galaxy Tours to conform to the latest version 
+(21.01 as of this writing) of the Galaxy UI.
+
+This program generates a Jinja2 template and externalizes tool IDs to a 
+separate vars.yaml file so tool versions can be updated automatically (see the
+update-tour.py program) without editing the tour file itself.
+
+A list of the tools used in the tour are  written to a tools.yaml file so 
+admins can easily see and install the tools needed by a particular tour. The
+tools.yaml file is also updated by the update-tour.py program.
+
+Use the render-tour.py program to generate the tour YAML file that will be 
+loaded by Galaxy.
+
+NOTE: Since YAML is a PITA to roundtrip cleanly and easily we don't use a YAML
+parser to read the original tour.yaml file, but simple read it line by line and
+do string replacements.  The problem is Block Literals, i.e. multi-line strings
+specified with >- or |-. Once those strings are stored in dictionaries the style
+of input is lost and the YAML processor will pick what it thinks is best when
+it outputs the string.
+'''
+
+DEFAULT_TOOLSHED = 'https://toolshed.g2.bx.psu.edu'
+shed = toolshed.ToolShedInstance(DEFAULT_TOOLSHED)
+
+# The strings that we do simple search/replace on.
+# TODO These values should be parameterized.
+replacements = [
+    (' .fa.fa-upload', ''),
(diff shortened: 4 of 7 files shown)


bin/fix_spaces.py
```diff
@@ -0,0 +1,22 @@
+import os
+import sys
+
+'''
+At least one tour (iwtomics) uses incorrect indentation that causes yamllint
```
hexylena, line 5: should we just yamllint the tours then? Surely would be better than needing an extra script

ksuderman: Yes, this is script is not one of the "deliverables" and will likely be removed in the future.  Yamllint was complaining about spaces and I wanted to see if this made it happy (it does).  A better solution would be to fix the spacing in the original tutorial.

ksuderman:
My motivation for updating the tours is to explore using them in front-end Selenium tests based on real world usage of Galaxy.  During the 21.01 release testing it struck me that many very smart people spent a great deal of time clicking buttons that could be automated by Selenium if we had an easy way to drive the tests. Since Galaxy already has the infrastructure in place to use tours to drive Selenium they seemed like an obvious starting point.

I will leave it up to the training group to decide if the ROI is worth it to maintain the tours going forward, but I wanted to put this out there for comments.  If there is no interest in maintaining the tours then they are not good candidates for Selenium testing and I won't put any more effort into this.  But, if we do decide to maintain the tours having them as part of the nightly test regime should make them easier to maintain in the future; or at least we will know as soon as one of them breaks.

hexylena:
Using them for testing sounds fantastic! I love the idea. In practice I think they may have clicked on fewer buttons than you think :(

```
ag element: --yaml --nofilename | sort | uniq -c | sort -n | sed 's/^\s*//g;s/\s\+/ | /'
```

Count | element
--- | ---
4 | element: "#current-history-panel .list-item:eq(0)"
4 | element: "#current-history-panel .list-item:eq(1)"
4 | element: "#current-history-panel .list-item:eq(2)"
4 | element: 'div[tour_id="files_grn"]'
4 | element: 'div[tour_id="files_red"]'
4 | element: "#execute"
4 | element: '#workflow .dropdown a[href$="/workflow"]'
5 | element: "#tool-search-query"
20 | element: >-
40 | element: '#center-panel'
45 | element: '#btn-new'
47 | element: '#btn-start'
48 | element: '#btn-close'
50 | element: .upload-text-column .upload-text .upload-text-content.form-control
53 | element: '#tool-panel-upload-button .fa.fa-upload'
65 | element: '#history-options-button'
195 | element: '.history-right-panel .list-items > *:first'
407 | element: '#tool-search-query'
817 | element: '#tool-search'

Which translates into:

1. a huge amount of searching for tools (promptly ignored because they just go directly to it via url.) 
2. creating a new history (options button, without a subsequent 'create new')
3. uploads (#btn-{new,start,close}, uploads), 
4. prompts for tool params ('#center-panel')

Some exceptions:

- iwtomics' tutorial does the most interesting things (mostly pointing out features of specific uploaded files.) seemingly walking through a tool? 
- mothur-miseq looks into the library and says 'import files' but doesn't actually enter said library or click through the steps to import files?)

So I'm starting to think this may be a less interesting dataset than you hoped. 

> During the 21.01 release testing it struck me that many very smart people spent a great deal of time clicking buttons that could be automated by Selenium if we had an easy way to drive the tests. Since Galaxy already has the infrastructure in place to use tours to drive Selenium they seemed like an obvious starting point.

like, absolutely, :100:, fantastic idea. And the idea of being able to easily turn those into videos someday makes me go :heart_eyes::heart_eyes::heart_eyes:, but it seems most tours don't do the hard work of entering tool inputs, clicking execute, or really automate much other than a very quick intro "and now do the rest of it yourself" which makes me think this may not be the best dataset

hexylena:
So with all of this depressing stuff said, in theory it's still good. Maybe there's a way we can rethink it though. 

Could we (in theory) auto-generate the tours from the tutorials with workflows? If so that would work, but it might not be 'interesting' to test. 

I'd love to be able to write a tour that could fill out the entire tool form but the IDs are basically useless. it's a pity they don't correspond to the parameter name hierarchy (maybe like in tests with `repeat|value`), otherwise we could walk through the entire tool form and it might be a more interesting tour from the testing standpoint?

ksuderman:
I am sure we can salvage *something*...

Of the four issues you mention (tool search, creating new histories, uploads, and tool parameters) I believe the uploads are mostly fixed and working, at least on the few patched tours I tested. The process for creating new histories has changed slightly, but it should be cookie-cutter fixes for every tour; once we figure out how to fix one we've fixed them all.  Same for the tool search.  The other problem I can foresee is selecting items from the history panel if the selector uses a dataset id as those will be different every time.  

Filling out the tool parameters will be the most problematic and the most likely show stopper.  Using the workflows to generate/supplement the tours is an interesting idea.  I don't think the workflows contain enough information to generate a complete tour, but it may be possible to pull out the tool parameters.

Since there seems to be some interest in at least continuing the investigation I propose we:

- [ ] Identify the tours that are likely to be the most *interesting* in terms of Selenium tests.  A tour that uses several different tools and interacts with more UI elements is more *interesting* than a tour that runs the same tool multiple times with different parameters.
- [ ] Determine the steps required to fill in tool parameters.
- [ ] Investigate extracting tool parameters from workflows. Can a tour be generated from a workflow?

Since you are much more familiar with the tutorials (and therefore the tours) if you could help with the first item I will investigate the other two.

hexylena:
> most likely show stopper

damn, yeah, just as I feared.

> I don't think the workflows contain enough information to generate a complete tour, but it may be possible to pull out the tool parameters

yeah, was thinking long term, between WFs with tests/inputs + tutorial maybe we could stitch something together, but not a short term plan!

>  Identify the tours that are likely to be the most interesting in terms of Selenium tests. A tour that uses several different tools and interacts with more UI elements is more interesting than a tour that runs the same tool multiple times with different parameters.

So I think this is where I fear it dies, that none of the tours do anything interesting. The vast majority are just upload, search, browse to tool, popup. I'm not sure we find enough tours that do interesting things to justify the time.

Most really look like:

```
element: '#history-options-button'
element: '#tool-panel-upload-button .fa.fa-upload'
element: '#btn-new'
element: .upload-text-column .upload-text .upload-text-content.form-control
element: '#btn-start'
element: '#btn-close'
element: '#tool-search-query'
element: '#tool-search'
element: '#tool-search'
element: '#tool-search-query'
element: '#tool-search'
element: '#tool-search'
element: '#tool-search-query'
element: '#tool-search'
element: '#tool-search'
element: '#history-options-button'
element: '#tool-search-query'
element: '#tool-search'
element: '#tool-search'
element: '#history-options-button'
element: '#tool-search-query'
element: '#tool-search'
element: '#tool-search'
element: '#history-options-button'
element: '#tool-search-query'
element: '#tool-search'
element: '#tool-search'
element: '#history-options-button'
element: '#tool-search-query'
element: '#tool-search'
element: '#tool-search'
element: '#history-options-button'
```

The most interesting things I've found are:
- 1 instance of going to /workflow (but not importing the WF, just text telling you you can import it from this URL). The same tour revisits /workflow to tell you to re-run with changed params, but doesn't click anything for you.
- 2 that visit /library/list, but does nothing before return to main page and running tools.
- 1 that clicks on uid-NNs which are now broken, and the execute button (without entering parameters, just telling you to.)
- 1 that initiates trackster but doesn't go any further

The post-click actions are no more interesting, just tools. So this, this doesn't seem like it's interesting enough for you, right? No workflows uploaded or run, no libraries searched, just the aforementioned uploads + search.

> Determine the steps required to fill in tool parameters. 

I think this was an old conversation too, the lack of stable IDs. (But it still requires figuring out the IDs, so unless the IDs (or selectors) come from the parameter names, we may need to stop here.)

>  Investigate extracting tool parameters from workflows. Can a tour be generated from a workflow?

I'll tentatively say yes here, but low priority.

ksuderman:
> The post-click actions are no more interesting, just tools. So this, this doesn't seem like it's interesting enough for you, right?

That is probably interesting enough to get started. It seems most real world Galaxy usage (from my limited experience) is `upload, (search, execute)+` so an interesting Selenium test only needs to do the same.  Probably the Interesting Metric, from my point of view, is something like `distinct_tools_used  / number_of_steps_in_tour`. 

>> most likely show stopper

> damn, yeah, just as I feared.

There may still be hope.

I had an opportunity to delve into the DOM generated for the tool forms and all of the input fields are wrapped in `<div>` elements with a tantalizingly named attribute `tour_id` with values that seem to correspond to the input parameter names specified in the wrapper XML file.  It would be nice if the `tour_id` attribute was on the actual `<input/>` element and not some `<div>` five layers above, but at least it is there.

My next step will be to see if I can come up with a CSS selector to locate the `<input/>` element and manipulate it (i.e. click, enter text, etc).

hexylena:
Oh, wow, yeah, never noticed the tour_id, that definitely looks nearly perfect. With a selector like `div[tour_id="ideogram|chromosomes_reverse"] input` we can select what we need for most input types then. (select2s and file inputs I guess are still difficult.)

looks like these are the top by your metric:

```
.15942028985507246376	./topics/epigenetics/tutorials/methylation-seq/tours/tour.yaml
.16475095785440613026	./topics/metagenomics/tutorials/mothur-miseq-sop/tours/tour.yaml
.16666666666666666666	./topics/metagenomics/tutorials/plasmid-metagenomics-nanopore/tours/tour.yaml
```

I'm doing tangentially related work for the admin training where we're recording clicking around the interface (using https://playwright.dev, not using any of the testing features, I just wanted something that'd give me an mp4 easily.) https://youtu.be/c6oYsA6p_mw?t=851 where we show a tool and click on the ref data selector to show the results and it's currently incredibly fragile because I didn't know about the tour_ids, those'll help brilliantly.

PR #2492, opened 2021-04-12
Update ATAC-seq tutorial: use new way to specify tool

lldelisle:
Hi,
I used the new syntax for tool which is super useful when you want to use it through the galaxy integration. Thanks for that.
I just have an issue, the {{page.topic_name}} is not correctly set when I serve it locally:
image

Do you have an idea how to fix it?
Thanks


lldelisle:
Please note that I modified `bin/check-broken-tool-links.sh`. Would you mind to check I was right?

shiltemann:
awesome @lldelisle, thanks for doing it! The changes to the linting script look good :+1: I will check the missing topic name, but 
pretty sure that is  broken when using the incremental build so nothing to worry about

shiltemann (approved):

PR #2503, opened 2021-04-16
correct image locations in plant RNA-seq slides

pavanvidem:
sorry, the image locations in the previous PR were wrong.


hexylena (approved):
you make it in under the wire hah. I should have caught this on first test, sorry. Our testing for slides is just not as good unfortunately

pavanvidem:
:-) thanks! it should have been fixed.

PR #2512, opened 2021-04-21
group tags tutorial improvements

bernt-matthias:
- there seems no way to define a URL colum
- be a bit more precise on how to add multiple tag columns

### DIFF
--- a/topics/galaxy-interface/tutorials/group-tags/tutorial.md
+++ b/topics/galaxy-interface/tutorials/group-tags/tutorial.md
@@ -32,8 +32,11 @@ Grouping datasets in this way has numerous advantages:
 
 While collections can be split in any way, doing so for multi-factor analysis
 quickly becomes cumbersome and messy. An alternative is to label collection
-elements with special group tags. These tags can be displayed in the Tool form,
-allowing users to select subsets of collections.
+elements with special group tags, i.e. tags prefixed by the string `group:`.
+Note that group tags currently do not propagate, i.e. they are not inherited
+to datasets resulting from analyses. These tags can be displayed in the Tool
+form, allowing users to select subsets
+of collections.
 
 This tutorial outlines how to set and use group tags with the DESeq2 tool.
 For a more detailed description and background for differential expression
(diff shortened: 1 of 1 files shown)


topics/galaxy-interface/tutorials/group-tags/tutorial.md
```diff
@@ -99,9 +99,8 @@ The second method works at any step during the analysis.
 >
 >    - Click on **Rules** and then **Add / Modify Column Definitions**
 >    - Click on **Add Definition** and select:
->      - *"URL"*: Column A
```
hexylena, line 102: I see the option to make A the URL (it should be absolutely necessary, otherwise how would the data get imported?) I don't think we should remove that.

image

hexylena (commented):
Looks ok other than removal of url.

topics/galaxy-interface/tutorials/group-tags/tutorial.md
bernt-matthias: True. Seems that I was "abusing" the tutorial and just applied the `Apply rule to collection` tool. But this is also suggested in the intro:  _There are three ways to set group tags: Using the rule builder / apply_rules tool ..._ Maybe we should re-formulate such that its clear that the URL option is unavailable to the `Apply rule` tool.

But in the `Apply rule` tool quite a lot of options are missing (certainly URL seems not needed in this context). I remember that there was an issue but can't find it. 

Given that group tags do not propagate at the moment (https://github.com/galaxyproject/galaxy/issues/8554) I find the `Apply rule to collection` more applicable in real world scenarios.

hexylena: Ahhh that makes sense then. Yes, we definitely need to clarify that.

And, agreed @ lack of propagation meaning doing it in upload is often less useful. We should update the tutorial to include a more relevant example.

hexylena: I've made a change to the intro to clarify. Are you ok with that change?

topics/galaxy-interface/tutorials/group-tags/tutorial.md
```diff
@@ -51,13 +54,16 @@ testing see the [Reference-based RNA-Seq data analysis]({{ site.baseurl }}/topic
 
 # Setting group tags using the apply rules tool
 
-There are three ways to set group tags:
-  - Using the rule builder / apply_rules tool
+There are several ways to set group tags:
+  - Using the Rule Based Uploader
   - Using the "Tag elements from file" tool
+  - Using the "Apply Rules" tool
   - Manually adding dataset tags with the prefix `group:`
```
bernt-matthias, line 61: Actually there is one more: from the history one can `For all selected ...` `->` `Build collection from rule`.

Maybe it would be good to reformulate the tutorial such that the upload, apply rules, build collection paths are covered. The only difference seems to be that the `URL` column needs to (can be) be  defined only in the upload option.

hexylena: Ah forgot about that option. We really should cover each, you're right, that sounds like a good improvement! You're probably more familiar with group tags, would you want to open a 'help wanted' issue for updating the tutorial for the other use-cases that are important in your view?

hexylena (approved):

topics/galaxy-interface/tutorials/group-tags/tutorial.md
hexylena: Moved to #2527

PR #2513, opened 2021-04-21
Fix to 10x pre-processing workflow

mtekman:
Change to previous version until newer version can be fixed

### DIFF
--- a/topics/transcriptomics/tutorials/scrna-preprocessing-tenx/tutorial.md
+++ b/topics/transcriptomics/tutorials/scrna-preprocessing-tenx/tutorial.md
@@ -249,12 +249,12 @@ We will now proceed to demultiplex, map, and quantify both sets of reads using t
 
 > ### {% icon comment %} Comment
 >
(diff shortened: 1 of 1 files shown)


gallardoalba (approved):

mvdbeek:
Y'all, this is problematic. As an outside contributor, I have no idea what "Change to previous version until newer version can be fixed" refers to and how I can be sure it is safe to update, and for users following the training outside of the GTN overlay this is totally not clear, as the version numbers are not displayed. Can you please create a issue somewhere and link it to this PR and include a comment in the training for why it is necessary to use an older version of the tool.

mtekman:
Yep, it was referenced before in an issue -- but sadly I did not get around to fixing it before the workshop

https://github.com/galaxyproject/tools-iuc/issues/3540

hexylena:
> the version numbers are not displayed

they are on hover. It was a discussion over how much should the version number be highlighted, and the conclusion was to not emphasize them too much. Some maintainers were worried about clutter and distraction during teaching so, only on hover. If you wish to revisit that we can discuss again?

also, gentle suggestion to avoid problematic here due to the connotations, unless that's what you intended.

PR #2522, opened 2021-04-29
Update metadata.yaml

gmauro:
The only tutorial of this section is about data downloading and deleting, as also stated by the description
This title helps better to figure out what is it talking about.


hexylena (approved):

hexylena:
thanks, good point!

PR #2530, opened 2021-05-10
SARS-CoV-2 variant analysis: Ensure that only metagenomic data is analysed

pvanheus:
The current version of the tutorial does not distinguish between amplicon and metagenomic sequencing and its techniques are only valid for metagenomic sequencing. This update adds explanations and updates the search instructions.


nekrut (approved):

pvanheus:
> We also do need a corresponding tutorial for ARTIC data based on @wm75 workflows.

I'm working on something like that but I have simpler workflows than @wm75  - we've been discussing this (different workflows for ARTIC) over on Galaxy PH and in emails.

PR #2535, opened 2021-05-17
Add Github action to optimize all images

hexylena:
Copied from https://github.com/galaxyproject/galaxy-hub/pull/733, thanks @davelopez!

This should make shallow clones faster in the future.


martenson (approved):

martenson:
Seems good, we should watch it for a bit since it has many triggers.

davelopez:
@martenson Sorry I didn't notice earlier the repo name in one of the conditions needs to match this repo. I made a PR with the change #2539.
In case you want to try it, you can trigger it manually. If there are many images in the repo, the first PR will contain all of them, but after that, the bot will create a PR only if there are new unoptimized images.

PR #2558, opened 2021-05-25
Add Abbreviation plugin

hexylena:
Referencing abbreviations is really kind of annoying to do, and I'm so used to LaTeX's abbreviation abilities which are fantastic and so easy to use. So I've re-implemented something similar here. Note that on first use the term is expanded and include `term (abbr)` and on subsequent uses is just abbreviated with the abbreviation and the abbr tag which provides hover information. 

After reviewing @rikeshi's new tutorial I noticed that terms like HDA/HDCA/LDA are defined and referenced and it made sense to implement a plugin to help make this process easier. I've done this with @rikeshi's tutorial to start with as a demonstrator. cc @simonbray 

## Syntax

The `/jobs` &lbrace;API&rbrace; will return &lbrace;JSON&rbrace;. When we call the &lbrace;API&rbrace; we'll get back this result &lbrace;JSON&rbrace;

## Output

image

## Built In Glossary

image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -77,6 +77,12 @@ figurify:
   skip_titles:
     - Example of an image with a caption
 
+abbreviate:
+  skip_layouts:
+    - introduction_slides
+    - tutorial_slides
+    - base_slides
+
 # Scholar
 scholar:
   source: topics/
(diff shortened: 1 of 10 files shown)


abretaud (approved):
Cool :+1: !

hexylena:
Merging this since @abretaud reviewed :)

PR #2563, opened 2021-05-26
2021 Backend Dev Architecture Update

jmchilton:
- Cover ASGI, FastAPI, Starlette, OpenAPI
- Talk about new controller types. More controller examples.
- Reframe WSGI in terms of a fallback.
- Update image build process to allow either curl or wget and not just require wget.
- Update various file lists for recent client and CI changes.


hexylena (approved):
Looks good!

PR #2568, opened 2021-05-28
Workflow automation tutorial

simonbray:


hexylena:
I'm not sure where this belongs, to be honest. I know it uses variant analysis but since it doesn't cover science really, just the dev portion, I'd lean to this belonging more in the dev section?

bgruening:
Maybe into "Using Galaxy and Managing your Data"

hexylena:
yeah better suggestion @bgruening!

simonbray:
Yes, I was also unsure about the section. I don't think it really fits in the dev section, it's definitely targeted at users rather than devs.

"Using Galaxy and Managing your Data" is a good idea, thanks @bgruening.

shiltemann:
@simonbray is this still WIP or can we merge it?

wm75:
Still WIP @shiltemann. Going to add a section or so this afternoon/tonight.

shiltemann:
ok great!

wm75 (approved):
It's a challenging tutorial, especially the second part, and I upgraded its level to "Advanced" to reflect that fact.
I think (with my biased perspective) that it's explaining things well though and that interested people can learn really interesting stuff by following it.

Thanks @simonbray!

hexylena (approved):
This looks fantastic y'all! Very impressed. Really nice to have such concrete examples

I removed the `$` from your console lines because people *will* try and copy and paste the CLIs and might as well make it less error prone/painful to do.

PR #2579, opened 2021-06-02
Updated version of RabbitMQ to install to 3.8.16 to make it compatibl…

Slugger70:
…e with new Erlang


hexylena:
Cherry picked onto the gat-fixes branch so it gets included in git-gat

PR #2582, opened 2021-06-02
switch installed and tested tools back to bwa and bamtools_filter

cat-bro:
@hexylena I was talking about changing the order of steps for this to address a feedback comment but I don't think that would be an improvement so am leaving the order as is.  I'll probably make another PR for this topic (updated tips, slides) but none of the code blocks will need to change.


hexylena (approved):
Works for me @cat-bro. It's no trouble if they need to change, and fine for me if the order doesn't change.

PR #2590, opened 2021-06-03
add vgp training skeleton from workflow (hidden)

astrovsky01:
Add hidden training for vgp assembly skeleton

PR #2593, opened 2021-06-04
Fix build failure from vgp

hexylena:
Missed this unexpected broken symlink in the vgp pr. @bgruening https://github.com/galaxyproject/training-material/pull/2590/files#diff-54a870e36cadec0f44bac0ff7df874adef7f9cdb0f8cb2ca5a426618cc0ba8f7

### DIFF
--- a/topics/metadata/assembly.yaml
+++ b/topics/metadata/assembly.yaml
@@ -1,1 +1,0 @@
-../topics/assembly/metadata.yaml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,16 @@ jobs:
         run: |
           python bin/mergeyaml.py --nondocker > /dev/null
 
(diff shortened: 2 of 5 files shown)


topics/metadata/assembly.yaml
```diff
@@ -1 +0,0 @@
-../topics/assembly/metadata.yaml
```
bgruening, line 1: Oh, was that deprecated recently? We need to remove it from the skeleton generator in that case.

Thanks for fixing.

bgruening (approved):
Nice, github check!

topics/metadata/assembly.yaml
hexylena: It wasn't removed, it seems like an accidental symlink was added. This was added to a topic called 'metadata' which doesn't exist.

PR #2595, opened 2021-06-04
Apollo slides

abretaud:
A few slides to introduce the new Apollo tutorial.
Also added links from other places where we talked about Apollo


hexylena (approved):
These slides are fantastic, the speaker notes are great, I love that we get the video of this 'for free'.

Also, the image alt-text is absolutely perfect, thank you for taking the time to write such good alt-text! :heart:

PR #2596, opened 2021-06-04
Update VGP tutorial: modify cutadapt options and version

gallardoalba:
This PR updates the cutadapt section with the most recent version and includes some additional parameters in accordance with Giulio's suggestions.

Screenshot from 2021-06-04 16-39-14


bgruening:
@Delphine-L you also need to update the workflow accordingly (not here, but your real-world one).

bgruening (approved):

PR #2598, opened 2021-06-04
master renamed to main in images

telukir:
* Changing of name Master branch to main
* Left the name master as strike to indicate that it got changed


hexylena (approved):
This looks fantastic, I thought we caught all of those during the rename but clearly not. Thank you for finding and fixing these!

PR #2611, opened 2021-06-08
Install minimap2 in singularity tutorial tests

mvdbeek:
and use `set -e` when running multiple commands in test script.


hexylena (approved):

hexylena:
I was too enthusiastic.

```
        * Error installing a repository (after 0:00:00.000012 seconds)! Name: minimap2,owner: iuc, revision: , error: "Either tool_panel_section_id or tool_panel_section_name must be defined for tool 'minimap2'."
```

PR #2614, opened 2021-06-09
New tutorial: Debugging Galaxy

ic4f:
Apologies for the numerous typos - will fix.
Any feedback very much appreciated!

TODO:
- [x] Address https://github.com/galaxyproject/training-material/pull/2614#discussion_r648442904 (point to specific link, see discusssion)
- [x] Address https://github.com/galaxyproject/training-material/pull/2614#discussion_r648445717 (yarn)
- [x] Address https://github.com/galaxyproject/training-material/pull/2614#discussion_r648462320 (explicit paths)
- [x] Address https://github.com/galaxyproject/training-material/pull/2614#discussion_r648458532 (selenium test issue)
- [x] replace todos in key points with key points

### DIFF
--- a/topics/dev/tutorials/debugging/tutorial.md
+++ b/topics/dev/tutorials/debugging/tutorial.md
@@ -1,293 +1,477 @@
 ---
 layout: tutorial_hands_on
 
-title: "Debugging with Galaxy"
+title: "Debugging Galaxy"
 questions:
   - "How do I debug Galaxy?"
 objectives:
   - "Fix a broken branch"
   - "Interpret the results of failed tests on GitHub"
   - "Run individual tests locally"
-  - "Fix simple errors identified by failing tests"
-  - "Write a simple test exposing the identified bug"
+  - "Fix errors identified by failing tests"
+  - "Debug simple runtime errors using the Python debugger"
+  - "Write tests exposing the identified bug"
 time_estimation: "2h"
 key_points:
-  - "Unit Tests"
-  - "API or Integration Tests"
-  - "Client Linting Failure"
-  - "Client Tests"
-  - "Selenium Tests"
-  - "Runtime Error"
+  - "TODO Unit Tests"
+  - "TODO API Tests"
+  - "TODO Client Linting Failures"
+  - "TODO Client Tests"
+  - "TODO Selenium Tests"
+  - "TODO Runtime Errors"
 contributors:
   - assuntad23
   - ic4f
   - jmchilton
 requirements:
-  - type: "none"
+  - type: none
     title: "Familiarity with basic Git commands"
-  - type: "none"
-    title: "Basic knowledge of Python and Javascript"
-  - type: "none"
+  - type: none
+    title: "Basic knowledge of Python and JavaScript"
+  - type: none
     title: "Mac OS or Linux that can run Galaxy & your favorite IDE or editor"
   - type: "internal"
     topic_name: dev
     tutorials:
       - architecture
+subtopic: core
 ---
 
-## Introduction
+# Introduction
+{:.no_toc}
 
-In this tutorial we are going to demonstrate how to find common types of bugs or errors that you may encounter as a contributor to Galaxy.We aim to step you through the process of finding and fixing a bug - from identifying them on GitHub, to finding and developing a solution, to committing your code change.
 
-With the skills from this tutorial, it is our hope that you will feel more prepared to develop solutions for Galaxy and more confidently navigate the obstacles along the way.
+In this tutorial we will demonstrate how to find and fix common types of bugs you may encounter as a contributor to Galaxy. We will step you through the process of finding and fixing a bug - from locating specific errors in the logs of Galaxy's GitHub workflows, to identifying their cause, developing a solution and committing your edits.
 
-## Some Basics
+With the skills from this tutorial, it is our hope that you will feel more prepared to develop solutions for Galaxy and more confidently navigate any obstacles along the way.
+
+> ### Agenda
+>
+> 1. TOC
+> {:toc}
+>
+{: .agenda}
+
+# Local development environment setup
 
 For almost every type of error you encounter, there will be a relatively similar process:
 
-1. View failing test/error on GitHub workflows
+1. Locate failing test or error on GitHub workflows
 2. Run failing test locally
-3. Search for issue causing failure
-4. Analyze the code, develop and implement a solution
+3. Identify issue causing failure
+4. Develop a solution
 5. Ensure test passes locally
-6. Push fixed branch to local fork
+6. Commit and push fixed branch to local fork
 
-Note: This process will not be the same for the runtime error, since there will be no failed test to review initially.
+The process for dealing with runtime errors will be somewhat different, since there will be no failed test to review initially. However, in either case, the first step is to setup your local development environment.
 
-## Getting Started: Pulling the Buggy Fork
+## Contributing to Galaxy
 
-Good news! All of the failing tests are on the same fork of Galaxy! That means you only need to download and work off of one branch. The downside is that this is one buggy branch with issues all over the place.
+To contribute to galaxy, a GitHub account is required. Changes are proposed via a [pull request](https://docs.github.com/en/github/collaborating-with-pull-requests). This allows the project maintainers to review the changes and suggest improvements.
 
-We have no doubts that you can solve them though!
+The general steps are as follows:
+1. Fork the Galaxy repository
+2. Clone your fork
+3. Make changes in a new branch
+4. Commit your changes, push branch to your fork
+5. Open a pull request for this branch in the upstream Galaxy repository
 
-The fork is located at : TODO
+In this tutorial, instead of cloning Galaxy's default branch, we will clone a branch that has been created for this tutorial and contains several bugs that you'll need to identify and fix. Also, we won't be opening a pull request in the upstream Galaxy repository.
 
-### Directions for Cloning/Basing off this fork
+## Getting started: Pulling the buggy branch
 
-TODO
+Good news! All of the failing tests are on the same Galaxy branch! That means you only need to download and work off of one branch. The downside is that this is one buggy branch with issues all over the place. We have no doubts that you can solve them though!
 
-> ### {% icon comment %} Galaxy Repository
->
-> Note: This tutorial uses a fork to house the issues created for the purposes of this tutorial. In the future, when you contribute to the Galaxy project, you'll want to use one of the repositories located at https://github.com/galaxyproject. Most likely you'll be using the main repo, simply titled "galaxy".
-{: .comment}
+> ### {% icon hands_on %} Hands-on: Setup your local Galaxy instance
+> 
+> 1. Use GitHub UI to fork Galaxy's repository at `galaxyproject/galaxy`. 
+> 2. Clone your forked repository to a local path, further referred to as GALAXY_ROOT and `cd` into GALAXY_ROOT. Note that we specify the tutorial branch with the `-b` option:
+> ```shell
+> $ git clone -b training_dev_debugging https://github.com/galaxyproject/galaxy GALAXY_ROOT
+> $ cd GALAXY_ROOT
+> ```
+> Note: In the future, when you contribute to Galaxy, you'll need to clone the `dev` branch, which is the default, so you don't need to specify the `-b` option in the `git clone` command).
+> 
+> 3. Before we can use Galaxy, we need to create a virtual environment and install the required dependencies. First, let's create a virtual environment:
+> ```shell
+> $ virtualenv .venv
+> ```
+> Make sure your Python version is at least 3.6 (you can check your Python version with `python --version`). If your system uses an older version, you may specify an alternative Python interpreter using the `-p` option:
+> ```shell
+> $ virtualenv -p PATH-TO-PYTHON-INTERPETER .venv
+> ```
+> 4. Activate your new virtual environment:
+> ```shell
+> . .venv/bin/activate
+> ```
+> Once activated, you'll see the name of the virtual environment prepended to your shell prompt: `(.venv)$`.
+> 5. Now you are ready to install the dependencies. Since you wil be doing development, you need to install the dependencies listed in `lib/galaxy/dependencies/dev-requirements.txt`:
+> ```shell
+> $ pip install -r lib/galaxy/dependencies/dev-requirements.txt
+> ```
+> 6. Finally, let's create a new branch for your edits:
+> ```shell
+> $ git checkout -b my-training
+> ```
+> Now when you run `git branch` you'll see that your new branch is activated:
+> ```shell
+>   training_dev_debugging
+> * my-training
+> ```
+> Note: `my-training` is just an example; you can call your new branch anything you like.
+{: .hands_on}
 
-## Unit Test Failure
+# Unit test failure
 
-TODO
+Most of Galaxy unit tests are designed to test a separate component or function, are very fast, and do not require a running Galaxy instance. Finding the issue causing a unit test to fail is relatively straightforward. 
 
-### Finding the failing test on GitHub
+> ### {% icon hands_on %} Hands-on: Fixing a failing unit test
+> 
+> 1. __Finding the failing test on GitHub__
+>    > 
+>    > Head over to the [Galaxy repository on GitHub](https://github.com/galaxyproject/galaxy) and in the branch drop-down select the "training_dev_debugging" branch. Click the red "x" next to the last commit, you'll get a drop-down listing of workflow results; find "Unit tests / Test (3..." and click "Details". This will display a detailed report that includes information about the failed tests. Try to find the failed tests.
+>    > 
+>    >  > ### {% icon solution %} Solution
+>    >  > 
+>    >  > The log lists 2 failed tests:
+>    >  > ```
+>    >  > - FAILED lib/galaxy/model/tags.py::galaxy.model.tags.TagHandler.parse_tags
+>    >  > - FAILED test/unit/util/test_utils.py::test_strip_control_characters
+>    >  > ```
+>    >  > The first is a doctest. While it will provide us with hints on what may have caused the error, the unit test is generally more helpful. 
+>    >  {: .solution }
+>    >  
+> 2. __Running the test locally__
+>    > 
+>    > An essential requirement for productive development is a fast feedback loop: we want to make a change, run a test, and get immediate feedback. If we push our edits to the remote fork (or submit a pull request to the upstream repository) and wait for the the tests to run remotely, it could take hours before we get any feedback, which is very inneficient (and detrimental to sustaining the joy of programming).
+>    > 
+>    > Instead, we will run the tests locally. Furthermore, we will run a specific test that we know is failing: this will give us the instant feedback we need.
+>    > 
+>    > Make sure your virtual environment is activated. If not, activate it: `. .venv/bin/activate`
+>    > 
+>    > The simplest way to run the failing test locally is using pytest directly:
+>    > ```shell
+>    > $ pytest test/unit/util/test_uril.py
+>    > ```
+>    > Do you see the error message?
+>    > 
+>    >  > ### {% icon solution %} Solution
+>    >  > 
+>    >  > The test gives us a very specific error message:
+>    >  > ```
+>    >  > AssertionError: assert 'b l a' == 'bla'
+>    >  > - bla
+>    >  > + b l a
+>    >  > ?  + +
+>    >  > ```
+>    >  > This means that instead of the expected `bla` we got `b l a` (with spaces). 
+>    >  {: .solution }
+>    >  
+> 3. __Finding and fixing the issue__
+>    > 
+>    > Our next step is to open the test file and see what part of the codebase (i.e., what module, class + method or function) is failing to provide the expected result. 
+>    > 
+>    > The failing test is calling the `strip_control_characters` function in the `util` module. That function, apparently, is not producing the expected output. Let's head over to `lib/galaxy/utl/__init__.py`, where we'll find the function definition:
+>    > 
+>    > ```python
+>    > def strip_control_characters(s):
+>    >     """Strip unicode control characters from a string."""
+>    >     return " ".join(c for c in unicodify(s) if unicodedata.category(c) != "Cc")
+>    > ```
+>    > The function takes a string as input, and, in theory, returns a copy of that string with certain characters stripped away (it doesn't really matter what characters are stripped away). However, from the test we see that the string is modified in a different way: spaces have been added between the characters. Can you see what's causing this behavior?
+>    > 
+>    >  > ### {% icon solution %} Solution
+>    >  > 
+>    >  > The genexp is evaluated and subsequently joined using a space as a delimiter. The space is, obviously, a typo: we don't want to *add* anything to the string, so the delimiter should be the empty string instead:
+>    >  > 
+>    >  > ```python
+>    >  > return "".join(c for c in unicodify(s) if unicodedata.category(c) != "Cc")
+>    >  > ```
+>    > {: .solution }
+>    > 
+>    > Make the change, save the file, run the same test. Now it should pass.
+>    >
+> 4. __Finishing up__
+>    > 
+>    > Now you are ready to commit your edits and push the branch to your fork. So, `git add [your modified files]` and `git commit -m "[insert your commit message here]"`. Please consult [Galaxy's contributor's guide](https://github.com/galaxyproject/galaxy/blob/dev/CONTRIBUTING.md#how-to-contribute) for details on how to write a useful and properly formatted commit message, and why it matters.
+>    > 
+>    > Congratulations - you found and fixed a unit test failure!
+{: .hands_on}
 
-TODO
+# API test failure
 
-### Running the test locally
+API test various aspects of the Galaxy API, as well as general backend aspects of Galaxy using the API. These tests require a Galaxy instance.
 
-TODO
+> ### {% icon hands_on %} Hands-on: Fixing a failing API test
+> 
+> 1. __Finding the failing test on GitHub__
+>    > 
+>    > We'll follow the same process we used in the unit test section. On GitHub, click the same red "x" and select the "API tests / Test (3..." workflow and click "Details". Try to find the failed tests.
+>    > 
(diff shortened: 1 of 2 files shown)


ic4f:
OK, so now I remember why we were holding off with this PR! The key points are not done. dammit... We'll submit then in a separate commit.

hexylena:
Not a problem :) this at least lets everyone else start looking and making suggestions but the formatting already looks quite good :)

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  
+>    >  Try to identify which tests are failing based on artifacts you downloaded.
+>    >  
+>    >  > ### {% icon solution %} Solution
+>    >  >
+>    >  > The failing tests are:
+>    >  >
+>    >  > - `test_build_list_and_show_items`
+>    >  > - `test_data_column_input`
+>    >  > - `test_upload_pair_specify_extension`
```
davelopez, line 422: [The Selenium test that is failing](https://github.com/galaxyproject/galaxy/runs/2784473904) in the `training` branch is:
```
lib/galaxy_test/selenium/test_history_dataset_state.py::HistoryDatasetStateTestCase::test_dataset_change_dbkey
```
According to https://github.com/galaxyproject/galaxy/commit/8f1e14bed8a24da6117051f13f30cd96ff62318f it is probably correct, but then this tutorial block probably needs to be updated to address it instead of the proposed ones or vice-versa.

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  > ### {% icon solution %} Solution
+>    >  >
+>    >  > The failing tests are defined in these files:
+>    >  >
+>    >  > - `lib/galaxy_test/selenium/test_collection_builders.py`
+>    >  > - `lib/galaxy_test/selenium/test_workflow_editor.py`
+>    >  > - `lib/galaxy_test/selenium/test_uploads.py`
+>    >  {: .solution }
+>    >  
+>    >  Now that we have the files, we can run the selenium tests. The command to do this should be very similar to `./run_tests.sh -selenium lib/galaxy_test/selenium/test_sign_out.py`. The only difference should be the python file where the tests are defined.
```
davelopez, line 442: Maybe it's too obvious, but since in the last tutorial block we were in the `GALAXY_ROOT/client` directory, it may be worth remember that we should be now again in the `GALAXY_ROOT` dir to run the commands.

davelopez (commented):
Excellent tutorial! Really easy to follow and enjoyable.

Only the Selenium block needs to be checked a bit in case there was some confusion with the tests that should fail in the `training` branch.

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+
+> ### {% icon comment %} Python Linting
+> For backend consistency, we also have a Python linter. While we won't be running you through that exercise today, you may encounter a Python linting issue in the future. The way to diagnose and treat these is very similar to the way to do it for the front end.
+{: .comment}
+
+> ### {% icon hands_on %} Hands-on: Fixing client linting error
+> 
+> 1. __Finding the failing test on GitHub__
+> 
+>    > One of the failing tests on GitHub says "get_code_and_test". Clicking on the word "Details" assocated with this failure will open a brief view of the CircleCI Checks, with a list of tests that are run as part of that command. You'll see that the one that is failing is js_lint. And if you click that, you'll be navigated to the CicleCI app, where you can see much more information about the failures.
```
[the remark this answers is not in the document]
davelopez: Oh! I just saw that you already tried that workaround in https://github.com/galaxyproject/galaxy/pull/12111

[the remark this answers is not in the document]
ic4f: Yeah, it did it again. I think we need to point to the specific commit in the tutorial where that workflow (or build) did in fact run: https://github.com/galaxyproject/galaxy/runs/2780158706 .

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    > 
+>    > We can also see that it has found 1 error, where that error is located, and a description of the problem.
+>    > 
+> 2. __Running the linter locally__
+>    > 
+>    > Alright, so we've seen the failure, but we want to verify it locally now. For this, you'll need to open up a terminal.
+>    > 
+>    > Make sure your virtual environment is activated. If not, activate it: `. .venv/bin/activate`
+>    > 
+>    > Then you've got two options. You can run the same exact command that CircleCI uses `cd client && yarn run eslint` or `make client-lint`, which does `cd client && yarn run eslint && yarn run prettier-check`. This will also make sure your code conforms to formatting standards.
```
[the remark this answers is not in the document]
ic4f: Hmm, I always took it for granted: I thought it was installed automagically! If not then yes, sure, we should include that!

topics/dev/tutorials/debugging/tutorial.md
ic4f: Good point. Maybe each time we refer to paths we should prepend GALAXY_ROOT - then the difference would be obvious.

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  > The failing test file is `src/components/RuleBuilder/SavedRulesSelector.test.js`, and the failing test in that file is `SavedRulesSelector › disables history icon if there is no history`. 
+>    >  {: .solution }
+>    >  
+> 2. __Running the test locally__
+>    >  
+>    > Make sure your virtual environment is activated. If not, activate it: `. .venv/bin/activate`
+>    >  
+>    >  To run the client tests locally, you must be in the client directory; if you're just in the galaxy directory, `cd client` should take you to the right place.
+>    >  
+>    >  Next, type `yarn run test` or `yarn jest` to run all the client tests.
```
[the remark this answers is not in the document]
ic4f: It runs on my setup. Misconfig on your end?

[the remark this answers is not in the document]
davelopez: Most probably, but I didn't really set up anything apart from the steps of the tutorial. I didn't even have `yarn` installed in this machine as you can see :laughing:

Feel free to ignore this, if someone has a similar problem they have `yarn jest` as an alternative so, no problem.

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  There may be actually a few failing Selenium tests, but the good new is, they are the result of only one little error. The failing tests on GitHub says "Selenium Test 3.7, 2". When you click on the details of this one, it's not always very clear which test is causing the failure, but if you download the artifacts, it's very clear which tests have failed.
+>    >  
+>    >  Try to identify which tests are failing based on artifacts you downloaded.
+>    >  
+>    >  > ### {% icon solution %} Solution
+>    >  >
+>    >  > The failing tests are:
+>    >  >
+>    >  > - `test_build_list_and_show_items`
+>    >  > - `test_data_column_input`
```
assuntad23, line 421: @ic4f This test is not failing... whoops? We should remove this line

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  Instead, we can build our commands to only run a subset of the tests. Here's an example: `./run_tests.sh -selenium lib/galaxy_test/selenium/test_sign_out.py`
+>    >  
+>    >  A command like that will still run all of the tests defined in that python file, so we need to find out where the tests that failed are defined. Try seaching the baseline for our three failing Selenium tests; what files are they in?
+>    >  
+>    >  > ### {% icon solution %} Solution
+>    >  >
+>    >  > The failing tests are defined in these files:
+>    >  >
+>    >  > - `lib/galaxy_test/selenium/test_collection_builders.py`
+>    >  > - `lib/galaxy_test/selenium/test_workflow_editor.py`
```
assuntad23, line 438: And the corresponding file here...

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  > - `lib/galaxy_test/selenium/test_uploads.py`
+>    >  {: .solution }
+>    >  
+>    >  Now that we have the files, we can run the selenium tests. The command to do this should be very similar to `./run_tests.sh -selenium lib/galaxy_test/selenium/test_sign_out.py`. The only difference should be the python file where the tests are defined. 
+>    >
```
assuntad23, line 443: @ic4f, I never give them a sample command to run in a Solution block... I'm certain that quick readers will simply think test_sign_out.py is what they're supposed to run and be confused when it all passes. Maybe we should explicitly state the command, as we do everywhere else.

topics/dev/tutorials/debugging/tutorial.md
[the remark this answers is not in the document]
assuntad23: This is unfortunate...

topics/dev/tutorials/debugging/tutorial.md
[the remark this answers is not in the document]
assuntad23: I'm pretty sure (though, by no means an expert & have great potential to be wrong) that yarn is installed as part of the virtual environment, which would make it unnecessary to install it first.

Someone -- fact check me.

topics/dev/tutorials/debugging/tutorial.md
assuntad23: Another solution, which is potentially not the best idea, is to just delete the selenium test that is failing here? 

Make it seem more like a runtime error that doesn't have any existing code coverage. ¯\_(ツ)_/¯

topics/dev/tutorials/debugging/tutorial.md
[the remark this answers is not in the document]
davelopez: You are right, the documentation says:

> The base dependencies used are Node.js and Yarn. Galaxy now includes these in
the virtual environment, and they can be accessed by activating that with `.
.venv/bin/activate` from the Galaxy root directory.

I don't know what went wrong then, I checked the tutorial in a freshly installed Ubuntu and I would have sworn that the virtual environment was activated (as I carefully followed every step) but, in my case, I have **huge** potential to be wrong :laughing:  so this can be safely ignored, as I said, is not a big deal and `make client-lint` points you to the installation instructions in case someone steps into it :)

topics/dev/tutorials/debugging/tutorial.md
davelopez: I think that's a great idea and makes the tutorial more reproducible and easy to follow :)

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+> 1. TOC
+> {:toc}
+>
+{: .agenda}
+
+# Local development environment setup
+
+For almost every type of error you encounter, there will be a relatively similar process:
+
+1. Locate failing test or error on GitHub workflows
```
OlegZharkov, line 59: maybe github actions?

topics/dev/tutorials/debugging/tutorial.md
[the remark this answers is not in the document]
ic4f: Thank you for checking and verifying!

topics/dev/tutorials/debugging/tutorial.md
ic4f: I agree, let's have explicit commands in solutions (otherwise users won't be able to verify, beyond any doubt, that what they did was right and not just coincidentally worked)

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+4. Commit your changes, push branch to your fork
+5. Open a pull request for this branch in the upstream Galaxy repository
+
+In this tutorial, instead of cloning Galaxy's default branch, we will clone a branch that has been created for this tutorial and contains several bugs that you'll need to identify and fix. Also, we won't be opening a pull request in the upstream Galaxy repository.
+
+## Getting started: Pulling the buggy branch
+
+Good news! All of the failing tests are on the same Galaxy branch! That means you only need to download and work off of one branch. The downside is that this is one buggy branch with issues all over the place. We have no doubts that you can solve them though!
+
+> ### {% icon hands_on %} Hands-on: Setup your local Galaxy instance
```
OlegZharkov, line 85: to be honest, I am not sure if we need this section. all this can be created with `sh run.sh`
Do we really want them to configure virtualenv themselves?

topics/dev/tutorials/debugging/tutorial.md
ic4f: If this is more appropriate/easy to understand, yes, sure! (was my idea to call them workflows - and I was not at all sure that was a good term in this context). Feel free to suggest the necessary edits - there may be more than one instance.

topics/dev/tutorials/debugging/tutorial.md
ic4f: It's much faster than `run.sh`, and we don't need `run.sh` to run tests in the first part of the tutorial. I think that'll get them going faster; besides, that may be yet another useful thing to learn - it might take the mystery out of .venv for some people, I think.

OlegZharkov: maybe (just to skip the installation part), we can show them how to run 1 random test with `run_tests.sh`? It should create .venv as well

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    > Make sure your virtual environment is activated. If not, activate it: `. .venv/bin/activate`
+>    > 
+>    > Then you've got two options. You can run the same exact command that CircleCI uses `cd client && yarn run eslint` or `make client-lint`, which does `cd client && yarn run eslint && yarn run prettier-check`. This will also make sure your code conforms to formatting standards.
+>    > 
+>    > After running that, you should see the same linting error from CircleCI; you should know where it is and how to fix it too!
+>    > 
+> 3. __Finding and fixing the issue__
+>    > 
+>    > > ### {% icon solution %} Solution
+>    > > In the file `client/src/components/RuleBuilder/RuleComponent`, there is an unused import. Simply delete the line that says `import Vue from "vue";`
```
OlegZharkov, line 331:
```suggestion
>    > > In the file `client/src/components/RuleBuilder/RuleComponent`, there is an unused import. Simply delete the line that says `import Vue from "vue";` 
Alternatively, you can run `yarn eslint --fix`
```

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    > If you think you've got it, try running the test locally again to be sure.
+>    > 
+> 4. __Finishing up__
+>    > 
+>    > Whoo-hoo! If you're here, you've identified the error, ran the appropriate command locally, found the problem, fixed it, confirmed that the command runs without errors, and are ready to commit your change. Write a useful commit message like `Fix a client linting issue in RuleComponent` or something similar, making sure it is clear what you've done.
+>    > (See [Galaxy's contributor's guide](https://github.com/galaxyproject/galaxy/blob/dev/CONTRIBUTING.md#how-to-contribute) for details on how to write a useful and properly formatted commit message.)
+>    > 
+>    > Awesome work - you now understand how to solve client linting issues!
+{: .hands_on}
+
```
OlegZharkov, line 343:
```suggestion
5. __Little Hint__
 You also can auto-fix your linting problems with `yarn run prettier` and `yarn eslint --fix`
```

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+# Selenium test failure
+
+Selenium is the end-to-end or integration testing framework that we use; so these tests often are designed to test both frontend and backend code. This makes fixing them a little difficult, but we are sure you can do it!
+
+> ### {% icon hands_on %} Hands-on: Fixing a failing Selenium test
+> 
+> 1. __Finding the failing test on GitHub__
+>    >  
+>    >  There may be actually a few failing Selenium tests, but the good new is, they are the result of only one little error. The failing tests on GitHub says "Selenium Test 3.7, 2". When you click on the details of this one, it's not always very clear which test is causing the failure, but if you download the artifacts, it's very clear which tests have failed.
```
OlegZharkov, line 412:
```suggestion
>    >  There may be actually a few failing Selenium tests, but the good news are, they are the result of only one little error. The failing tests on GitHub says "Selenium Test 3.7, 2". When you click on the details of this one, it's not always very clear which test is causing the failure, but if you download the artifacts, it's very clear which tests have failed.
```

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  
+>    >  Running any of these three tests will show us that the tests are still failing on our local machine. So, you should see a failing test that matches one of the three we identified from the downloaded artifacts. Onward to fixing them!
+>    >  
+> 3. __Finding and fixing the issue__
+>    >  
+>    >  Good news - there's an easy way to figure out what's going wrong with a selenium test! In your terminal type `export GALAXY_TEST_SELENIUM_HEADLESS=false`. This will allow you to see the test running in real time. Re-run the selenium test to watch the headless browser step through the test. Looking at the test run in the headless browser, alongside the definition for the test should help you understand what might be going wrong. Try and figure out where the problem is on your own, before looking at the solution.
+>    >  
+>    >  > ### {% icon solution %} Solution
+>    >  >
+>    >  > The problem is that the tests are expecting a different HID than what is being made available to them.  Since all the failing tests have to do with Lists, you might want to check out the ListCollectionCreator.
```
OlegZharkov, line 452:
```suggestion 
The problem is that the tests are expecting a different HID (history identifier) than what is being made available to them.  Since all the failing tests have to do with Lists, you might want to check out the ListCollectionCreator.
```

topics/dev/tutorials/debugging/tutorial.md
OlegZharkov: > Another solution, which is potentially not the best idea, is to just delete the selenium test that is failing here?
> 
> Make it seem more like a runtime error that doesn't have any existing code coverage. ¯_(ツ)_/¯

I support this idea. It's not like we're developing something for production. Our goal here is to create a small usecase, deleting them will reduce confusion and that will contribute the teaching process.

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+>    >  > Here you should notice that in the PairCollectionCreator or in the PairedListCollectionCreator, `:hide-source-items="hideSourceItems"` instead of `:hide-source-items="!hideSourceItems"`. It's a little change, just flipping the boolean, but it makes a big difference! 
+>    >  {: .solution }
+>    >  
+>    >  Awesome! If you're here, you've found and fixed the problem, at least, in theory.  
+>    >  
+>    >  If you think you've got it, try running the tests locally again to be sure. In this simulated case, you can get by with just running one of the tests, but in reality, you should probably run all three Selenium tests to be sure that it's totally fixed. 
+>    >  
+> 4. __Finishing up__
+>    >  
+>    >  Excellent! If you're here, you've found the problem, ran it locally, fixed the problem, ran it locally again without errors, and are ready to commit your change. Write a useful commit message like `Fix logic error in ListCollectionCreatory` or something similar, making sure it is clear what you've done.
```
OlegZharkov, line 467:
```suggestion
>    >  Excellent! If you're here, you've found the problem, ran it locally, fixed the problem, ran it locally again without errors, and are ready to commit your change. Write a useful commit message like `Fix logic error in ListCollectionCreator` or something similar, making sure it is clear what you've done.
```

topics/dev/tutorials/debugging/tutorial.md
```diff
@@ -0,0 +1,730 @@
...
+    tutorials:
+      - architecture
+subtopic: core
+---
+
+# Introduction
+{:.no_toc}
+
+
+In this tutorial we will demonstrate how to find and fix common types of bugs you may encounter as a contributor to Galaxy. We will step you through the process of finding and fixing a bug - from locating specific errors in the logs of Galaxy's GitHub workflows, to identifying their cause, developing a solution and committing your edits.
```
OlegZharkov, line 44:
```suggestion
In this tutorial we will demonstrate how to find and fix common types of bugs you may encounter as a contributor to Galaxy. We will step you through the process of finding and fixing a bug - from locating specific errors in the logs of Galaxy's GitHub Actions, to identifying their cause, developing a solution and committing your edits
```

topics/dev/tutorials/debugging/tutorial.md
OlegZharkov, line 59:
```suggestion
1. Locate failing test or error on GitHub Actions
```

topics/dev/tutorials/debugging/tutorial.md
ic4f: Applied in separate commit.

topics/dev/tutorials/debugging/tutorial.md
ic4f: Applied, only "new is", not "are". (I know it feels like it should be plural, but the noun is uncountable, so "is")

topics/dev/tutorials/debugging/tutorial.md
ic4f: Applied. Good point!

topics/dev/tutorials/debugging/tutorial.md
ic4f: Done.

topics/dev/tutorials/debugging/tutorial.md
ic4f: I didn't prepend it to all paths - I thought it looked more confusing that way; but I did make it explicit throughout the tutorial: 6966e47593be42e208d4c27edccdcbfb4c45db36

topics/dev/tutorials/debugging/tutorial.md
ic4f: Deleting a test is an option.. but that opens precendent: on a future update some other tests will fail due to our "installed" bugs - we would have to delete those too, etc.
If there are multiple failing tests and only some of them are relevant to a given bug, I think we can explain why there are to be expected multiple test failures - somewhere at the beginning of the tutorial (or at the beginning of that section). If the test is not available (i.e., it did not run due to some other bug we introduced), then we can just name that test.

I'm not opposed to deleting tests to accommodate the tutorial, but I think we should try to avoid this if we can.

topics/dev/tutorials/debugging/tutorial.md
ic4f: We do ask them to use `run _tests.sh` for the api and selenium sections. Also, I've added this line that says that run_tests.sh will activate the venv (not create, since we ask them to create it at the beginning of the tutorial, so they already have it):
https://github.com/galaxyproject/training-material/pull/2614/files#diff-03c321ac89fe848061d63bb53e4d75eec9a2f914b4b13efa2f86afae057572a4R238

hexylena:
I pushed a couple of formatting fixes just fyi. The main thing was a whitespace issue causing your numbering to reset in the setup instructions.

We also have the code-in / code-out classes for showing commands with their output very clearly. Since there are a lot of commands run at the start maybe it's more annoying than it's worth, but I thought I'd show it as an example, feel free to revert.

image

additionally I removed some of the prefixed `$` from commands as it makes it easier for people to copy and paste without it erroring.

ic4f:
@hexylena thank you for the edits!
 
The solution wrapped in a Question box looks very nice (and logical).

I considered code-in/code-out, but, I think, that's the only place where we need to display output that is a result of code we've just typed in. But for this example it looks good and makes it clear, thanks!

I was torn about the prefix - other tutorials use it inconsistently. I decided to go with the prefix to distinguish between (a) code that a user is expected to type in at the command prompt versus (b) output from tests or Python code, etc. So it's a trade-off between simplifying copying and pasting -- and distinguishing between a command to be run and code to be read.

topics/dev/tutorials/debugging/tutorial.md
ic4f: Added as a tip to the linting section. Thanks!

hexylena:
Ahh interesting to hear your rationale. That's a bit why we added the in/out boxes as well, to help make it clear what is something you type vs something that's an output.

I'd love to get rid of the `$` completely from every other tutorial because people really like to copy and paste based on feedback. That said, it's a more technical audience here, maybe they're used to that. And additionally there is the option to prefix with `$` in css which would prevent it being copied, but we'd probably need to annotate the code blocks where there's a command.

ic4f:
@hexylena Thanks for the suggestion. At first I thought of in/out boxes only to be applicable to exampes of direct cause and effect - which is not often the case; but I never considered input only - and that seems fine. I've added more boxes for consistency.

topics/dev/tutorials/debugging/tutorial.md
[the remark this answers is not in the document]
ic4f: Handled in 9d055bcb6570fbf87cf3819bdf3684f618cb7709

topics/dev/tutorials/debugging/tutorial.md
ic4f: This has been updated. I'll edit the wording too to make it clear that more than one workflow needs to be checked.

topics/dev/tutorials/debugging/tutorial.md
ic4f: already fixed - tests fail, so this is fine.

ic4f:
OK, this should be done.

hexylena:
I'll get it reviewed today.

hexylena (approved):
This looks fantastic @assuntad23 @ic4f @jmchilton. I really wish I'd had this tutorial when I was starting out as a galaxy contributor, it would've made my life so much better. I like the variety of error cases presented, and the improvements to the linting such that a number of steps are "read the error message, you should know how to fix it". Fantastic

I'm happy with the formatting, everything looks good to me. I'll merge when tests pass.

hexylena:
Special thanks to @davelopez @OlegZharkov for the reviewing!

PR #2615, opened 2021-06-09
Revision on the architecture of wiring up Galaxy components.

jmchilton:
- Broaden topic of that new section from DI to `app` and DI (there was one random slide in this direction but it felt very disconnected).
- Add a lot more about app and how we typed it.
- Describe why interfaces are important to flush out some typing information.
- UML to describe how we decomposed `app` for Celery and tasks.
- End section with a description of best practices around typing and DI and wiring components.


hexylena (approved):
Your blockquote styling looks so good that I made it global. We should've been doing something like that from the start!

jmchilton:
Thanks for the blockquote fix and quick merge. You're a super hero @hexylena !

PR #2617, opened 2021-06-10
Update the tutorial for creating tutorials 

bebatut:
.. to start from workflow and also using ptdk webserver


faqs/galaxy/user_get_api_key.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Get API key on a Galayxy server
```
hexylena, line 2: I think this could probably use `preferences_admin_api_key.md`? Right now I know it has text for admin, but, there's already an if/else for dev, we could make it the same for admin maybe?

bebatut: I missed this snippet. I will use this one

hexylena (commented):
This looks like a great improvement! So much nicer to use ptdk rather than the more manual optinos.

hexylena (approved):
This looks like a great improvement, thanks @bebatut

topics/galaxy-interface/tutorials/workflow-automation/tutorial.md
```diff
@@ -189,7 +189,7 @@ The `tutorial.ga` file defines the workflow in JSON format; if we are confident
 
 Now we have a simple workflow, we can run it using `planemo run`. At this point you need to choose a Galaxy server on which you want the workflow to run. One of the big public servers would be a possible choice. You could also use a local Galaxy instance. Either way, once you've chosen a server, the next step is to get your API key.
 
-{% snippet faqs/galaxy/preferences_admin_api_key.md dev=true %}
+{% snippet faqs/galaxy/preferences_admin_api_key.md %}
```
wm75, line 192: Thanks for fixing @bebatut !
Just thought of doing this, but you were very careful it seems :)

PR #2619, opened 2021-06-10
Convert remaining admin tutorials to new format

hexylena:


hexylena:
git-gat now features dated branches. `main` will always be the latest version of the code, but we'll simultaneously push commits to a branch with today's date, thereby ensuring we have a more complete history of git-gat (since previously we were overwriting main branch and losing 'old' gitgats)

https://github.com/hexylena/git-gat/branches

Slugger70 (approved):

PR #2623, opened 2021-06-11
Small enhancements to Short Intro slides

nsoranzo:
- Add contributors and some text from topics/introduction/slides/introduction.html
- Use different wording between slides and notes.

Working towards making the old "Introduction to Galaxy" slide deck obsolete. I'd like to port some content from the "Tool Shed" and "Datatypes" slide in a future PR, then replace it for good.

### DIFF
--- a/topics/introduction/tutorials/galaxy-intro-short/slides.html
+++ b/topics/introduction/tutorials/galaxy-intro-short/slides.html
@@ -18,16 +18,7 @@ key_points:
 contributors:
   - shiltemann
   - hexylena
-  - bagnacan
-  - bebatut
-  - pajanne
   - nsoranzo
-  - chrisbarnettster
-  - michelemaroni89
-  - annefou
-  - nagoue
-  - olanag1
-  - tnabtaf
 ---
 
 # What is Galaxy?
--- a/topics/introduction/tutorials/galaxy-intro-short/slides.html
+++ b/topics/introduction/tutorials/galaxy-intro-short/slides.html
@@ -18,6 +18,16 @@ key_points:
 contributors:
   - shiltemann
   - hexylena
+  - bagnacan
+  - bebatut
+  - pajanne
+  - nsoranzo
+  - chrisbarnettster
+  - michelemaroni89
+  - annefou
+  - nagoue
+  - olanag1
+  - tnabtaf
 ---
 
 # What is Galaxy?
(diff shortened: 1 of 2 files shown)


hexylena (approved):

shiltemann:
Thanks @nsoranzo I am ok with replacing the old slide deck, but just want to be a bit careful not to add too much information to this one. The whole reason I made these was because the existing one was information overload for a lot of my audiences, too many details before ever even seeing Galaxy. 

I wanted to focus here on the very basic flow of an analysis (uploading data, analysing, viewing results, repeating, sharing) and things like datatypes/toolshed etc maybe can wait until they have a bit more of a sense of Galaxy? So maybe we just turn it around a bit, make this the intro deck, and maybe add a more extensive slide deck to go with the Galaxy 101 tutorial for example, what do you think? (there are other quite nice things like FTP upload, reference databases, the top menu etc in the other slide deck that would be good to keep in a longer slide deck for people who maybe have had their first look at Galaxy already and want to dive a little deeper)

topics/introduction/tutorials/galaxy-intro-short/slides.html
```diff
@@ -18,6 +18,16 @@
...
+  - bagnacan
+  - bebatut
+  - pajanne
+  - nsoranzo
+  - chrisbarnettster
+  - michelemaroni89
+  - annefou
+  - nagoue
+  - olanag1
+  - tnabtaf
```
shiltemann, line 30: These are relatively minor edits, shall we add just you here? (if we end up replacing the existing intro slide deck I am all for not erasing that attribution, but if we keep a long and a short version I would argue this one is different enough to be considered separately)

shiltemann (approved):

shiltemann:
Thanks @nsoranzo!

nsoranzo:
Thanks for merging!

> So maybe we just turn it around a bit, make this the intro deck, and maybe add a more extensive slide deck to go with the Galaxy 101 tutorial for example, what do you think?

I like this idea, maybe that could just be a collection of "extra" slides?

I guess the only info from the "Tool Shed" and "Datatypes" slides I'd like to port to the short intro is that normal users cannot install tools themselves; and that Galaxy tools accept inputs in specific formats, so it may be necessary to verify that uploaded data are correctly recognized. So these could probably just be small additions to the "Finding a tool" and "Running a tool" slides respectively.

PR #2624, opened 2021-06-11
Add document IDs for subtopics

hexylena:


_layouts/topic.html
```diff
@@ -50,7 +50,7 @@ <h2 id="material">Material</h2>
             {% endunless %}
             {% if topic.subtopics %}
               {% for subtopic in topic.subtopics %}
-                <h3>{{ subtopic.title}}</h3>
+                <h3 id="st-{{ subtopic.title }}">{{ subtopic.title}}</h3>
```
shiltemann, line 53: do we need to `| slugify` this in case there are spaces in title?

hexylena: ahhh you're brilliant, yep, we do. Good call.

shiltemann (approved):

PR #2652, opened 2021-06-18
Galaxy instance translation

nomadscientist:
This is a list of the stuff that is still not translated in the Galaxy Spanish Instance. Help!


hexylena:
Hey @nomadscientist I think this belongs as a comment in the galaxyproject/galaxy#11982 so I'll move it there if that's ok

PR #2654, opened 2021-06-18
first version of ftp tutorial

lldelisle:
Hi there,
Here is an attempt of a tutorial to link ftp to galaxy. I don't know much about ftp but I installed one on my local instance that's how I came to write this tutorial.
I made it work on gat-79.be.training.galaxyproject.eu thanks to @hexylena for the ports.
For the moment, the ports are the one I used on this machine because the port 21 was not available but ports 50k to 60k yes. I know @hexylena needed to change something on the cloud but I don't know what...
Comments both on the content and the form are welcome.
I am not sure that the `{:` I used are correct.


lldelisle:
@hexylena In the patches you commit there is a `usegalaxy_eu.gxadmin`

hexylena:
Ahhh that's why git-gat is failing.

hexylena (dismissed):
This looks fantastic @lldelisle I'm going to include it in this year's GAT!

I've made a couple small changes:
- swap to port 21: this would be a sensible default, it only isn't available because this year's VM supplier has an additional firewall we can't control, in the future we'll definitely ask to expose it.
- add some test cases to ensure everything works in the git-gat deployment
- moved /upload to /data/upload (since we store user data + backups there)

I'm so happy!

lldelisle:
Thanks I have a question. How do you know the password for the user 'admin@example.org' is password. It looks like we do not set it, do we?

lldelisle:
I found a 55000 that left.

lldelisle:
Building a tutorial in 2 days. We did a great job. I think you can add your name to the author list.

hexylena:
We set it to that for the testing scripts. We generally don't write that in the tutorial because we want people to set better passwords.

Not sure if you've seen the `.scripts` directory in git-gat but it has all of the tests cases + every `data-cmd` block that's extracted from the tutorials.

It's written here the default admin user + password + api key, it's automatically setup for us:
https://github.com/galaxyproject/training-material/blob/main/bin/knit-frog.py#L119

Oh good catch with the other port! Nah, no worries, I just did some gtn formatting, you did all the hard work of figuring out the configuration.

lldelisle:
I just saw it in the `.script`.
Thanks

hexylena (approved):

hexylena:
:tada: thank you so much for writing this up!

lldelisle:
:tada: Thank you for the proposition and for your reactivity these 2 days.

lldelisle:
Well I should say 365 days a year... :wink:

PR #2662, opened 2021-06-22
Add semantically-important whitespace in diffs

natefoo:
Some diffs were not displaying correctly:

bad-diff

@hexylena figured out that this was because the previous line must have a character in the first column of the diff, be it `+`, `-`, or ` ` (space). ~~I wrote some [quick and dirty scripts](https://gist.github.com/natefoo/4f9ce9717ed77f4e97347304de1d94b4) to detect and fix these.~~ this has now been added to the linting in CI.


hexylena (dismissed):

hexylena:
Huzzah! It caught one 8n FTP

hexylena (approved):

PR #2671, opened 2021-06-23
Make slides 16:9

hexylena:
Most common slide formats default to 16:9 since a while, this enables the more modern aspect ratio for us as well.

(probably merge after gcc to not surprise anyone?)

Before | After
-- | --
image | image


shiltemann (approved):

shiltemann:
thanks!

PR #2678, opened 2021-06-24
Remove slides folder and move introduction slide decks with other material

### LINKED ISSUE #2162: Suggestion: Moving content of `slides` folders to tutorials
Hi,

For a while, I found the `slides` folders in topics confusing:
- The content is a bit lonely on its side
- It can not be listed as prerequisite or follow-up in tutorials (in different topics)

What do you think about: 
- moving content of the `slides` folders to tutorials
- use the subtopics and define one as "introduction" that always goes to top
?

bebatut:
To fix #2162

The `slides` folder was a bit "lonely". The introduction decks fit with the other materials.

BTW, what about renaming `tutorials` folder to `materials`?

With the moving, the introduction slide decks do not appear anymore in the Supporting material in the tutorials. Should we try to add them? Or should we add the introduction manually as requirements?

To do:
- [x] Update Planemo training to remove `slides` folder
- [x] Update documentation/tutorials
- [x] Add redirect to all introduction slide decks


hexylena:
Nice! 

> BTW, what about renaming tutorials folder to materials?

We could, it would require some work I think to make sure the redirections work correctly, but it's possible.

Should we try remove `topics` and `materials` at the same time? Just keep the topic name/tutorial name? e.g. `/training-material/assembly/illumina/tutorial.html`

hexylena:
I'm doing this in another PR because it's a bit easier to cherry-pick out the individual commits, since it seems they got mixed in with a tutorial update that's harder to rebase (the scrna), but overall it still looks good and useful! We have no documentation (as far as I can tell) which mention `introduction.html` or creating those, so, nothing to update/remove! 

planemo updated as well. https://github.com/galaxyproject/planemo/pull/1344

Issue #2583, opened 2021-06-02
Pre-GAT 2022 Improvements


hexylena:
- [ ] Fully automated videos
- [ ] @mvdbeek suggests that the job_conf.xml file can be a job_conf.yaml, and the dependency reoslver related stuff can be yaml, even within galaxy.yml, doens't have to be a separate file. *This would trigger re-recording of like 10 videos, we'll wait until helena's automated videos work.*
- [ ] add cleanup cron job role

These require nginx package changes:

- [ ] add mod_zip, thanks @mvdbeek @lldelisle  https://docs.galaxyproject.org/en/latest/admin/nginx.html#creating-archives-with-mod-zip
- [ ] maybe add https://docs.galaxyproject.org/en/latest/admin/nginx.html#receiving-files-with-nginx too? it's better for performance

natefoo:
Regarding the job conf, yes, we have converted it to a (standalone) YAML file on .org, this works now. For the admin training, putting it directly in galaxy.yml may be preferable as it's fewer steps.

I tested dependency resolvers and container resolvers as standalone YAMLs and added support for those in the galaxy role.

Dependencies are also supported under `dependency_resolvers` or `dependency_resolution.resolvers` in `galaxy.yml`. Pulsar supports `dependency_resolvers.resolvers` as well (I dropped support for the standalone file in the pulsar role since you can just specify it directly in the app YAML), although we don't use it in the training.

hexylena:
awesome!!

natefoo:
Couple more things:

- Maybe should consider creating a `dbservers` group and putting postgres stuff in `group_vars/dbservers.yml`. It might be confusing up front but it will help people who later deploy on their own infra and have a separate DB server.
- Should we create `galaxy_user_name` and `galaxy_db_name` vars to avoid duplication in `postrgresql_objects_*` and `database_connection`?

natefoo:
Maybe move job_conf.xml and job_metrics_conf.xml to their own "adding additional configs" tutorial that gets run right after ansible-galaxy?

hexylena:
> Should we create galaxy_user_name and galaxy_db_name vars to avoid duplication in postrgresql_objects_* and database_connection?

it's so early in the training, later it makes sense, but I worry about defining a variable in step like 1, and students thinking they have to define variables for everything (even if that's often best practice.)

hexylena:
[...]

PR #2700, opened 2021-07-01
colour blind friendly diff colouring

hexylena:
@natefoo  https://github.com/galaxyproject/training-material/issues/2583#issuecomment-871492929 This is more or less the styling I use in my terminal (as a red-green impaired human) and it works quite well, and should be black-white colourblind friendly

```
[color "diff"]
	meta = bold
	frag = magenta bold
	old = strike reverse
	new = blue dim reverse
```

screenshot of diff with removed lines having a black background and strikethrough, and added changes highlighted with blue background.


natefoo (approved):

matuskalas:
Awesome, this works very nicely even in monochrome 👍🏽

PR #2713, opened 2021-07-09
Fix workflow maker

gallardoalba:
In the previous PR the protein dataset was disconnected.


hexylena (approved):

hexylena:
hate how hard these are to review, short of pulling them out. Definitely need the cytoscape IWC things and a diff of the WFs.

PR #2717, opened 2021-07-14
New GTN how to reproduce Galaxy analysis

foellmelanie:
@annefou thanks for all your help. Could you please check in the last Q&A that the image I put into the solution is the one that you meant?


topics/introduction/tutorials/galaxy-reproduce/tutorial.md
```diff
@@ -0,0 +1,312 @@
...
+> 4. Add an `#iris` tag {% icon galaxy-tags %} to the dataset
+>
+>    {% snippet faqs/galaxy/datasets_add_tag.md %}
+>
+>    Make sure the tag starts with a hash symbol (`#`), which will make the tag stick not only to this dataset, but also to any results derived from it.
+>    This will help you make sense of your history.
+>
+{: .hands_on}
+
+> ### {% icon comment %} Different types of datasets
```
hexylena, line 128: I love the content of this box, it's super useful, but is it relevant here? The reviewer isn't uploading composite datatypes, are you envisioning it being "for someone who might be reproducing another, unrelated analysis?" Might it make sense to collapse it by default (e.g. make it a comment for people to expand if they need it?)

hexylena (approved):
I love this, super helpful!

I noticed the output datasets are similar in the shared history step, is it precisely the same WF? Should the tutorial maybe mention something like "here's the workflow re-analysing another dataset, showing re-usability across inputs"?

topics/introduction/tutorials/galaxy-reproduce/tutorial.md
foellmelanie: yes yes and yes. How can I collapse the box?

foellmelanie:
> I love this, super helpful!
> 
> I noticed the output datasets are similar in the shared history step, is it precisely the same WF? Should the tutorial maybe mention something like "here's the workflow re-analysing another dataset, showing re-usability across inputs"?

I think so but not 100% sure anymore, @annefou do you remember?

annefou:
> > I love this, super helpful!
> > I noticed the output datasets are similar in the shared history step, is it precisely the same WF? Should the tutorial maybe mention something like "here's the workflow re-analysing another dataset, showing re-usability across inputs"?
> 
> I think so but not 100% sure anymore, @annefou do you remember?

 All the datasets used in this training are identical to those used in Galaxy 101 for everyone. Yes the shared history reuse the same workflow but with another dataset (the diamond dataset). Thanks @hexylena for your very good suggestion. It is indeed nice to mention one can reuse the same published workflow but with another dataset.

annefou:
@foellmelanie Is there anything else to do with this tutorial?

bgruening (approved):

bgruening:
I got a offline ok :)

PR #2718, opened 2021-07-15
Add a page with plots from embed feedback

bebatut:
We have now quite a bunch of feedback from the embed feedback forms. They are posted on the GitHub issue but not a good overview of the feedback. I worked before on some Jupyter notebooks but not here... So here is it

The page displays
- On the top, feedback from all topics (the different lines being the different topics)

   image 

- Feedback topic by topic (the different lines being the different tutorials in the topic)
- Feedback per tutorial (inside a accordion) with also a table to display the pro/con answers (only for non anonymous feedback)

Screenshot 2021-07-15 at 14-20-57 Galaxy Training

This page is available at `feedback.html` and linked within the extra menu (directly to topic feedback from topic pages, to tutorial feedback from tutorial pages). As the stat page, it uses chart.js

The data used to generate the plot is stored in `metadata/feedback.csv`. This file is created using a Python script `bin/prepare_feedback.py`. This script:
- get the results of the feedback form from Google sheet
- clean data and do some formatting (date, etc)
- rename tutorials, topics, etc when they have changed over time
- export as a CSV

Currently, the script needs to be run manually. Would it be possible to add it to a CRON job @shiltemann @hexylena?


bgruening:
That is so cool @bebatut! And the CSV export - awesome! Very nice work!

bin/prepare_feedback.py
```diff
@@ -0,0 +1,173 @@
...
+        .query('tutorial != "RNA-seq counts to genes and pathways"')
+        .query('tutorial != "Galaxy Tool Management"')
+        .query('tutorial != "Recording Job Metrics"')
+        #
+        .query('pro != "TESTING"')
+        .to_csv(out_file))
+
+
+if __name__ == '__main__':
+    url = 'https://docs.google.com/spreadsheets/d/1NfZhi5Jav7kl9zFCkeb7rIC2F8xW1isruv1TeO4WpNI/export?format=tsv'
```
hexylena, line 171: this would require making it public, right? that url 404s for me.

If we want to do that, should we remove the people who didn't consent to their feedback being public? Should we remove the checkbox completely and just write in bold text "This will be public"?

bin/prepare_feedback.py
```diff
@@ -0,0 +1,173 @@
...
+        .pipe(fix_tutorial_info)
+        # remove extra columns
+        .drop(columns=["tutorial_topic", "timestamp"])
+        # remove some rows 
+        .query('topic != "TESTIN"')
+        .query('topic != "tes"')
+        .query('topic != ""')
+        # remove tutorials
+        .query('tutorial != "RNA-seq counts to genes and pathways"')
+        .query('tutorial != "Galaxy Tool Management"')
```
hexylena, line 163: any reason why are these being removed? was the data bad or something? it isn't clear for me.

_layouts/feedback.html
```diff
@@ -0,0 +1,388 @@
...
+{% for f in fdbk_by_month %}
+    {% assign new_fdbk = f.items | size %}
+    {% assign fdbk_count = fdbk_count | plus: new_fdbk %}
+    {% assign fdbk_over_time = fdbk_over_time | push: fdbk_count %}
+    {% assign fdbk_over_time_labels = fdbk_over_time_labels | push: f.name %}
+{% endfor %}
+{% assign fdbk_by_note = fdbk | sort:'note' | group_by: 'note' %}
+<script>
+    var fdbk_over_month = {
+        all_topics: [
```
hexylena, line 116: would it make sense to have the python script generate these "APIs"? And put the data in the api folder for the javascript to access? It might be a bit easier to maintain than doing it in jekyll (but this looks like a lot of work, totally understand if you just want to keep it.)

bin/prepare_feedback.py
hexylena: @shiltemann reminds me that we've already done the "it's always public", shall we just remove those rows from the table and make it public?

(If we're making it public I guess we could just embed the google graphs directly and wouldn't need the round trip?)

bin/prepare_feedback.py
```diff
@@ -0,0 +1,173 @@
...
+        .query('tutorial != "Recording Job Metrics"')
+        #
+        .query('pro != "TESTING"')
+        .to_csv(out_file))
+
+
+if __name__ == '__main__':
+    url = 'https://docs.google.com/spreadsheets/d/1NfZhi5Jav7kl9zFCkeb7rIC2F8xW1isruv1TeO4WpNI/export?format=tsv'
+    out_file = Path('metadata') / Path('feedback.csv')
```
hexylena, line 172: should we make this TSV formatted? it'd be a bit safer for others to work with I guess. (Or JSON/yaml?)

bin/prepare_feedback.py
bebatut: Interesting, It works for me within a private window.
Not sure what to do with the private feedback (no private information are stored), as they are quite a lot. Currently I just used the notes from the private feedback, not the free text...
I am not sure to understand what do you mean by the google graphs?

bin/prepare_feedback.py
bebatut: I could not find these tutorials anymore. I think they were renamed or completely reshaped into new ones

_layouts/feedback.html
bebatut: I was not sure which approach is the best. I did not want to add too much on the GitHub repo but if it makes more sense, let's do that

bin/prepare_feedback.py
bebatut: I am not sure if Jekyll can read TSV file :smile:

bin/prepare_feedback.py
hexylena: Galaxy tool management is this one: https://training.galaxyproject.org/training-material/topics/admin/tutorials/tool-management/tutorial.html, job metrics was deprecated, yeah.

bin/prepare_feedback.py
hexylena: > Interesting, It works for me within a private window.

seems temporary, it's working now :)

>  Not sure what to do with the private feedback (no private information are stored)

it feels quite wrong to expose it just since they didn't consent (yeah they probably misunderstood/misread the question, but, still, we hadn't been exposing this previously)

>  I am not sure to understand what do you mean by the google graphs?

if we're going to publicly expose that feedback table, then we could make many of the same graphs in google sheets probably, right? Score by topic (since that's in the tutorial name), score by tutorial.

And then we could just embed those directly? It would save us the bother of setting up a cron job, we could remove the python, the jekyll template would just be a couple iframes

image

hexylena: we'd just have to handle the renaming/reshaping directly in the tables.

I think we can also publish graphs without publishing the entire spreadsheet which could help address the 'private' data? (But we wouldn't get the nice pro/con lists you've gone to the trouble to extract)

bebatut: I tried before (at the beginning) to do it within Google sheet but it did not render nicely (that is why I spent quite some times polishing data and creating these graphs first within Jupyter notebooks and then here). But feel free to give it a try

hexylena: I had not remembered that! No, not better, just thought maybe it would be easier. I trust your experience doing that, that that path isn't worth going down!

hexylena:
Something looks odd in genome annotation

image

maybe due to identical topic + tutorial title?

image

_layouts/feedback.html
```diff
@@ -201,7 +199,7 @@ <h2>Overall</h2>
                 <div class="col-md-6">
                     <div class="card">
                         <div class="card-body">
-                            <h5 class="card-title">{{ fdbk | size }} feedback</h5>
+                            <h5 class="card-title">Feedback over time</h5>
```
bebatut, line 202: I liked have the number displayed. Could we add it again in the title? Then at least the screen reader can have the number (if not the graphs)

bebatut:
I will check for the "Genome annotation" issue

_layouts/feedback.html
hexylena: sure! I was finding it a bit confusing with the multiple lines since there was an "overall" line and several individual ones, maybe it just needs the right wording or so.

hexylena (approved):

hexylena:
This looks fantastic! It's a great start and I look forward to a lot of things we can improve with this data.

shiltemann:
whoo!! great to have this accessible to all our tutorial authors! thanks a lot @bebatut !!

would you like to write a news item for it?

bebatut:
Yes I will write an news item :smiley:

PR #2728, opened 2021-07-27
Make path option in DL import snippet more useful

wm75:
You cannot be sure about paths in the DL on different Galaxy instances,
but you can provide a likely path to try unless the instructor says
otherwise.


shiltemann (approved):

shiltemann:
thanks @wm75!

PR #2729, opened 2021-07-27
SARS-CoV-2 mutation and lineage calling: Fix issues found in first round of user training

wm75:
Among others:
- Fixes links to auxiliary datasets
- Mentions need to adjust autodetected format for some of them
- Fixes mix-up of by-sample and by-variant reports in hands-on
- Adds workflowhub link, which became available only recently, for
  consensus WF


bgruening (approved):
Thanks @wm75!

PR #2746, opened 2021-08-12
Fix variant numbers mentioned in question boxes

wm75:
Also fixes a few typos and adds param info to the consensus workflow run hands-on.
All this based on user feedback from our SARS-CoV-2 Data Analysis and Monitoring with Galaxy workshop.


bgruening (approved):
Thanks 👍 @wm75

PR #2758, opened 2021-09-02
Tag all tutorials using interactive tools

hexylena:


martenson (approved):
looks good, thanks @hexylena !

PR #2770, opened 2021-09-15
Publish the workflows

hexylena:


hexylena:
@mvdbeek you mind approving this so we can test?

hexylena:
ignore the build failure, it's due to hub2 removing a url we were using.

mvdbeek (approved):

PR #2764, opened 2021-09-07
[WIP] tutorial "Introduction to Python"

mcmaniou:
First version of the tutorial, based on the Carpentries tutorials: https://swcarpentry.github.io/python-novice-inflammation/, https://swcarpentry.github.io/python-novice-gapminder/ . 

Submitted for the Gallantries project @mcmaniou, @fpsom

PR #2779, opened 2021-09-17
Automatically generate Jupyter notebooks in the GTN for the upcoming Python tutorial

hexylena:
The best learning experience for students when covering topics like Python and R and similar is that they run the training materials directly as notebooks. We have a couple tutorials already which function like this, e.g. https://training.galaxyproject.org/training-material/topics/transcriptomics/tutorials/scrna-JUPYTER-trajectories/tutorial.html which has as an initial step loading an ipynb file into Jupyter.

This takes this a step further, and allows tutorial authors to write content in GTN Markdown, have it render nicely in the GTN, and, **automatically generates a notebook** using [notedown](https://github.com/aaren/notedown) converting it directly into a reasonably ok looking notebook. Importantly the student can then run through each cell and test things out.

It automatically generates the links to mybinder.org which will let students click and immediately start running the notebook, as well as instructions on Jupyter in Galaxy.

*Current drawbacks*

The main current drawback of this approach is it necessitates removing and output preview from the markdown, which is a bit sub-optimal. However, given the ease of "launch on mybinder", I don't think this is a huge drawback, and it is something that @shiltemann and I have discussed and we think we have a plan forward to knit back the output, time and energy permitting.

Additionally tip boxes won't render nicely without some additional CSS, and we don't support includes currently (as it is processing the markdown directly.)

I still think this will be a fantastic feature for the set of tutorials that @gallantries and @avans-atgm are building over the coming years which will include Python, SQL, and maybe Bash/Git tutorials which could all leverage this framework.

ping @fpsom xref #2764 

Markdown Source | Built Notebook
--- | ---
image | image


hexylena:
All of this is also in #2787 so I'll close this one since it didn't get merged, and just continue work there.

PR #2782, opened 2021-09-20
Reads to counts: update tool name

mblue9:
'Column Join on Collection' is now called 'Column join on multiple datasets'  so updated that in tutorial


hexylena (approved):
Bril, thanks @mblue9 !

PR #2783, opened 2021-09-21
Fix build issues

hexylena:
I really wish we could enforce tests to pass :sob: come on github, support requiring tests if they're there, don't force us to run the entire test suite every time just to enforce passing tests.

Anyway this might fix the build


gallardoalba (approved):

hexylena:
danke @gallardoalba

PR #2790, opened 2021-09-23
MuSiC Tutorial

mtekman:
This is a work-in-progress of a MuSiC tutorial for bulk RNA deconvolution

@nomadscientist 

TODO:

- [x] Update tool links to latest music_deconvolution version after tool upgrade


mtekman:
@nomadscientist First full version of tutorial is finished. Please review and improve!

nomadscientist:
@mtekman Do you think we should keep the 'import data via data library' thing? I generally delete those on my tutorials. Is there a reason to keep it?

nomadscientist:
How were the datasets initially created? i.e. are they in a format that is an output of a Galaxy tutorial, or something we could pull directly from EBI, for instance?

nomadscientist:
- [ ] Workflow for generating scRNA input
- [ ] Workflow for generating bulk RNA input

nomadscientist:
Beautiful question / inspecting data section, really guides the user from AHH OVERWHELMED to OH THIS IS FINE. Nice work there

nomadscientist:
Construct Expression Set Object not working - https://humancellatlas.usegalaxy.eu/u/wendi.bacon/h/deconvolution-cell-type-inference-of-human-pancreas-data

mtekman:
> @mtekman Do you think we should keep the 'import data via data library' thing? I generally delete those on my tutorials. Is there a reason to keep it?

I think they're automatically there because I guess different Galaxy instances might have the data hosted locally (e.g. for workshops). I say leave it, since it's pretty much in all tutorials.

mtekman:
> How were the datasets initially created?

The input datasets are the raw count matrices and phenotype table pulled out of the RData objects from the example datasets (https://zenodo.org/record/5554814). 

There's a section in the tutorial (exploring the datasets)[https://github.com/galaxyproject/training-material/pull/2790/files#diff-9cc9f53223dd806885761598db0d4f1e8bb1f7960f0b0b0ae22fdecfb21bffaaR165] where we discuss how the input data is structured, and how the count matrices are related to the phenotype tables. 

This could maybe made a little bit more clearer -- all that really needs to be said there is that we just required two tables: expression counts, and a phenotype table describing the samples.

> i.e. are they in a format that is an output of a Galaxy tutorial, or something we could pull directly from EBI, for instance?

Yes, and yes. Both the count matrices and the phenotypes should be able to be extracted from the AnnData objects using Galaxy tools (in the case of the single cell data). For the bulk data, I think this is already in tabular format.

mtekman:
> Beautiful question / inspecting data section, really guides the user from AHH OVERWHELMED to OH THIS IS FINE. Nice work there

Hehe! If you know of a way to ease this transition more, then please make the required changes. I was describing it in a way that made sense to me, but my thinking is often at complete odds to others :P

nomadscientist:
> > Beautiful question / inspecting data section, really guides the user from AHH OVERWHELMED to OH THIS IS FINE. Nice work there
> 
> Hehe! If you know of a way to ease this transition more, then please make the required changes. I was describing it in a way that made sense to me, but my thinking is often at complete odds to others :P

LOL - this was nice and British of you, I give you a complement and you turn it into a critique

mtekman:
> @mtekman Construct Expression Set Object not working - https://humancellatlas.usegalaxy.eu/u/wendi.bacon/h/deconvolution-cell-type-inference-of-human-pancreas-data

It looks like it's still using the old version of the tool (0.1.1+galaxy0) instead of (0.1.1+galaxy1), which had some issues before, so this is expected.

We need to refresh this. It was hosted under the "Testing Tools" subsection of the tool list, but it doesn't seem to be there anymore.

@bgruening  Any chance of a refresh for the MuSiC tools? They run fine locally on my machine, and they're ready to replace the current broken version.

bgruening:
Its installed now, normal updates of all tools are running this weekend.

nomadscientist:
Screenshot 2021-12-22 at 09 30 36
This bit is not letting me select a dataset... nor is it automatically picking up the Rdatasets...

nomadscientist:
Screenshot 2021-12-22 at 09 33 33
Also the tools look doubled?

nomadscientist:
Screenshot 2021-12-22 at 09 34 37
Same issue here - even new tool fails to let me select inputs

nomadscientist:
OK so I got stuck on the tutorial tools not working, so I turned to looking at the dataset generation workflows
1. With the scRNA-seq data production, I see how you can download it from ArrayExpress and it works fine. BUT, what if they are using their own dataset? Then it would be processed (i.e. not integer counts), and I can't figure out how to create an integer-count matrix from a scRNA-seq AnnData object. Do you know how to do this?
2. With the bulk RNA-seq data production, I looked at its repository here: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE50244. But none of those files have integers. How was the dataset for the tutorial created?
Screenshot 2021-12-22 at 09 52 30 The MUSIC original vignette doesn't actually say. The good news here, however, is that the Galaxy RNA-seq reads to counts tutorial gives the data in the exact correct format, (with some minor adding of columns to the sample information table), so that's a massive win!

mtekman:
> Screenshot 2021-12-22 at 09 30 36
> This bit is not letting me select a dataset... nor is it automatically picking up the Rdatasets...

Yep - for now you need to physically drag and drop it from the history into the input slot. It's a known bug that should be fixed in the next release when the `rdata.eset` datatype is included in Galaxy.

mtekman:
> Screenshot 2021-12-22 at 09 33 33
> Also the tools look doubled?

Yep, also temporary - some issues with galaxy at the moment. One of the doubled tools should be the correct version, and one should be the older version.

It's nothing to worry about, it'll be gone by the next workshop

mtekman:
> OK so I got stuck on the tutorial tools not working, so I turned to looking at the dataset generation workflows
> 
>     1. With the scRNA-seq data production, I see how you can download it from ArrayExpress and it works fine. BUT, what if they are using their own dataset? Then it would be processed (i.e. not integer counts), and I can't figure out how to create an integer-count matrix from a scRNA-seq AnnData object. Do you know how to do this? Here's the answer key scRNA-seq history: https://docs.galaxyproject.org/en/latest/releases/20.01_announce_user.html

Integer or decimal counts can both be used AFAIK. You can extract the full raw integer count matrix from an AnnData object in Galaxy via the Inspect AnnData tool

Screenshot 2021-12-22 at 12-08-19 Galaxy Europe

>     2. With the bulk RNA-seq data production, I looked at its repository here: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE50244. But none of those files have integers. How was the dataset for the tutorial created?
>        Screenshot 2021-12-22 at 09 52 30 
> The MUSIC original vignette doesn't actually say. The good news here, however, is that the Galaxy RNA-seq reads to counts tutorial gives the data in the exact correct format, (with some minor adding of columns to the sample information table), so that's a massive win!

I'm not sure the integer counts are necessary, I think MuSiC can work on both normalised and unnormalised matrices.
All the RData datasets in the tutorials were generated using a count matrix and phenotype table pair to create the R objects for the rest of the tutorial, but it's literally just those two plain-text tabular inputs in a compact form.

MMmm maybe we might need another tutorial on how to structure phenotype data for MuSiC to work, but for the most part MuSiC just expects a count matrix and some accompanying table describing the samples. I think it should just work with everything

nomadscientist:
> > Screenshot 2021-12-22 at 09 30 36
> > This bit is not letting me select a dataset... nor is it automatically picking up the Rdatasets...
> 
> > Screenshot 2021-12-22 at 09 34 37
> > Same issue here - even new tool fails to let me select inputs
> 
> Yep - for now you need to physically drag and drop it from the history into the input slot. It's a known bug that should be fixed in the next release when the `rdata.eset` datatype is included in Galaxy.

Drag and drop is not working for me - does it work for you?

nomadscientist:
> 

1. Inspect AnnData only lets you extract the full data matrix, which I believe does not default to raw.

However, this matters less if non-integers works, that's brilliant news!!! I will have a play with this when the tools are working to verify for my own brain :)

2. I think that's great, my issue is, we can't actually say how the data was generated (which is a problem in the original vignette, they must have done SOMETHING to the raw data, because what is currently available is not in the integer form used in the tutorial). I might make a proof of principle history (in January) to show how all 3 tutorials (RNA-seq, scRNA-seq, and MUSIC) can work together... But I can add this after the tutorial is published, this concern is minor and definitely not necessary to solve before publishing this tutorial.

mtekman:
> Drag and drop is not working for me - does it work for you?

Yep, works for me!

test

> 
>     1. Inspect AnnData only lets you extract the full data matrix, which I believe does not default to raw.

Oh whoops you're right, I guess the raw is only saved in some backup slot, but I think that can be accessed via the `uns` selector in the Inspect AnnData tool

> However, this matters less if non-integers works, that's brilliant news!!! I will have a play with this when the tools are working to verify for my own brain :)

Okay cool!

>  2. I think that's great, my issue is, we can't actually say how the data was generated (which is a problem in the original vignette, they must have done SOMETHING to the raw data, because what is currently available is not in the integer form used in the tutorial). 

Oh I see the problem.... but is that really an issue? This is a tool that will be used well after pre-processing and after the downstream analysis, so the data will likely be modified many times over before it gets to the MuSiC stage me thinks

> I might make a proof of principle history (in January) to show how all 3 tutorials (RNA-seq, scRNA-seq, and MUSIC) can work together... But I can add this after the tutorial is published, this concern is minor and definitely not necessary to solve before publishing this tutorial.

Yep - that would be a fantastic cradle-to-grave workflow, could be used as a standalone workshop!

nomadscientist:
Such a great demonstration of why tags are awesome

nomadscientist:
Screenshot 2022-01-17 at 17 56 19
The tool should be able to run without selecting a phenotype target.

nomadscientist:
I can already see how it would be very (very) important for users to be able to specify components of the charts (for example, setting all of their y axes to 1). 


- [ ] Can these parameters be included?

nomadscientist:
Screenshot 2022-01-17 at 18 35 31
I don't understand how this chart is generated - where do we tick the box that says 'look in the beta cell proportions'?

nomadscientist:
On that same avenue, researchers will want to be able to input a phenotype (so imagine they were analysing two sets of samples, one with disease, and one without)


- [ ] Rather than stipulating a 'phenotype target', can they stipulate a metadata column?

- [ ] Where does 'TSD' come from? I thought all the bulk RNA-seq samples were from healthy participants... were some of them labelled TSD? - might need to change the label of the dataset from 'healthy'

mtekman:
> Screenshot 2022-01-17 at 17 56 19
> 
>     * [ ]  The tool should be able to run without selecting a phenotype target.

I'm not sure if that's possible.... but I'll dig into the code, maybe I can split it out.

> I can already see how it would be very (very) important for users to be able to specify components of the charts (for example, setting all of their y axes to 1).
> 
>     * [ ]  Can these parameters be included?

Likely! 

> Screenshot 2022-01-17 at 18 35 31
> I don't understand how this chart is generated - where do we tick the box that says 'look in the beta cell proportions'?

This is the phenotype factor (or metadata column) from the rna-seq data given in the first image you shared above. The user would write `hba1c` and this factor would be treated as the X-axis in the cell type proportion plot.


> On that same avenue, researchers will want to be able to input a phenotype (so imagine they were analysing two sets of samples, one with disease, and one without)
> 
>     * [ ]  Rather than stipulating a 'phenotype target', can they stipulate a metadata column?
> 

So I could likely change it so that they can just specify a column that specifically labels a sample as "diseased" or "healthy", but I like the idea of specifying a factor for it (like hba1c), because it changes the designation from a binary classification into a more linear one, where you can effectively set the threshold at which someone becomes diseased (in this case, when they have a hba1c level greater than 6.5%).

Also, not changing it means I don't need to mess with the internal code too much ;-P


>     * [ ]  Where does 'TSD' come from? I thought all the bulk RNA-seq samples were from healthy participants... were some of them labelled TSD? -
> 
>     * [ ]  Change label of dataset from 'healthy' (as the #scrna contains both healthy and TSD)

T2D is the disease factor given from the metadata column of the scrna-seq this time.

https://github.com/galaxyproject/training-material/blob/3e1552b0be5291fa429068e6f4b67b47edce68fa/topics/transcriptomics/tutorials/bulk-music/tutorial.md?plain=1#L218

This I believe *is* a binary classifier, but it could be anything the user wishes to use to compare a disease status in an scRNA dataset against another disease status in the RNA-seq dataset.

As I type this out, I realise how unclear that is

nomadscientist:
SO CLOSE!

The last set of 'construct expression objects' isn't working
url: https://humancellatlas.usegalaxy.eu/u/wendi.bacon/h/deconvolution-dendrogram-of-mouse-data

mtekman:
That large dataset shouldn't be used in the tutorial, so not sure why you downloaded it :P 

That being said, good that you tested it because it shows either a weakness in the code or a weirdness to the dataset.

The error message says:

```
Number of Samples between phenotypes and assays are not the same
```

and when I download the large files and test the first 6 sample names in each set, I see:

    > head(rownames(phenotypedata), 6)
    [1] "TGGTTCCGTCGGCTCA-2" "CGAGCCAAGCGTCAAG-4" "GAATGAAGTTTGGGCC-5"
    [4] "CTCGTACGTTGCCTCT-7" "TTCTCAATCCACGCAG-5" "CCTTCCCCATACCATG-4"
    > head(colnames(expressiondata), 6)
    [1] "TGGTTCCGTCGGCTCA.2" "CGAGCCAAGCGTCAAG.4" "GAATGAAGTTTGGGCC.5"
    [4] "CTCGTACGTTGCCTCT.7" "TTCTCAATCCACGCAG.5" "CCTTCCCCATACCATG.4"

so the difference is that phenotypes use a `-` and the expression set uses a `.` for the sample names, which is hilarious.

Here was me adding checks to the code for stringency, and it fails on their own major datasets.

I guess I could add a tickbox to the tool that disables the sample stringency test

mtekman:
Okay, so I have no time this week to do anything - but I've made a checklist for next week based on your review (and thanks!):

  - [ ] Add y-axis scale fix
  - [ ] Labelling fix of "healthy" in tutorial
  - [ ] Remove the need to specify a disease metacolumn phenotype in the bulk
    - Need to understand exactly whether the deconvolution can operate without it
  - [ ] Add "disable sample stringency check" for large datasets with slightly differing sample names

if I've missed anything, let me know!

mtekman:
The problem we're likely to run into is that the this MuSiC tool appears to be a specific RNA deconvolution tool to help understand a specific phenotype, and we're trying to generalise it as much as we can.

I'm not sure how easy this will be, and it took a lot of head-scratching to detangle it from the `hba1c` phenotype target. Future users might just need to create some kind of disease factor in their metadata for it to work (I realise how dumb that sounds)

nomadscientist:
> 

That is HILARIOUS!
Ok, but also, I am using the data linked in the tutorial -
Screenshot 2022-01-18 at 13 26 44
Is this the wrong dataset?

nomadscientist:
> 



> > Screenshot 2022-01-17 at 17 56 19
> > ```
> > * [ ]  The tool should be able to run without selecting a phenotype target.
> > ```
> 
> I'm not sure if that's possible.... but I'll dig into the code, maybe I can split it out.
> 
> > I can already see how it would be very (very) important for users to be able to specify components of the charts (for example, setting all of their y axes to 1).
> > ```
> > * [ ]  Can these parameters be included?
> > ```
> 
> Likely!
> 
> > Screenshot 2022-01-17 at 18 35 31
> > I don't understand how this chart is generated - where do we tick the box that says 'look in the beta cell proportions'?
> 
> This is the phenotype factor (or metadata column) from the rna-seq data given in the first image you shared above. The user would write `hba1c` and this factor would be treated as the X-axis in the cell type proportion plot.
Yes, but why beta cells? There are 6 different cell types, how was 'beta' specified? What if I wanted to know about the alpha cell types? In fact it's weird it doesn't spit out all the cell types, quite frankly.
> 
> > On that same avenue, researchers will want to be able to input a phenotype (so imagine they were analysing two sets of samples, one with disease, and one without)
> > ```
> > * [ ]  Rather than stipulating a 'phenotype target', can they stipulate a metadata column?
> > ```
> 
> So I could likely change it so that they can just specify a column that specifically labels a sample as "diseased" or "healthy", but I like the idea of specifying a factor for it (like hba1c), because it changes the designation from a binary classification into a more linear one, where you can effectively set the threshold at which someone becomes diseased (in this case, when they have a hba1c level greater than 6.5%).
> 
> Also, not changing it means I don't need to mess with the internal code too much ;-P
> 
Don't hate me, but it would be very valuable to do both...
> > ```
> > * [ ]  Where does 'TSD' come from? I thought all the bulk RNA-seq samples were from healthy participants... were some of them labelled TSD? -
> > 
> > * [ ]  Change label of dataset from 'healthy' (as the #scrna contains both healthy and TSD)
> > ```
> 
> T2D is the disease factor given from the metadata column of the scrna-seq this time.
So that makes sense, but why is it appearing in the bulk-deconvolution graph outputs? We never even labelled TSD anywhere, so it must be within the code to automate that. Was that 'sample disease group', in which case, is it just doing the phenotype target threshold? Or is it taking into account the scRNA TSD vs non-TSD data?
What if I wanted the triangles to be male vs female in a dataset, for instance? Would I type 'male' into 'Sample Disease Group'?

mtekman:
> > 
> 
> That is HILARIOUS! Ok, but also, I am using the data linked in the tutorial - Screenshot 2022-01-18 at 13 26 44 Is this the wrong dataset?

Ah I see, no it's correct I just apparently never tested it

> Yes, but why beta cells? There are 6 different cell types, how was 'beta' specified? What if I wanted to know about the alpha cell types? In fact it's weird it doesn't spit out all the cell types, quite frankly.

Ah... Hmm. Yes. Good question....

I'm looking through the code now, and I genuinely don't know.... it could just be that the title is misleading, and that it's showing UserSetFactorX vs UserSetFactorY and I've misinterpreted it.... Hmmm.... I'm not sure and will have to dig a bit

> Don't hate me, but it would be very valuable to do both...

Lol no, I've already torn the code apart once, I can do it again, and yes I can see how it would be more practical.

> So that makes sense, but why is it appearing in the bulk-deconvolution graph outputs? We never even labelled TSD anywhere, so it must be within the code to automate that. Was that 'sample disease group', in which case, is it just doing the phenotype target threshold? 

Exactly it's doing phenotype target threshold and determining what is healthy and what is not. 

> Or is it taking into account the scRNA TSD vs non-TSD data?

This I don't know....

>  What if I wanted the triangles to be male vs female in a dataset, for instance? Would I type 'male' into 'Sample Disease Group'?

Yes, and give it instead of `hba1c` as the disease factor, but a column maybe named  `gender` in the phenotype data. (Though I should change the label "Disease" to something less accusatory).

*****

I'm actually presenting this tomorrow morning in a lab meeting, and I think I'll take a screenshot of this convo as a prime example of how the revision process works, and what seems completely normal to the devs is not bounded in reality by the scientists :laughing:

nomadscientist:
Nice! Tell them I'm a useful member of society too

mtekman:
You shall be introduced as ranting lunatic I met on the web :P

nomadscientist:
Woo! Awesome, ping me when these bits are sorted and then I'll finish testing the tutorial and I think I can even add myself as a reviewer on Github, how exciting!

mtekman:
Will do! ETA: sometime towards the end of next week

nomadscientist:
> The problem we're likely to run into is that the this MuSiC tool appears to be a specific RNA deconvolution tool to help understand a specific phenotype, and we're trying to generalise it as much as we can.
> 
> I'm not sure how easy this will be, and it took a lot of head-scratching to detangle it from the `hba1c` phenotype target. Future users might just need to create some kind of disease factor in their metadata for it to work (I realise how dumb that sounds)

You're bang on with this, and I think this is also just underlying a general problem of 'Wow isn't my tool awesome' "But it needs major rewrites to work for anyone else?" 'That's fine, they can figure it out..."

Maybe the scale of rewrites is good, because then this counts as even more of an output for you and can generate even more impact. MMusic :)

mtekman:
@nomadscientist I rebased everything just to keep this PR up to date with the main branch, since it's a long WIP.

All your edits and commits are still there, but if you make any future edits -- please pull everything first

bgruening:
@mtekman @nomadscientist what an epic work, very cool. Not sure if you are aware, but we can get this merged but hidden by default. This way you can already use a link to point to the tutorial, but people browsing the GTN would not see it. Maybe that is useful for you here.

This one for example is such a tutorial: https://training.galaxyproject.org/training-material/topics/assembly/tutorials/vgp_genome_assembly/tutorial.html

mtekman:
@bgruening nice - that looks optimal for us, since we're including it in the GTN and will need to link to it soon. How can we get this merged and hidden?

@nomadscientist  - so I've updated the tools and trainings, but because I re-based it's hard to see exactly what I changed unless you click on each and every commit after yours(!) Instead, I'm hosting both the tools and the training on a public server so you can go through it without having to reset and rebase your local history (I've shared the link in the whatsapp).

Things that have changed:

- Added a fixed y-axis option to the scatter plots
- Made the "disease" stuff now optional for the deconvolution. People can now just supply a bulk set and a scRNA set and hit execute and it should just work without any other fiddling.
- Fixed the "where is the beta cell specified?" issue you mentioned before - which I previously hand-waved away with some nonsense about factors and levels due to me being misled by the code -  but after looking at it more deeply code, yes you should definitely be able to specify exactly which cell type you want to explore when measuring a bulk disease phenotype.

Hopefully these changes are reflected both in the tools and the updated training text. Please can you run through it again and see if it makes sense still?

bgruening:
`enable: false` is your friend, see https://github.com/galaxyproject/training-material/blob/main/topics/assembly/tutorials/vgp_genome_assembly/tutorial.md

nomadscientist:
Wait, how can I edit it now???

mtekman:
@nomadscientist Wew, back at a computer. So the live site is at that website I shared, but to review it you need to do inline comments on Github here:

1. Go here: https://github.com/galaxyproject/training-material/pull/2790/files
2. Scroll down to the file: "topics/transcriptomics/tutorials/bulk-music/tutorial.md"
3. (It might say "Load Diff" and show nothing for the file. Click on the "Load Diff" to load the whole file
4. Every line in that file will start with "+" symbol, which you can click on to add a comment.
5. If the comment is part of a larger review, then start a review, then add comments, and when finished submit the review with a judgement ("accepted" "needs changes" etc.)

nomadscientist:
1) OK, I think there are a few (very minor) text things that I can fix when this is merged - they in no way take away from the learning experience or quality of the tutorial, so I think it should just be merged!
2) I LOVE how you're rejigged the Music tool. It's much clearer and can now be adapted in a heap of ways, very cool.
3) The drag & drop is no longer necessary, yay! I hope this works all the time and not just on the Galaxy you shared...
4) For the scRNA T2D - is this comparing the expression in the T2D cells with the bulk, or is it just saying anything above 6.5 we're counting as T2D? I'm not sure how the relationship works between these metadata and how it analyses the cell types. OH or is it saying that it identified bulk samples that likely correspond to the T2D samples, AND AS WELL plotted this along hba1c, and shows a direct correlation, so therefore clearly the deconvolution is working well at identifying 'likely disease'? Because if that's how it works, THAT is cool.
5) If somebody had a bunch of samples from a knockout and a wildtype and wanted to compare cell proportions in the two samples, how would they do this? Would they select this as the phenotype factor? I get that it would be calculated in the log of MuSiC fitting, but I'm trying to figure out how they'd display this.
5b) Actually yeah, we need to be able to distinguish this in the jitter plots side-by-side
6) Can we input multiple into the 'scRNA phenotype cell target'? i.e. beta,gamma (I tried this and we can't, but damn it would be helpful)
7) There should be a toggle to tick 'DO NOT SHOW NNLS' because it's unnecessary for use. Much more useful would be for the plot to show wildtype where MuSiC is and (knockout or similar group) where NNLS is. [I know this is a big change :( 5b)
8) Other than that, I finished the tutorial and it's awesome!
@mtekman

mtekman:
> 4. For the scRNA T2D - is this comparing the expression in the T2D cells with the bulk, or is it just saying anything above 6.5 we're counting as T2D?

So the 6.5 threshold is measured against the bulk RNA dataset which contains this hba1c factor (see Dataset 2 in history), and this bulk dataset has 89 samples, all of which are plotted in the graph. 

MuSiC, using scRNA data to work out cell proportions, then tells us in that graph what the percentage of beta cells is for each of the 89 bulk RNA samples, and plots this value against the hba1c factor.

The expectation from the literature/caption is that we see a decrease in beta cells as the hba1c bulk RNA factor increases. Before using MuSiC, we had no idea what the proportion of beta cells in that sample was, but now we do!

>  OH or is it saying that it identified bulk samples that likely correspond to the T2D samples

So that is shown in Figure 9 with the heatmap, with RNA samples as rows and scRNA-derived cell types as columns.

I realise that if someone asked "what is the cell composition of my bulk data?" a pie chart might be easier visualize.... but as we see in that heatmap, there are a lot (89) of samples in this case and so it's not clear whether we need to generate 89 seperate pie charts or not.

> AND AS WELL plotted this along hba1c, and shows a direct correlation, so therefore clearly the deconvolution is working well at identifying 'likely disease'? Because if that's how it works, THAT is cool.

Yes and Yes :D

> 5. If somebody had a bunch of samples from a knockout and a wildtype and wanted to compare cell proportions in the two samples, how would they do this?

I believe they would run the program twice, both times using the same scRNA dataset (to have the same cell profiling) but different bulk RNA dataset samples.

> Would they select this as the phenotype factor? I get that it would be calculated in the log of MuSiC fitting.

There would be no need to set a phenotype factor, since a disease would not be the thing being compared there. Under, "Show proportions of a disease factor?", "No" should be selected

Screenshot 2022-02-01 at 13-50-10 Galaxy Configured by Planemo

> but I'm trying to figure out how they'd display this.

They'd literally compare two separate heatmaps me thinks. If they have mutant and WT in the same bulk RNA dataset, then they need to split this dataset into two.

> 6. Can we input multiple into the 'scRNA phenotype cell target'? i.e. beta,gamma (I tried this and we can't, but damn it would be helpful)

I believe so, it would take a few days though to implement.

> 7. There should be a toggle to tick 'DO NOT SHOW NNLS' because it's unnecessary for use. 

This I can easily remove, but I keep it just so that users can validate their cell type proportions (via MuSiC) with another older method (NNLS). It's part of the original training, so I figured I should not change it. What do you think? Give users the option? 

> Much more useful would be for the plot to show wildtype where MuSiC is and (knockout or similar group) where NNLS is. [I know this is a big change :( 5b)

So I know how I would go about doing this, but this will definitely be tricky to implement for two reasons:

1. The plots will definitely require some TLC to get right
2. It would deviate heavily from the original training material "vignette" that this training was based on. We would have to write a completely new one for the use case of comparing mutant and wildtype. And maybe we should?

  If so, we would use the Mutant vs WT training first, and then lead them onto this training where we look at how a disease factor that correlated with a single cell type can be seen in bulk Data.


*******

Also, we're running out of time!

mtekman:
More thoughts:
* If we were to do a Mutant vs WildType comparison, maybe I should make another tool called "MuSiC Compare" which takes three (or four) files as input: WT bulk, Mutant bulk,  WT scRNA, (and optionally a Mutant scRNA).
  * The output of the tool would just be two heatmaps side by side, and two pie charts which summarize the cell compositions. Users can then use the main MuSiC tool if they want to probe disease phenotypes further.

* However, the above tool limits the analysis only to a pairwise comparison. If there are multiple condition groups and we wish to see the cell composition of each group? 
  * It gets tricky, do we do yet another tool for this use case or do we try to merge it with the above tool that takes 4 inputs? 

I'm not complaining here btw, I really think we could flesh this thing out. We would just need a new training for it, and that likely won't be presented in the GTN

nomadscientist:
> > 7\. For the scRNA T2D - is this comparing the expression in the T2D cells with the bulk, or is it just saying anything above 6.5 we're counting as T2D?
> 
> So the 6.5 threshold is measured against the bulk RNA dataset which contains this hba1c factor (see Dataset 2 in history), and this bulk dataset has 89 samples, all of which are plotted in the graph (Figure 8).
> 
> MuSiC, using scRNA data to work out cell proportions, then tells us in that graph what the percentage of beta cells is for each of the 89 bulk RNA samples, and plots this value against the hba1c factor.
> 
> The expectation from the literature/caption is that we see a decrease in beta cells as the hba1c bulk RNA factor increases. Before using MuSiC, we had no idea what the proportion of beta cells in that sample was, but now we do!
> 
> > OH or is it saying that it identified bulk samples that likely correspond to the T2D samples
> 
> So that is shown in Figure 9 with the heatmap, with RNA samples as rows and scRNA-derived cell types as columns.
> 

My question was more: how they determine what gets the triangle (i.e. the T2D)? Are they calling those triangles purely from the hba1c level, or are they from the single cell dataset and both the scRNA and the bulk are being plotted at the same time?

> I realise that if someone asked "what is the cell composition of my bulk data?" a pie chart might be easier visualize.... but as we see in that heatmap, there are a lot (89) of samples in this case and so it's not clear whether we need to generate 89 seperate pie charts or not.
> 
> > AND AS WELL plotted this along hba1c, and shows a direct correlation, so therefore clearly the deconvolution is working well at identifying 'likely disease'? Because if that's how it works, THAT is cool.
> 
> Yes and Yes :D
> 

Pie chart or stacked bar graph, and I'd argue just one averaged (per phenotype factor of interest). I *think* we can make them with the matrix it spits out anyway (i.e. it has samples x proportions, so I could do this in excel in a jiffy, so it's workable, but obvs nice if it is in Galaxy, but I think we can agree this is a problem/workflow for another occasion!)

> > 5. If somebody had a bunch of samples from a knockout and a wildtype and wanted to compare cell proportions in the two samples, how would they do this?
> 
> I believe they would run the program twice, both times using the same scRNA dataset (to have the same cell profiling) but different bulk RNA dataset samples.
> 

Nuts! Because then then can't coplot it (although again, they could take raw into excel and plot it as described above)

> > Would they select this as the phenotype factor? I get that it would be calculated in the log of MuSiC fitting.
> 
> There would be no need to set a phenotype factor, since a disease would not be the thing being compared there. Under, "Show proportions of a disease factor?", "No" should be selected
> 
> Screenshot 2022-02-01 at 13-50-10 Galaxy Configured by Planemo
> 
> > but I'm trying to figure out how they'd display this.
> 
> They'd literally compare two separate heatmaps me thinks. If they have mutant and WT in the same bulk RNA dataset, then they need to split this dataset into two.
> 
> > 6. Can we input multiple into the 'scRNA phenotype cell target'? i.e. beta,gamma (I tried this and we can't, but damn it would be helpful)
> 
> I believe so, it would take a few days though to implement.
> 
> > 7. There should be a toggle to tick 'DO NOT SHOW NNLS' because it's unnecessary for use.
> 
> This I can easily remove, but I keep it just so that users can validate their cell type proportions (via MuSiC) with another older method (NNLS). It's part of the original training, so I figured I should not change it. What do you think? Give users the option?

Definitely give users the option. Tbh, users won't give a damn. People analysing algorithms will (but they're less likely to be doing this on Galaxy anyway)

> 
> > Much more useful would be for the plot to show wildtype where MuSiC is and (knockout or similar group) where NNLS is. [I know this is a big change :( 5b)
> 
> So I know how I would go about doing this, but this will definitely be tricky to implement for two reasons:
> 
> 1. The plots will definitely require some TLC to get right
> 2. It would deviate heavily from the original training material "vignette" that this training was based on. We would have to write a completely new one for the use case of comparing mutant and wildtype. And maybe we should?
> 

1) :( 
2) Actually we wouldn't need to! The second section already has control and a phenotype in its 10 bulk samples, 6 of normal and 4 for the other thing, so we could keep the vignette (but SUBSTANTIALLY improve it and make it usable)
Again, I think we can argue there are current workarounds, but obviously for usability, this would be great to do. This would be the difference, for example, of us being able to include it in an undergrad capstone project or not. The more times you move data from one application to another, the more change there is of a student losing the plot (pardon the pun!)

> If so, we would use the Mutant vs WT training first, and then lead them onto this training where we look at how a disease factor that correlated with a single cell type can be seen in bulk Data.
> 

Actually, that's a genius restructure and totally would make more sense.

> Also, we're running out of time!

How about we go with it as is (with the two toggles described above fixed) for the Smorgasbord (I mean, the original vignette was published and we're just repeating that with some very very helpful toggles for users), so why not? Then we have a bit more time and could aim for GCC for an updated version incorporating these fixes? It took me about 3-4 times doing the tutorial to realise what we needed (you can call me a focus group of 4!), and we might get more feedback during the Smorgasbord if we're lucky, so then we take that and make the 2.0 version for GCC?

> > 7\. For the scRNA T2D - is this comparing the expression in the T2D cells with the bulk, or is it just saying anything above 6.5 we're counting as T2D?
> 
> So the 6.5 threshold is measured against the bulk RNA dataset which contains this hba1c factor (see Dataset 2 in history), and this bulk dataset has 89 samples, all of which are plotted in the graph (Figure 8).
> 
> MuSiC, using scRNA data to work out cell proportions, then tells us in that graph what the percentage of beta cells is for each of the 89 bulk RNA samples, and plots this value against the hba1c factor.
> 
> The expectation from the literature/caption is that we see a decrease in beta cells as the hba1c bulk RNA factor increases. Before using MuSiC, we had no idea what the proportion of beta cells in that sample was, but now we do!
> 
> > OH or is it saying that it identified bulk samples that likely correspond to the T2D samples
> 
> So that is shown in Figure 9 with the heatmap, with RNA samples as rows and scRNA-derived cell types as columns.
> 
> I realise that if someone asked "what is the cell composition of my bulk data?" a pie chart might be easier visualize.... but as we see in that heatmap, there are a lot (89) of samples in this case and so it's not clear whether we need to generate 89 seperate pie charts or not.
> 
> > AND AS WELL plotted this along hba1c, and shows a direct correlation, so therefore clearly the deconvolution is working well at identifying 'likely disease'? Because if that's how it works, THAT is cool.
> 
> Yes and Yes :D
> 
> > 5. If somebody had a bunch of samples from a knockout and a wildtype and wanted to compare cell proportions in the two samples, how would they do this?
> 
> I believe they would run the program twice, both times using the same scRNA dataset (to have the same cell profiling) but different bulk RNA dataset samples.
> 
> > Would they select this as the phenotype factor? I get that it would be calculated in the log of MuSiC fitting.
> 
> There would be no need to set a phenotype factor, since a disease would not be the thing being compared there. Under, "Show proportions of a disease factor?", "No" should be selected
> 
> Screenshot 2022-02-01 at 13-50-10 Galaxy Configured by Planemo
> 
> > but I'm trying to figure out how they'd display this.
> 
> They'd literally compare two separate heatmaps me thinks. If they have mutant and WT in the same bulk RNA dataset, then they need to split this dataset into two.
> 
> > 6. Can we input multiple into the 'scRNA phenotype cell target'? i.e. beta,gamma (I tried this and we can't, but damn it would be helpful)
> 
> I believe so, it would take a few days though to implement.
> 
> > 7. There should be a toggle to tick 'DO NOT SHOW NNLS' because it's unnecessary for use.
> 
> This I can easily remove, but I keep it just so that users can validate their cell type proportions (via MuSiC) with another older method (NNLS). It's part of the original training, so I figured I should not change it. What do you think? Give users the option?
> 
> > Much more useful would be for the plot to show wildtype where MuSiC is and (knockout or similar group) where NNLS is. [I know this is a big change :( 5b)
> 
> So I know how I would go about doing this, but this will definitely be tricky to implement for two reasons:
> 
> 1. The plots will definitely require some TLC to get right
> 2. It would deviate heavily from the original training material "vignette" that this training was based on. We would have to write a completely new one for the use case of comparing mutant and wildtype. And maybe we should?
> 
> If so, we would use the Mutant vs WT training first, and then lead them onto this training where we look at how a disease factor that correlated with a single cell type can be seen in bulk Data.
> 
> Also, we're running out of time!

nomadscientist:
> > 7\. For the scRNA T2D - is this comparing the expression in the T2D cells with the bulk, or is it just saying anything above 6.5 we're counting as T2D?
> 
> So the 6.5 threshold is measured against the bulk RNA dataset which contains this hba1c factor (see Dataset 2 in history), and this bulk dataset has 89 samples, all of which are plotted in the graph (Figure 8).
> 
> MuSiC, using scRNA data to work out cell proportions, then tells us in that graph what the percentage of beta cells is for each of the 89 bulk RNA samples, and plots this value against the hba1c factor.
> 
> The expectation from the literature/caption is that we see a decrease in beta cells as the hba1c bulk RNA factor increases. Before using MuSiC, we had no idea what the proportion of beta cells in that sample was, but now we do!
> 
> > OH or is it saying that it identified bulk samples that likely correspond to the T2D samples
> 
> So that is shown in Figure 9 with the heatmap, with RNA samples as rows and scRNA-derived cell types as columns.
> 

My question was more: how they determine what gets the triangle (i.e. the T2D)? Are they calling those triangles purely from the hba1c level, or are they from the single cell dataset and both the scRNA and the bulk are being plotted at the same time?

> I realise that if someone asked "what is the cell composition of my bulk data?" a pie chart might be easier visualize.... but as we see in that heatmap, there are a lot (89) of samples in this case and so it's not clear whether we need to generate 89 seperate pie charts or not.
> 
> > AND AS WELL plotted this along hba1c, and shows a direct correlation, so therefore clearly the deconvolution is working well at identifying 'likely disease'? Because if that's how it works, THAT is cool.
> 
> Yes and Yes :D
> 

Pie chart or stacked bar graph, and I'd argue just one averaged (per phenotype factor of interest). I *think* we can make them with the matrix it spits out anyway (i.e. it has samples x proportions, so I could do this in excel in a jiffy, so it's workable, but obvs nice if it is in Galaxy, but I think we can agree this is a problem/workflow for another occasion!)

> > 5. If somebody had a bunch of samples from a knockout and a wildtype and wanted to compare cell proportions in the two samples, how would they do this?
> 
> I believe they would run the program twice, both times using the same scRNA dataset (to have the same cell profiling) but different bulk RNA dataset samples.
> 

Nuts! Because then then can't coplot it (although again, they could take raw into excel and plot it as described above)

> > Would they select this as the phenotype factor? I get that it would be calculated in the log of MuSiC fitting.
> 
> There would be no need to set a phenotype factor, since a disease would not be the thing being compared there. Under, "Show proportions of a disease factor?", "No" should be selected
> 
> Screenshot 2022-02-01 at 13-50-10 Galaxy Configured by Planemo
> 
> > but I'm trying to figure out how they'd display this.
> 
> They'd literally compare two separate heatmaps me thinks. If they have mutant and WT in the same bulk RNA dataset, then they need to split this dataset into two.
> 
> > 6. Can we input multiple into the 'scRNA phenotype cell target'? i.e. beta,gamma (I tried this and we can't, but damn it would be helpful)
> 
> I believe so, it would take a few days though to implement.
> 
> > 7. There should be a toggle to tick 'DO NOT SHOW NNLS' because it's unnecessary for use.
> 
> This I can easily remove, but I keep it just so that users can validate their cell type proportions (via MuSiC) with another older method (NNLS). It's part of the original training, so I figured I should not change it. What do you think? Give users the option?

Definitely give users the option. Tbh, users won't give a damn. People analysing algorithms will (but they're less likely to be doing this on Galaxy anyway)

> 
> > Much more useful would be for the plot to show wildtype where MuSiC is and (knockout or similar group) where NNLS is. [I know this is a big change :( 5b)
> 
> So I know how I would go about doing this, but this will definitely be tricky to implement for two reasons:
> 
> 1. The plots will definitely require some TLC to get right
> 2. It would deviate heavily from the original training material "vignette" that this training was based on. We would have to write a completely new one for the use case of comparing mutant and wildtype. And maybe we should?
> 

1) :( 
2) Actually we wouldn't need to! The second section already has control and a phenotype in its 10 bulk samples, 6 of normal and 4 for the other thing, so we could keep the vignette (but SUBSTANTIALLY improve it and make it usable)
Again, I think we can argue there are current workarounds, but obviously for usability, this would be great to do. This would be the difference, for example, of us being able to include it in an undergrad capstone project or not. The more times you move data from one application to another, the more change there is of a student losing the plot (pardon the pun!)

> If so, we would use the Mutant vs WT training first, and then lead them onto this training where we look at how a disease factor that correlated with a single cell type can be seen in bulk Data.
> 

Actually, that's a genius restructure and totally would make more sense.

> Also, we're running out of time!

How about we go with it as is (with the two toggles described above fixed) for the Smorgasbord (I mean, the original vignette was published and we're just repeating that with some very very helpful toggles for users), so why not? Then we have a bit more time and could aim for GCC for an updated version incorporating these fixes? It took me about 3-4 times doing the tutorial to realise what we needed (you can call me a focus group of 4!), and we might get more feedback during the Smorgasbord if we're lucky, so then we take that and make the 2.0 version for GCC?

> > 7\. For the scRNA T2D - is this comparing the expression in the T2D cells with the bulk, or is it just saying anything above 6.5 we're counting as T2D?
> 
> So the 6.5 threshold is measured against the bulk RNA dataset which contains this hba1c factor (see Dataset 2 in history), and this bulk dataset has 89 samples, all of which are plotted in the graph (Figure 8).
> 
> MuSiC, using scRNA data to work out cell proportions, then tells us in that graph what the percentage of beta cells is for each of the 89 bulk RNA samples, and plots this value against the hba1c factor.
> 
> The expectation from the literature/caption is that we see a decrease in beta cells as the hba1c bulk RNA factor increases. Before using MuSiC, we had no idea what the proportion of beta cells in that sample was, but now we do!
> 
> > OH or is it saying that it identified bulk samples that likely correspond to the T2D samples
> 
> So that is shown in Figure 9 with the heatmap, with RNA samples as rows and scRNA-derived cell types as columns.
> 
> I realise that if someone asked "what is the cell composition of my bulk data?" a pie chart might be easier visualize.... but as we see in that heatmap, there are a lot (89) of samples in this case and so it's not clear whether we need to generate 89 seperate pie charts or not.
> 
> > AND AS WELL plotted this along hba1c, and shows a direct correlation, so therefore clearly the deconvolution is working well at identifying 'likely disease'? Because if that's how it works, THAT is cool.
> 
> Yes and Yes :D
> 
> > 5. If somebody had a bunch of samples from a knockout and a wildtype and wanted to compare cell proportions in the two samples, how would they do this?
> 
> I believe they would run the program twice, both times using the same scRNA dataset (to have the same cell profiling) but different bulk RNA dataset samples.
> 
> > Would they select this as the phenotype factor? I get that it would be calculated in the log of MuSiC fitting.
> 
> There would be no need to set a phenotype factor, since a disease would not be the thing being compared there. Under, "Show proportions of a disease factor?", "No" should be selected
> 
> Screenshot 2022-02-01 at 13-50-10 Galaxy Configured by Planemo
> 
> > but I'm trying to figure out how they'd display this.
> 
> They'd literally compare two separate heatmaps me thinks. If they have mutant and WT in the same bulk RNA dataset, then they need to split this dataset into two.
> 
> > 6. Can we input multiple into the 'scRNA phenotype cell target'? i.e. beta,gamma (I tried this and we can't, but damn it would be helpful)
> 
> I believe so, it would take a few days though to implement.
> 
> > 7. There should be a toggle to tick 'DO NOT SHOW NNLS' because it's unnecessary for use.
> 
> This I can easily remove, but I keep it just so that users can validate their cell type proportions (via MuSiC) with another older method (NNLS). It's part of the original training, so I figured I should not change it. What do you think? Give users the option?
> 
> > Much more useful would be for the plot to show wildtype where MuSiC is and (knockout or similar group) where NNLS is. [I know this is a big change :( 5b)
> 
> So I know how I would go about doing this, but this will definitely be tricky to implement for two reasons:
> 
> 1. The plots will definitely require some TLC to get right
> 2. It would deviate heavily from the original training material "vignette" that this training was based on. We would have to write a completely new one for the use case of comparing mutant and wildtype. And maybe we should?
> 
> If so, we would use the Mutant vs WT training first, and then lead them onto this training where we look at how a disease factor that correlated with a single cell type can be seen in bulk Data.
> 
> Also, we're running out of time!

nomadscientist:
> More thoughts:
> 
> * If we were to do a Mutant vs WildType comparison, maybe I should make another tool called "MuSiC Compare" which takes three (or four) files as input: WT bulk, Mutant bulk,  WT scRNA, (and optionally a Mutant scRNA).
>   
>   * The output of the tool would just be two heatmaps side by side, and two pie charts which summarize the cell compositions. Users can then use the main MuSiC tool if they want to probe disease phenotypes further.
> * However, the above tool limits the analysis only to a pairwise comparison. If there are multiple condition groups and we wish to see the cell composition of each group?
>   
>   * It gets tricky, do we do yet another tool for this use case or do we try to merge it with the above tool that takes 4 inputs?
> 
> I'm not complaining here btw, I really think we could flesh this thing out. We would just need a new training for it, and that likely won't be presented in the GTN

Ok this is actually blooming genius and would be VERY popular. This is what I'd imagined the tool did initially... 
Also output a matrix and the chi square log so they could download in excel (if they wanted) and see the probabilities! [but this calculated based on combining the bulk into a single file, idk if it would also work if you did it separately...]
But yes, you need it to work for X # of phenotypes. Although I *think* (if I have understood correctly how the T2D worked, which is questionable) that you could have the input scRNA-seq contain all of the phenotypes in one file, you wouldn't need them separated.... right?

I could even see this working in a crispr scenario of phenotypes which could be 100s... actually probably not, idk if MuSiC works for that level of phenotypic difference and there is probably something made for it that's better (she says...)

mtekman:
> My question was more: how they determine what gets the triangle (i.e. the T2D)? Are they calling those triangles purely from the hba1c level, or are they from the single cell dataset and both the scRNA and the bulk are being plotted at the same time?

Oh right, yeah that's determined purely from the hba1c level. AFAIK, the only datapoints in that plot are bulk (with the y-axis being derived from single-cell)

> Pie chart or stacked bar graph, and I'd argue just one averaged (per phenotype factor of interest). I think we can make them with the matrix it spits out anyway (i.e. it has samples x proportions, so I could do this in excel in a jiffy, so it's workable, but obvs nice if it is in Galaxy, but I think we can agree this is a problem/workflow for another occasion!)

Exactly, averaged per phenotype, and yep we'd be using that exact matrix given by "<algorithm> Estimated Proportions of Cell Types".

> Definitely give users the option. Tbh, users won't give a damn. People analysing algorithms will (but they're less likely to be doing this on Galaxy anyway).

Okay, will do -- it's an easy change I think,

> The second section already has control and a phenotype in its 10 bulk samples, 6 of normal and 4 for the other thing, so we could keep the vignette (but SUBSTANTIALLY improve it and make it usable)

Oh good spot, we could keep the existing datasets then when develop it.

> Again, I think we can argue there are current workarounds, but obviously for usability, this would be great to do.

Okay, the rest of this week I don't have much to do, and could likely implement such a tool using MuSiC as a backend.

> How about we go with it as is (with the two toggles described above fixed) for the Smorgasbord (I mean, the original vignette was published and we're just repeating that with some very very helpful toggles for users), so why not? Then we have a bit more time and could aim for GCC for an updated version incorporating these fixes? 

So I'm *sure* I could hammer out a comparison tool by the end of the week, but the question would be whether we'd be able to hammer out a decent training for it before the training video submission date on the 28th.

> Also output a matrix and the chi square log so they could download in excel (if they wanted) and see the probabilities! 

I like it, but a chi-square comparing what to what though? Cell type proportions in each bulk phenotype (e.g. proportion of alpha cells in mutant vs proportion of alpha cells in WT, etc.)?

> But yes, you need it to work for X # of phenotypes

Okay, so then if multiple comparisons are a common use-case, then I should make the "MuSiC Compare" tool handle multiple phenotypes as the default case.

> you could have the input scRNA-seq contain all of the phenotypes in one file, you wouldn't need them separated.... right?

Technically no, the user would just specify a column name in the phenotypes file that seperates sample groups.... but what's more common:

 1. A user has a single bulk RNA dataset containing all phenotypes and wants to compare everything within.
 2. A user takes a bunch of seperate bulk datasets from various papers and wants to compare everything across.

I think 2 would be more common, though I'm not actually sure.

> It took me about 3-4 times doing the tutorial to realize what we needed (you can call me a focus group of 4!), and we might get more feedback during the Smorgasbord if we're lucky, so then we take that and make the 2.0 version for GCC?

Definitely, but I think we should still try in the next few weeks for the Mutant vs WT training, just while we've still got momentum.

********

TO DO:
- [ ] Music, NNLS toggle

TO TRY:
- [ ] Music Compare tool
- [ ] Music Compare training
- [ ]

mtekman:
> Okay, so then if multiple comparisons are a common use-case, then I should make the "MuSiC Compare" tool handle multiple phenotypes as the default case.

Idea for MuSiC Compare:

* Tool takes N bulk RNA datasets as input and 1 scRNA dataset.
   * Each bulk is deconvoluted from this one scRNA, and then bulk compositions are compared.
* Tool takes N bulk RNA datasets as input and N scRNA datasets
   * Each bulk RNA dataset is compared to the corresponding scRNA dataset, and then bulk compositions are compared.
* Comparisons output: detailed heatmap of cell type proportion for each bulk sample, and an averaged heatmap or pie-chart for each condition.

nomadscientist:
If stats are a bit sketchy (mine too, although I do have a stats professor here I could ask if it came to it), then outputting a 'general info' file with:
average per each cell type for each phenotype + stdev + sterror + # samples each
Shit it's an ANOVA. That's the test I think... Problem for another day!
We gives people the data then people can choose what calculations they want!

I think adding that tool and making the training for it will be doable. If you can make the tool this week, I can commit to the training tweaks next week (if it's in a github file I can branch, which it currently is not) and make the video at the end. Worst comes to worst, I can take the current tutorial and roll with that as a back-up.

mtekman:
I have a partially working "Music Compare" solution working on that server. The general way it works is:

* Inputs (scRNA dataset "Groups" that affect Bulk Datasets):

   * scRNA1 → Bulk1, Bulk2, Bulk3,...
   * scRNA2 → Bulk4,...
   * scRNA3 → Bulk5, Bulk6,...
   *  ...

* Outputs (heatmaps and pie/bar charts):
   * DONE
   * [X] Heatmaps
      * [X] Individual heatmaps (e.g. Bulk1 from scRNA1, Bulk2 from scRNA1, ..., Bulk6 from scRNA3)
         * Rows are individual samples in a Bulk dataset, Columns are Cells 
      * [X] Grouped heatmaps (e.g. Bulk1 + Bulk2 + Bulk3 all shown together,  Bulk4 by itself, Bulk5 + Bulk6 together)
         * Rows are cells (shared across all heatmaps), Columns are samples in each bulk grouped by bulk.
         * (I had to invert axes from the individual heatmaps because the Y-axis appears to be the common factor)
    * TODO
   * [ ] Summarized:
        * [ ] Heatmaps by Sample (e.g. all possible cell types across all scRNA datasets against all possible samples across all bulk datasets)
        * [ ] Heatmaps by Bulk Dataset (e.g. all possible cell types across all scRNA datasets against all Bulk datasets summarizing across all samples.
        * [ ] Pie/Bar Chart by Bulk Dataset (e.g. represent cell types in a given scRNA group across all bulk datasets in that group)
        * [ ]  Others...

Some issues:
* How to summarize multiple samples in a bulk dataset? For example if SampleA in Bulk1 is 50% stem cells and 50% PT cells, but SampleB in Bulk1 is 25% stem cells and 75% PT cells, what do I say about the overall composition of Bulk1?
    * Solution1: Average the percentages across cell types, so that Stem is (50 + 25 / 200), and PT is (50 + 75 / 200)
    * Solution2: Scale the raw reads to these proportions, so that if SampleA has 10000 reads and SampleB has 5000 reads, then I can get very real numbers for how many reads in total are Stem and PT.

ETA for completion: Monday evening(ish)

nomadscientist:
@mtekman Amazing stuff! 

How to summarize multiple samples in a bulk dataset? For example if SampleA in Bulk1 is 50% stem cells and 50% PT cells, but SampleB in Bulk1 is 25% stem cells and 75% PT cells, what do I say about the overall composition of Bulk1?

    Solution1: Average the percentages across cell types, so that Stem is (50 + 25 / 200), and PT is (50 + 75 / 200)

YES! This is where you also include n values (samples) + standard deviation + st error + median + 25/75 percentiles. Then people can do whichever tests are best.

    Solution2: Scale the raw reads to these proportions, so that if SampleA has 10000 reads and SampleB has 5000 reads, then I can get very real numbers for how many reads in total are Stem and PT.

This is SUCH an interesting idea. Does more reads mean more cells? I would think not? In which case, don't do this? Or they'd have to input how many cells were in each RNA-seq library, which I think is probably usually unknown so don't do it.

mtekman:
> YES! This is where you also include n values (samples) + standard deviation + st error + median + 25/75 percentiles. Then people can do whichever tests are best.

Ah of couse - and we can do boxplots to show the spread too ! Okay, I can implement that

mtekman:
> This is SUCH an interesting idea. Does more reads mean more cells? I would think not? In which case, don't do this? Or they'd have to input how many cells were in each RNA-seq library, which I think is probably usually unknown so don't do it.

I'm not sure myself, but I think its just a way to scale the samples in a dataset, since not all samples are equal

nomadscientist:
>I'm not sure myself, but I think its just a way to scale the samples in a dataset, since not all samples are equal. If one sample has 90% of the reads in a dataset and represents 90% of the cell types discovered in that dataset, but is only represented once, then the other samples in that same dataset which make up only 10% of the discoverer cell types would be overrepresented, no?

>Edit: But then again, samples in the same dataset should be normalised against one another to make a more fair comparison right? Ah god I don't know

No, this is ok, this is normal. We don't over-represent one vs the other. Weight them equally. If there are weird datasets in there, that's the job of the people analysing them to deal with (i.e. is this an outlier? Should we analyse it on its own?) etc. That's why you have a box plot WITH dots representing each datapoint (like a violin plot).

mtekman:
> Weight them equally. If there are weird datasets in there, that's the job of the people analysing them to deal with (i.e. is this an outlier? Should we analyse it on its own?) etc. That's why you have a box plot WITH dots representing each datapoint (like a violin plot).

This makes sense, since if we scale the cell type proportions by the reads then we get a box plot like this:

test

which gives the false impression that PT cells have a higher percentage in one bulk set (the "L1" colors) compared to another, but all three datasets are actually completely the same just scaled slightly differently (I used the same dataset three times for testing purposes).

mtekman:
Also, slightly related question - what looks better: 

* Samples in the Y-axis, and colouring by cell type?
  test2 

* Cell types in the Y-axis, and colouring by samples?
  test1

nomadscientist:
2nd one, definitely

BUT
would that be the 3 colours by SAMPLES or by METADATA? I..e I'd want to see that for 'disease vs wildtype' but not for 'samples 1, 2, 3, 4, 5, 6' etc.

mtekman:
Ah I didn't consider that. I could definitely do it, but it'll take a day or two to implement.

Should the phenotype factor be given directly in the metadata of each dataset?
Or should the phenotype factor be stated explicitly as a text input in Galaxy?

I'm thinking the Galaxy solution for marking a dataset for a given phenotype would be better, since the metadata in multiple bulk datasets is likely to differ (e.g "Disease1" in bulk1, and "NameOfDisease1" in bulk2, even though they refer to the same thing)

nomadscientist:
Probably wise, yes, let Galaxy mark it!\
People might want to look at things in different ways (i.e. sex, disease factor, sex AND disease factor) so that would be really helpful

If you were doing by sample rather than group of samples, what did the box plot represent? Because surely then you only have 1 value for 'T lymph' %...

mtekman:
Oh these were summarized boxplots, and each of the 3 datasets contains multiple samples. 

So for example 'T Lymph' is actually made up 3 datasets across 30 data points:

| Cell Type | Sample Name     | Fraction | Dataset |
|-----------|-----------------|----------|---------|
| T-lymph   | APOL1.G1NF42    |    0.337 | Two     |
| T-lymph   | APOL1.GINF_44   |    0.338 | Two     |
| T-lymph   | APOL1.GNA73     |    0.215 | Two     |
| T-lymph   | APOL1.GNA78M    |    0.272 | Two     |
| T-lymph   | control.G1NF28  |    0.317 | Two     |
| T-lymph   | control.GNA1.18 |    0.048 | Two     |
| T-lymph   | control.NA.27   |    0.268 | Two     |
| T-lymph   | control.NA.30   |    0.220 | Two     |
| T-lymph   | control.NA.39   |    0.463 | Two     |
| T-lymph   | control.NAP.3   |    0.144 | Two     |
| T-lymph   | APOL1.G1NF42    |    0.094 | Three   |
| T-lymph   | APOL1.GINF_44   |    0.198 | Three   |
| T-lymph   | APOL1.GNA73     |    0.178 | Three   |
| T-lymph   | APOL1.GNA78M    |    0.468 | Three   |
| T-lymph   | control.G1NF28  |    0.329 | Three   |
| T-lymph   | control.GNA1.18 |    0.016 | Three   |
| T-lymph   | control.NA.27   |    0.096 | Three   |
| T-lymph   | control.NA.30   |    0.367 | Three   |
| T-lymph   | control.NA.39   |    0.235 | Three   |
| T-lymph   | control.NAP.3   |    0.190 | Three   |
| T-lymph   | APOL1.G1NF42    |    0.225 | B       |
| T-lymph   | APOL1.GINF_44   |    0.383 | B       |
| T-lymph   | APOL1.GNA73     |    0.402 | B       |
| T-lymph   | APOL1.GNA78M    |    0.394 | B       |
| T-lymph   | control.G1NF28  |    0.231 | B       |
| T-lymph   | control.GNA1.18 |    0.414 | B       |
| T-lymph   | control.NA.27   |    0.224 | B       |
| T-lymph   | control.NA.30   |    0.292 | B       |
| T-lymph   | control.NA.39   |    0.253 | B       |
| T-lymph   | control.NAP.3   |    0.124 | B       |

mtekman:
Although... now that I'm looking at the table, I definitely should give a way for the user to select different sample groups within a dataset... because I see that all datasets appear to have a control group embedded within...

mtekman:
Okay I've implemented factors into the analysis. Now when users are specifying their bulk datasets, they can state a phenotype column which contains a factor they want to use for grouping. 

If the phenotype column they state doesn't exist, then it assumes that this is a new factor to apply to all samples in that bulk dataset:

* Bulk dataset on Y-axis

  When I was setting the `factor name` for bulk datasets "Two" and "B", I set it to `Control` which **is** a real column in the metadata that specifies whether the sample is one of two values: "control" (green) or "APOL1" (red).

  But when I was setting the `factor name` for bulk dataset "Three", I set it to `Pheno1` which **is not** a real column in its metadata, and so we assume this is a new phenotype factor for that whole dataset.

   test4


* Celltype on Y-axis

   test3

    Same info as above, just showing cell types instead.

mtekman:
> But when plotted together, you lose the dataset but keep the factor.

Exactly, we just a get a pool of factors to group by and we're not interested where they're from. (Although, there will be plots to shown how these factors are expressed in each dataset too).

> BUT, if they wanted them plotted together, you'd just combine the bulk RNA datasets (which I think is in the RNA-seq tutorial anyway, right?) before running them through your tool.

No, hopefully not necessary - every bulk dataset is represented in the above plots. If you're worried that different factors expressing the same phenotype will be shown in these plots (confusing what the true factor is), then yes, the user should change the metadata in these bulk datasets so that they're the same across multiple bulk datasets, or as you say, combine all bulk datasets together and make sure the metadata column has factors that describe the same phenotype.

> Assuming a tabular output of average/st dev / median/quartiles /n values for each (i.e. for T lymph of CONTROL and T-lymph of APOL1), then people can do whatever comparisons they like!

Will do!

> OHHH follow-up: you need to make sure they can select which cell types to show. For example, cell types with no differences just make the graph harder to interpret, so 'hide factors' or something would be good for making final figures.

This I am implementing right now

mtekman:
Okay, the current tutorial is ready but the tools won't be refreshed into Galaxy until I resolve some merging issues in https://github.com/bgruening/galaxytools/pull/1192.

I think for now, record the tutorial on that test server and just crop out the address bar when recording the video?

nomadscientist:
OK cool - so currently we ignore the awesome and kick-ass MuSiC compare tool and go for the normal tutorial, yes? Yes, ok :) And then when I come back, assuming it's working, I'll add some tutorial snippets and an extra bit of video in case the GTN team allow a little bit of lateness. Yes? Yes.

mtekman:
Yes!

bgruening:
Tools are installed on EU :)

mtekman:
Excellent!

nomadscientist:
Can this be merged then too? What's stopping this from being merged?

bgruening:
@nomadscientist are you happy? Then we merge it :)

nomadscientist:
Yes! I think as is will work well. 
I might make some tweaks as I test it repeatedly, but ditto for any other tutorial! It's certainly great as is! Woo well done Mehmet!

bgruening (approved):

bgruening:
Wuhu! :1st_place_medal: @mtekman and @nomadscientist!

nomadscientist:
Huh. I think I could have probably approved it. DOH! Still getting my head around these newfound Github powers...

PR #2787, opened 2021-09-21
Topic: Data Science Survival Kit + GTN notebook Support

hexylena:
## GTN Jupyter Notebook Support

The best learning experience for students when covering topics like Python and R and similar is that they run the training materials directly as notebooks. We have a couple tutorials already which function like this, e.g. https://training.galaxyproject.org/training-material/topics/transcriptomics/tutorials/scrna-JUPYTER-trajectories/tutorial.html which has as an initial step loading an ipynb file into Jupyter.

image showing markdown, gtn, and rendered notebook side-by-side

This takes this a step further, and allows tutorial authors to write content in GTN Markdown, have it render nicely in the GTN, and, **automatically generates a notebook** using [notedown](https://github.com/aaren/notedown) converting it directly into a reasonably ok looking notebook. Importantly the student can then run through each cell and test things out.

It automatically generates the links to mybinder.org which will let students click and immediately start running the notebook, as well as instructions on Jupyter in Galaxy.

screenshot of tutorial showing box linking to notebooks

*Current drawbacks*

The main current drawback of this approach is it necessitates removing and output preview from the markdown, which is a bit sub-optimal. However, given the ease of "launch on mybinder", I don't think this is a huge drawback, and it is something that @shiltemann and I have discussed and we think we have a plan forward to knit back the output, time and energy permitting.

ping @fpsom xref #2764 

## Bash Lessons
This builds on https://github.com/swcarpentry/shell-novice/ 

This is a thing now :) a bash tutorial in the GTN converted to a notebook which can automatically run on EU/Org(?) + mybinder.

I like that setup for the 'workshop' will just be "Open the notebook and get started" because we can have it run setup for us:

image

This is @gallantries work. Fixes https://github.com/gallantries/organisation/issues/2

PR #2792, opened 2021-09-24
[WIP] Intro SQL lessons from Carpentries, building on our Notebook support

hexylena:
**Requires https://github.com/galaxyproject/training-material/pull/2787**

The SQL kernel requires some really kinda ugly magic
image

But I would rate the rest of the experience as "acceptable"

image

And the python lesson bits are nice.

image

I think I might change out the database and re-build it around a more Galaxy-ish dataset? Maybe a galaxy DB itself, then it'd double for admin training. What does @galaxyproject/training-admin  think?


hexylena:
Discussed out of band with @shiltemann, whether we should change the example to be more 'galaxy', i considered a galaxy database from a local instance, then we could talk about joining across tables, etc. and it would work nicely for admin training.

But @shiltemann raised the point that it's probably hard to generalise for others, so, let's stick to the existing example.

shiltemann:
Thanks @hexylena! And thanks for review @Dirowa for reviewing!!

PR #2796, opened 2021-09-28
Update advanced rule-building to use new 21.09 tool name

wm75:
The "Apply rule to Collection" tool has been renamed to "Apply Rules" as part of https://github.com/galaxyproject/galaxy/pull/11068.

Also fixed a few typos while at it.


hexylena (approved):
this looks great, thank you for fixing this!

PR #2797, opened 2021-09-29
Fix regex find example

wm75:
Not sure how this could go unnoticed until now, but the sed slashes must not be part of the find pattern.

### DIFF
--- a/topics/galaxy-interface/tutorials/workflow-parameters/tutorial.md
+++ b/topics/galaxy-interface/tutorials/workflow-parameters/tutorial.md
@@ -146,12 +146,12 @@ compose this text parameter using the `Compose text parameter value` tool.
 > 3. Add **Compose text parameter value** {% icon tool %} to the workflow
 >     - Add two more "Components" using the "Insert components" button in the workflow editor tool form on the right of the screen. You will need three "Component" blocks in total.
 >     - In the first component:
->       - *"Enter text that should be part of the computed value"*: `/(`
+>       - *"Enter text that should be part of the computed value"*: `(`
 >     - In the second component:
 >       - *"Enter text that should be part of the computed value"*: Leave empty and click "Add connection to module" {% icon galaxy-wf-connection %}
 >       - Connect the output of the **Input parameter** {% icon tool %} to this new input
 >     - In the third component:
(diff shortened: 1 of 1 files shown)


mvdbeek:
The actual regex tool in question was switched out (more than once ?), I'd assume this used to be necessary in a previous version. Does this work on the .org and .eu ? There was some trick to add the complete tool id, can you add this too just to be sure for the future ?

hexylena:
Yep, documentation for linking the tool here if you want to add that too! 

We have a [guide on how to do it](https://training.galaxyproject.org/training-material/topics/contributing/tutorials/create-new-tutorial-content/tutorial.html#tool-links), but in short just switch to `{% tool [Your Tool](toolshed.g2.bx.psu.edu/repos/..../your-tool/1.0.0+galaxy0) %}`

wm75:
The fixed version works on main and .eu, the original one on neither. There are only two versions of the tool and they both behave the same way. Also checked the file's history whether it used an entirely different tool at some point, but that doesn't seem to be the case either.
So this remains a mystery.

I can certainly add the tool links, but then I guess I should do this for all tools in the tutorial at once? Also I'm not sure how helpful that would be when you're inside the WF editor?

hexylena:
It can be done on one-by-one basis, it isn't necessary for all

wm75:
> Also I'm not sure how helpful that would be when you're inside the WF editor?

Hmm, ok but this one? Users are supposed to work in the WF editor throughout the tutorial so dropping them into the tool interface is not really helpful for them and they don't even have a hat icon to click on there.

wm75:
> and they don't even have a hat icon to click on there.

Oh, that was wrong - they have it.
Clicking on tool links doesn't have an effect though.

wm75:
Do you think the hover text and the fact that it puts the tested tool version in the markdown is valuable enough?

hexylena:
> ok but this one?

someday I'll not get distracted before I reach the end of the sentence and can actually answer your question :woman_facepalming: 

Yeah that's a very good point. 

>  Clicking on tool links doesn't have an effect though.

yeah, that's sub-optimal. But you're right, I think there's still enough value in just annotating that, at least then we know which version and testers can hover

PR #2816, opened 2021-10-05
Update tutorial.md

thepineapplepirate:
Changed line 244 from instructing the user to choose "Apply position restraints" to instead choosing "No position restraints". This is vital, and the without this, the simulation will crash. The reason is because, once the system is heated in NVT conditions and then goes into NPT, the atoms need to be allowed to move freely. If one restrains the atoms in NPT, and then takes those restraints off in the production run, said restrained atoms cannot suddenly adopt steady state velocities and the simulation crashes. 

It is important to note: Production run settings should generally, always be the same as those in the NPT runs. The only exception could be the number of steps (if the user wants a longer simulation for example), as well as the step size. Great tutorial though!

-BR

### DIFF
--- a/topics/computational-chemistry/tutorials/md-simulation-gromacs/tutorial.md
+++ b/topics/computational-chemistry/tutorials/md-simulation-gromacs/tutorial.md
@@ -241,7 +241,7 @@ Note that we can continue where the last simulation left off (with new parameter
 >    - *"Use a checkpoint (CPT) file"*: `Continue simulation from a CPT file.`
 >    - {% icon param-file %} *"Checkpoint (CPT) file"*: Checkpoint file produced by NVT equilibration
 >    - *"Produce a checkpoint (CPT) file"*: `Produce CPT output`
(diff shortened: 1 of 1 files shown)


thepineapplepirate:
Hi Simon,


No problem. I don't have a citation off the top of my head. But you could maybe run the tutorial once as is and see if the production simulation crashes on your end. I changed this one parameter and was able to get a succesful run.




Best regards,



Bryan A. Raubenolt, PhD

​Postdoctoral Researcher

The Rick Group​

Department of Chemistry

University of New Orleans

B.S. Chemical & Biomolecular Engineering (Tulane '12)

M.S. Chemistry (UNO '16)

Ph.D. Chemistry (UNO '20)







________________________________

simonbray:
Okay, I will take a look. Thanks again. :+1:

simonbray (approved):

simonbray:
ok, let's merge, thanks again @thepineapplepirate!

thepineapplepirate:
Thanks guys!

PR #2824, opened 2021-10-08
Performance improvements

hexylena:
Here are a couple performance improvements, please see the individual commits for details about the individual improvements but `--incremental` builds now take ~2s for me locally. Most of the improvements come from pre-calculating things instead of looping over `site.pages` in liquid, since liquid is ridiculously slow.

Here's the report after touching a file.

```
| PHASE      |   TIME |
+------------+--------+
| RESET      | 0.0011 |
| READ       | 0.6709 |
| GENERATE   | 0.1861 |
| RENDER     | 0.6689 |
| CLEANUP    | 0.1336 |
| WRITE      | 0.2709 |
+------------+--------+
| TOTAL TIME | 1.9315 |
```

<del>Full builds still are miserable due to citations. No workaround for that yet.</del> Full builds are now equivalent to dev builds, see below.


hexylena:
I've re-written `jekyll-scholar` and replaced it completely. Build times are down to ~30 seconds for the whole site, doing a 'full build'. Please see the comparison of these two actions:

Action | build time
--- | ---
[3860458190](https://github.com/galaxyproject/training-material/runs/3860458190) | 19m56s
[3867911994](https://github.com/galaxyproject/training-material/runs/3867911994) | 36s

I simultaneously discovered about 6 broken references. 

The new slowest step is the sitemap plugin, which needs file access times. I suspect that will be very painful on windows due to their file access behaviour.

PR #2849, opened 2021-10-13
Fix label of Auto-detect button in edit dataset snippet

Ajadi-Abiola:
The current tutorial in galaxy 101 for everyone has `detect datatype` button  text but the button text is `Auto-detect
changeDataType


shiltemann (approved):

shiltemann:
Thanks @Ajadi-Abiola! ..sorry for the delay, I guess this got lost in the deluge of PRs during outreachy, but better late than never ;)

PR #2850, opened 2021-10-13
Update datasets_change_datatype.md

Ajadi-Abiola:
The tutorial doesn't have **Change datatype** button text . Just **Save** to save the changes made. 
changeDataType


bgruening:
Oh, that is a nice catch. I think this is only with the upcoming Galaxy release. Have you used usegalaxy.org? On usegalaxy.eu which is still running the 21.05 release its still called "change datatype"

grafik

I guess its fine to merge, as the new release is out very soon and we all will update in a few weeks.

bgruening:
@Ajadi-Abiola one recommendation for your PRs, please make sure the title is meaningful that will help reviewers.

Ajadi-Abiola:
> Yep, that is indeed a very good observation, which we - the 21.09 release testing group - should have made ourselves, but didn't. Thanks @Ajadi-Abiola ! As @bgruening said, you are more likely to get a speedy review if you choose a title that summarizes the intention of the change, not which file you changed. Here you could have used something like: "Fix label of Save button in edit datatype snippet" or similar.

 You are welcome @wm75  and Thank you for your suggestion. I appreciate.

Issue #2401, opened 2021-02-23
From peaks to genes - slighly outdated and possible tool error


davelopez:
In general, the tutorial can be followed in 21.01 but some minor updates can be made as part of the next cofest:

These minor issues mainly affect the [section related to the Workflow Editor](https://training.galaxyproject.org/training-material/topics/introduction/tutorials/galaxy-intro-peaks2genes/tutorial.html#extracting-workflow):
- Some of the screenshots are a bit outdated.
- Instead of the new checkboxes for outputs, the `asterisks` are mentioned to hide intermediate steps in a workflow.
- There is a save icon to save the workflow without accessing the 'gear' icon (which is for the 'save as...' option).

### Possible version incompatibility error?
I also faced an error when running the workflow at the [Repeat workflow step](https://training.galaxyproject.org/training-material/topics/introduction/tutorials/galaxy-intro-peaks2genes/tutorial.html#repeat-workflow). I can not tell if this is some problem with the datasets, the version of the tools, or my own clumsiness, but here is the error report after running the workflow on usegalaxy.org:

`An error occurred while running the tool toolshed.g2.bx.psu.edu/repos/devteam/intersect/gops_intersect_1/1.0.0.`

std-err:
````sh
Exception while parsing command line:
Traceback (most recent call last):
  File "/cvmfs/main.galaxyproject.org/shed_tools/toolshed.g2.bx.psu.edu/repos/devteam/intersect/33b3f3688db4/intersect/gops_intersect.py", line 35, in main
    chr_col_1, start_col_1, end_col_1, strand_col_1 = parse_cols_arg( options.cols1 )
  File "/cvmfs/main.galaxyproject.org/deps/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b723249629522600dc67/lib/python2.7/site-packages/galaxy/tools/util/galaxyops/__init__.py", line 29, in parse_cols_arg
    col_list = map( lambda x: int( x ) - 1, cols.split(",") )
  File "/cvmfs/main.galaxyproject.org/deps/_conda/envs/mulled-v1-a4cb0f5781d6f9de5e288400a2ac9fa34202068e5723b723249629522600dc67/lib/python2.7/site-packages/galaxy/tools/util/galaxyops/__init__.py", line 29, in <lambda>
    col_list = map( lambda x: int( x ) - 1, cols.split(",") )
ValueError: invalid literal for int() with base 10: 'None'


[...]

PR #2882, opened 2021-10-19
Updated the outdated screenshots

rahmot:
I updated the following outdated screenshots on From Peaks to Genes Tutorial [issue 2401](https://github.com/galaxyproject/training-material/issues/2401): login_register, upload_button, intro_02, and intro_summits


bebatut:
Thanks @rahmot. Could you rename all files with updated screenshots so they replace the previous ones? Thanks

rahmot:
@bebatut, what about renaming the updated screenshots with the previous name so that it automatically replaced them in the files?

Issue #2833, opened 2021-10-11
Create "how to get started" slide for new contributors


rahmot:
I observed that over 70% of Outreachy Interns that are interested in contributing to Galaxy get lost in the pool of information available. How about those outside Outreachy that are desirous of contributing to the project? Do they find it easy to get started? Galaxy needs diverse talents and skills to achieve its vision. 

So, I propose that we create a 10-page slide that clearly and visually explains how to get started. This would encourage more people to contribute to the project and also eliminate the time spent repeating the steps to get started. We can pin this slide when completed to the Gitter Chat. 

I would like to collaborate with whoever is willing to work on this. Let's create the magic together! :)

rahmot:
@bebatut I never knew we have something like this but I think we need to tweak it and I would like to work on that ASAP. I will make changes to the slide and share it with you for your feedback. What do you think?

PR #2883, opened 2021-10-19
Added a welcome slide to the slide deck

rahmot:
I added a welcome slide to the existing " Overview of the Galaxy Training Material " slide deck. [issue 2833](https://github.com/galaxyproject/training-material/issues/2833)


shiltemann:
Thanks @rahmot!

If you would like to keep working on this slide deck, it would also be very useful to have speaker notes on each slide

```
# Slide title

slide content

???
Speaker notes here

```

We can then automatically turn these slides into a video like [this one](https://training.galaxyproject.org/training-material/videos/watch.html?v=introduction/tutorials/galaxy-intro-short/slides), where the speaker notes will be spoken over each slide.

This can be done in a follow-up PR, and no pressure to do it, but just an idea if you wanted to do more with this slide deck.

rahmot:
@shiltemann, thanks for the feedback! Yes, I've chosen to work on this slide deck fully but since I am still grappling my way around Github and the whole PR thing, I am taking it a step at a time. I hope that is not a problem!

shiltemann:
@rahmot absolutely not a problem! I will merge this as soon as I fix the failing test (which is unrelated to your change) thanks!

PR #2888, opened 2021-10-20
Build videos on push to main now that build is fast.

hexylena:
Originally we built these on cron due to the very very long build times. We now build these alongside the deploy step which means they'll show up a bit faster, because we have very fast build times so it's not a huge waste of time for 95% of PRs which don't include video changes. Additionally this adds a second enhancement to not bother running the rest of the steps if no videos would be built.

There's no good way to test it :(


shiltemann (dismissed):

shiltemann:
oh, theres a merge conflict

hexylena:
I think this is done now, ready @shiltemann

hexylena:
This probably works fine, but, meh.

PR #2894, opened 2021-10-20
Add AfiDede as contributor.

AfiDede:
Please add AfiDede as contributor.


shiltemann:
Thanks for the PR fixing the alt text!

PR #2948, opened 2021-10-26
Update core-contributing tutorial

pauldg:
Suggested change for the "Contributing a New Feature to Galaxy Core" tutorial. Moving the table definition from `mapping.py` to `__init__.py`. Also, I think it should be noted that python 3.7 is required for FastAPI.

### DIFF
--- a/topics/dev/tutorials/core-contributing/__init__.py_diff.md
+++ b/topics/dev/tutorials/core-contributing/__init__.py_diff.md
@@ -4,20 +4,29 @@
 > Possible changes to file ``lib/galaxy/model/__init__.py``:
 > 
 > ```diff
-> index 2c0b8a4dc4..69a74b0ad2 100644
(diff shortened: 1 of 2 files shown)


hexylena:
I've pinged the author @jmchilton who might have time to check over this update. Thanks for contributing it @pauldg! Please feel free to add yourself to the CONTRIBUTORS.yaml file as well!

ic4f:
Thank you for fixing this, @pauldg! All looks good!
In the actual code base we would add a `back_populates` attribute to the relationship + a matching relationship to the `User` table  (for details see https://github.com/galaxyproject/galaxy/pull/12064, under "Notable Issues > backref vs back_populates"). Also, we would add tests to `test/unit/data/model/mapping/test_model_mapping.py` - but that's not necessary for now for this particular example.

ic4f (approved):

hexylena:
Thanks for reviewing @ic4f , appreciate it!

PR #2959, opened 2021-10-29
enabled video in the option for using galaxy tutorial

Alaa-Elghamry:
I was going through the tutorial of adding automatic video to slides and I created one for the " option for using galaxy "  tutorial

### DIFF
--- a/topics/introduction/tutorials/options-for-using-galaxy/slides.html
+++ b/topics/introduction/tutorials/options-for-using-galaxy/slides.html
@@ -1,6 +1,7 @@
 ---
 layout: tutorial_slides
 logo: GTN
+video: true
 
 title: "Options for using Galaxy"
 zenodo_link: ""


bebatut:
Thanks @Alaa-Elghamry.
To really be able to create a video out of this slide deck, all slides should have speaker notes as explained [here](https://training.galaxyproject.org/training-material/topics/contributing/tutorials/slides-with-video/tutorial.html#enabling-video)

hexylena:
We'd love it if you could write those @Alaa-Elghamry ! Just imagine you are giving these slides as a presentation to an audience, and write down what you would say on each slide.

Alaa-Elghamry:
of course! I would be happy to help but I'm not sure I fully understand
where to write my speaker notes though
I tried the tutorial but I just don't get the idea
it would be great if you could explain that to me

bebatut:
You can speaker notes directly with the slides. See the documentation here: https://training.galaxyproject.org/training-material/topics/contributing/tutorials/create-new-tutorial-slides/slides.html#18

PR #2964, opened 2021-10-31
Changes sdm tut

yokobird:
I noticed an error with the column number 51 on the Cut section, the column option that worked for me was 54. I also noticed BISON was linked to gbif, so I changed that to their current url, however BISON does have a notice their website that they will be going dark this year and merging to gbif.us, so I wasn't sure if that should be kept or not.

I also noticed a few spelling and grammar errors so I changed them.


yvanlebras:
Thank you so much @yokobird !!!!! Will look at it ASAP and also retest the Wallace GxIt as it seems to me during last test I wasn't able to do all the tuto. One ""big"" issue related to your correction is the fact that gbif data formatting output is changing with time :(

Issue #2973, opened 2021-11-02
Support contributors on the FAQs


hexylena:
These are full fledged pages now, we could recognise all of the contributors to individual FAQs. This would help recognise newcomers to the GTN who might be contributing to one of those as an easy way to get started.

To do this one would need to:

- Add an `contributors:` field to the FAQ pages
- Update `_layouts/faq.html` to include the contributor badges (see _layouts/tutorial_hands_on.html for an example)

Anyone is welcome to work on this! Just write that you are here and open a PR with your changes. We don't need to update all past FAQs to start with, we can do that later (and it won't be easy, we'll have to dig through the git history for that.)

saritadc:
@hexylena I want to work on this.

millen2223:
@hexylena, me too :)

Ezichi-Fav:
can I collaborate too?

saritadc:
> @hexylena I want to work on this.

@hexylena  can you have a look on this PR https://github.com/galaxyproject/training-material/pull/2980

PR #2980, opened 2021-11-03
add contributors to FAQ page

saritadc:
added contributors to the FAQ page.

### DIFF
--- a/_layouts/faq.html
+++ b/_layouts/faq.html
@@ -12,6 +12,6 @@ layout: page
     {{ page.description }}
     <hr>
 
-    {{content | regex_replace: '<hr.*—<\ /p>', '' }}
+    {{content | regex_replace: '<hr.*—<\/p>', '' }}
 
 </section>
--- a/_layouts/faq.html
+++ b/_layouts/faq.html
@@ -2,10 +2,16 @@
 layout: page
 ---
 
-<section class="tutorial">
-{{ page.description }}
-<hr>
+{% assign contributors = site.data['contributors'] %}
 
-{{content | regex_replace: '<hr.*—<\/p>', '' }}
+<section class="tutorial">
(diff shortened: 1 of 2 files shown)


shiltemann:
Thanks for your contribution @saritadc! We don't track contributors of FAQs, only tutorials and slides, so this will not display any contributors currently.

If you would like to work on FAQs, we could use somebody who walks through all of them to check if they are still up-to-date with the latest version of Galaxy, since the FAQs cover frequently used tasks in the interface, which changes a lot. 

A slightly bigger project here would be to support multiple versions of the FAQs, connected to differenct Galaxy versions, and the ability to switch between them. But we would need to think together about how to do that a bit first.

hexylena:
@shiltemann this is to fix https://github.com/galaxyproject/training-material/issues/2973

_layouts/faq.html
```diff
@@ -2,10 +2,18 @@
...
+    <div class="contributors-line">{{ locale['authors'] | default: "Authors" }}: {% include
+        _includes/contributor-list.html
+        contributors=page.contributors badge=true newcontributors=true %}</div>
+  {% endif %}
+
+    {{ page.description }}
+    <hr>
 
-{{content | regex_replace: '<hr.*—<\/p>', '' }}
+    {{content | regex_replace: '<hr.*—<\ /p>', '' }}
```
hexylena, line 17:
```suggestion
    {{content | regex_replace: '<hr.*—<\/p>', '' }}
```

I don't think this can have a space here.

hexylena (commented):
This looks good @saritadc I've tested locally. Would you like to fix the one review comment and also maybe add yourself to the CONTRIBUTORS.yaml file as well? And then I think I'm happy to merge.

saritadc:
@hexylena I have fixed the review comment and I think it's good to go. Thank you!

hexylena (approved):

PR #2988, opened 2021-11-05
Testing self-signed ssl certs in git-gat

hexylena:
We've written a self-signed-certs role (it was so easy!) And I'm testing replacing certbot with the SSCs in testing environments because it makes life a LOT easier.


hexylena:
The changes are probably still good, but maybe we need a flag for it. I'll revisit it someday.

PR #2992, opened 2021-11-06
Add<Enchanteur14>as contributor

Enchanteur14:
I would like to work on the UI of this page, this is because I had an issue navigating though and some other persons must have had the same issue. It is not very user friendly, I raised an issue on it and it can be found here: https://github.com/galaxyproject/galaxy/issues/12862

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -1046,5 +1046,11 @@ yvanlebras:
     orcid: 0000-0002-8504-068X
     joined: 2017-09
 
-
+Enchanteur14:
+    name: Osuagwu Oluchi
+    email: oosuagwu7@gmail.com
+    twitter: Enchanteur14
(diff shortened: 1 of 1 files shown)


hexylena:
Hi @Enchanteur14 that is a Galaxy issue (not a Galaxy Training Network issue) and thus we cannot be very helpful there :(  You can work on it, but it will require a lot of javascript knowledge to make changes to that interface.

Enchanteur14:
Oh ok I understand... I thought is something a UI designer would implement first before the front end developer

hexylena:
UI Suggestions are always welcome! We don't have anyone who only has the job of UI designer so I did not even consider it, but if you have suggestions for making that UI better please share them :)

Enchanteur14:
Ok I'll drop my suggestions on the UI soon

PR #3088, opened 2021-12-01
Add a news for assembly/annotation

abretaud:
Adding a news for new assembly/annotation stuff, does it look good like this?


hexylena (approved):
This is 100% perfect, exactly what I was hoping for.

PR #3089, opened 2021-12-01
Add two new FAQs: How the GTN stays fair + collaborative development

hexylena:
@bebatut maybe you can review, I extracted these from the paper since we get asked it occasionally and maybe it's nice to 'show off'.


bebatut (approved):
It looks good. Thanks @hexylena for doing that

Issue #2015, opened 2020-07-29
More options for acknowledging contributions on tutorials?


pvanheus:
In line with a [all-contributors](https://github.com/all-contributors/all-contributors) type model, it would be good to be able to specify the kind of contribution made to training material by the different people listed. This could also help clarify what "contribution" is. 

A possible technical mechanism is to allow the `contributors` key to accept a list of maps, with name and contribution type.

hexylena:
Interesting, hadn't heard of that! Since we're science oriented, we could also consider [SCoRO](https://sparontologies.github.io/scoro/current/scoro.html) or similar?

hexylena:
https://allcontributors.org/docs/en/emoji-key I think we might want to come up with our own emoji key because those don't cover the full set of cases. but let's revive this maybe in the linked discussion, I want to see something like this happen.

Updating the `contributors` key would be easy, just instead of 

```
contributors:
- hexylena
- shiltemann
```

We could replace it with

```
contributors:
  shiltemann: [writing]
  hexylena: [spell-checking, copy-editing, testing]
```

or so.

hexylena:
I've opened a discussion https://github.com/galaxyproject/training-material/discussions/2986 here that's related

hexylena:
Ok, our own categories for contribution could be: (writing down for my memory since @shiltemann and I had a discussion in person a while back and I already forgot it :woman_facepalming: )

- :pencil2: writing / authorship
- :memo: copy-editing (spell-checking, typos, very small fixes?)
- :gear: testing
- :nail_care: layout/design?
- :desktop_computer:  infrastructure / sysadmin / github actions editing / etc

hexylena:
https://eos.org/opinions/credit-where-credit-is-due
https://groups.niso.org/higherlogic/ws/public/download/26466/ANSI-NISO-Z39.104-2022.pdf
https://www.ebi.ac.uk/ols/ontologies/cro
https://ddialliance.org/

PR #3111, opened 2021-12-14
Annotate funding entities + Smörgåsbord2 news post

hexylena:
We will now support annotating the funding agencies that have supported work in the GTN, the erasmus+ programme is the first one (as they funded this feature ;) ). This is part of https://github.com/galaxyproject/training-material/issues/2015 where we're doing more work to annotate how the GTN is made and recognising all contributions.

image


abretaud (approved):
:+1: for me!

PR #3118, opened 2021-12-20
Video library: add column to topic page, add dates

shiltemann:
added a column for the GTN video library recordings to the topic page. Also added the date of the latest recording to the dropdown.

Removed the tours column since we are not really mainaining those actively.

image

and also: 
- use different icon for video slides to distinguish from live recordings
- add link to video slides on main slide deck
  
image

- re-use the slides include for topic-introduction slides
- generate plaintext slides for introductory slide decks
- rename "Galaxy instances" column to "Galaxy servers"

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -174,6 +174,7 @@ icon-tag:
   workflow: fas fa-share-alt
   workflow-run: fas fa-play
   video: fas fa-video
+  video-slides: far fa-play-circle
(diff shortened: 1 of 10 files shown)


hexylena (approved):
This looks fantastic, I like the use of an alternative icon for the slides. Maybe in the future we should add support for multiple languages in the video-library and then here, but for now this looks fantastic!

PR #3139, opened 2022-01-18
Update snippet import_via_link to remove rename

mblue9:
Galaxy doesn't use the URL as the name anymore so should this bit below be removed from the snippet? (did that in this PR)
```
{% if include.renaming == undefined or include.renaming == true %}
* By default, Galaxy uses the URL as the name, so rename the files with a more useful name.
{% endif %}
```


hexylena (approved):

hexylena:
I'm so glad this if fixed!

PR #3145, opened 2022-01-26
Galaxy 101: Change tool id for Select first

bebatut:
`Show beginning1` is failing in Galaxy Europe (broken pipe in cat). `toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_head_tool/1.1.0` works better

The error seems discussed and somehow fixed (https://github.com/galaxyproject/galaxy/pull/12112). Not sure why not still failing


hexylena:
Urgh. We'd have to rerecord this. I hope they can get it fixed in time

hexylena (approved):

PR #3146, opened 2022-01-26
Funannotate tapas update

abretaud:
Small changes to the funannotate tuto before recording https://github.com/gallantries/video-library/issues/38
First, an updated version of eggnog-mapper => with the old version the "funannotate annotate" was not able to parse the output and include it in the final gff3 file

### DIFF
--- a/topics/genome-annotation/tutorials/funannotate/data-library.yaml
+++ b/topics/genome-annotation/tutorials/funannotate/data-library.yaml
@@ -10,30 +10,30 @@ items:
   items:
   - name: Structural and functional genome annotation with Funannotate
     items:
-    - name: 'DOI: 10.5281/zenodo.5729114'
+    - name: 'DOI: 10.5281/zenodo.5906635'
       description: latest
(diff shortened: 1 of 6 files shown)


abretaud:
Ah yes wonderful! I'll do it when I'm done updating the workflow

hexylena:
merci!

abretaud:
This one's ready I think, the tutorial works fine on usegalaxy.fr and .eu (after fighting with interproscan!)
I've tried to add a workflow test, but the test-data dir is >1Gb... I guess there's no way to use data from a remote url in the test?

abretaud:
Switched back to wip: I'll adapt the workflow test based on https://github.com/galaxyproject/iwc/blob/main/workflows/sars-cov-2-variant-calling/sars-cov-2-pe-illumina-artic-ivar-analysis/pe-wgs-ivar-analysis-test.yml#L9

gallardoalba (approved):

PR #3148, opened 2022-01-28
Document how to use the new features

hexylena:
- Updates the contributing page to mention new RMarkdown
- Adds a news post about it
- Embeds images directly in notebooks automatically (for png, jpg, svg) rather than using normal links. This was continually a problem until now, as the image would be in the fork of the GTN that I was building, but the notebook wouldn't know how to reference that image in a stable/useful way. So: embed the image! It works at the expense of file size, which I think is OK for this use case.


bgruening (approved):
Very cool! Thanks @hexylena

PR #3154, opened 2022-02-06
Add zenodo download link in molecular systems tutorial

mvdbeek:
We've had a bug report from someone that pasted the link directly.
The direct link seems better here.

 Note also that this is sniffed as a pqr file. Is that correct (or unavoidable) or do we need to tighten Galaxy's sniffers ?


simonbray (approved):

simonbray:
Thanks!

> Note also that this is sniffed as a pqr file. Is that correct (or unavoidable) or do we need to tighten Galaxy's sniffers ?

I've noticed that before as well, I'll have a look at it.

PR #3162, opened 2022-02-09
GxIT developement tutorial proposal

eancelet:
Thanks to all contributors @neoformit, @Lain-inrae, @abretaud, @yvanlebras.


topics/dev/tutorials/interactive-tools/tutorial.md
```diff
@@ -0,0 +1,872 @@
...
+Let's check out
+[the Dockerfile](https://github.com/Lain-inrae/geoc-gxit/blob/master/Dockerfile)
+that we'll use to containerize our application.
+
+This container recipe can be used to build a Docker image which can be pushed to a
+container registry in the cloud, ready for consumption by our Galaxy instance:
+
+```dockerfile
+# build on an existing container
+FROM debian:testing
```
hexylena, line 238: Any reason to not build this on rocker/r-shiny?

topics/dev/tutorials/interactive-tools/tutorial.md
```diff
@@ -0,0 +1,872 @@
...
+    fi
+done
+```
+
+This script can then be built into the docker image and called in the tool XML `<command>` before the main application call. Alternatively, it can be called within a `run.sh` script inside the container itself (see below).
+
+```sh
+/scripts/monitor_traffic.sh &
+bash my_run_script.sh
+```
```
hexylena, line 809: This seems a bit outdated, this was true for GIEs but with GxITs that support the model of "leave it running in the background", we've mostly stopped doing this, I don't think we should recommend it here.

neoformit: Thanks Helena, I wasn't sure of the history of this script so great to know, thanks!

topics/dev/tutorials/interactive-tools/tutorial.md
neoformit: I have rebuilt this to be simpler - just have to PR against Lain's repo and add to the tutorial

topics/dev/tutorials/interactive-tools/tutorial.md
```diff
@@ -0,0 +1,872 @@
...
+> ### {% icon comment %} Interactive tool infrastructure
+> Interactive tools require some rather complex infrastructure in order to work! However, most of the infrastructure requirements are taken care of by Galaxy core. As such, wrapping a new GxIT requires only three components:
+> - Application script(s)
+> - Docker container image
+> - Galaxy tool XML
+>
+> However, as we will see in the next section, testing and deploying a GxIT is not so simple.
+{: .comment}
+
+---
```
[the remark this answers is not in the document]
neoformit: Done

topics/dev/tutorials/interactive-tools/tutorial.md
hexylena: Ok, that works! Thanks @neoformit

hexylena:
Reviewing currently :)

hexylena:
@eancelet could you give me permission to push to this branch?

hexylena:
overall it looks quite good! This is going to be fantastic to have.

I feel slightly worried about the setup required for following this tutorial (there were things like 'sudo apt install docker' which is ubuntu/debian specific, won't work for other users, I think Windows WSL2 has a special docker install procedure as well) so I've replaced some of those where possible. 

_But_ this is also me comparing to the 'science' tutorials (where we know galaxy is going to be the same everywhere) or the 'admin' tutorials (where we are very explicit about the required starting environment.

Anyway, whenever you give me permissions I can push my changes to make it a bit more gtn-style following and we can merge!!

hexylena:
@eancelet could you resovle the merge conflict, and then I'm ok to merge :)

hexylena (approved):

yvanlebras:
THANK you all, thank you Helena for this finishing !

topics/dev/tutorials/interactive-tools/tutorial.md
```diff
@@ -0,0 +1,795 @@
...
+{: .hands_on}
+
+## Run Galaxy
+
+Go to the Galaxy directory and:
+```sh
+./run.sh
+```
+
+Galaxy is available at [http://localhost:8080/](http://localhost:8080/) and you should be able to use your GxIT.
```
mvdbeek, line 592: Does that really work for ITs that talk back to Galaxy, like Jupyter  ? You'd probably have to either use the host network or `add-host host.docker.internal:host-gateway` in the `docker_run_extra_arguments`

neoformit: @hexylena would this work for Rstudio or would you have to run this differently?

hexylena: it worked for as long as I'd been working on IEs/ITs, did that change? Is it now more isolated? They use the docker gateway address usually to talk back to galaxy, and that worked at the time?

Lain-inrae: I don't know if it helps, but when I have a tool that needs to communicate with galaxy, I use these env vars in the tool xml:
```xml
        <environment_variable name="HISTORY_ID" strip="True">$__history_id__</environment_variable>
        <environment_variable name="GALAXY_URL">${__app__.config.galaxy_infrastructure_url}
  ```
  And in the R side, I do the following:
  ```R
        
    gx <- .self$context$configuration$raw_config$galaxy
    if (is.na(API_KEY <- Sys.getenv("API_KEY", unset=NA))) {
        API_KEY <- gx$api_key
    }
    if (is.na(GALAXY_URL <- Sys.getenv("GALAXY_URL", unset=NA))) {
        GALAXY_URL <- gx$galaxy_url
    }
    if (is.na(HISTORY_ID <- Sys.getenv("HISTORY_ID", unset=NA))) {
        HISTORY_ID <- gx$history_id
    }
    Sys.setenv(GX_API_KEY=API_KEY)
    Sys.setenv(GX_GALAXY_URL=GALAXY_URL)
    GalaxyConnector::gx_init(
        GALAXY_URL=GALAXY_URL,
        API_KEY=API_KEY,
        HISTORY_ID=HISTORY_ID
    )
    
    ...
    
    GalaxyConnector::gx_put(output_file_path)
```
This will work **if your galaxy is exposed to the internet** or your local **DNS can resolve its name**.
If it is not exposed to the internet - for example in a dev environment - you'll not be able to access galaxy with env variables.

Instead, you'll have to write something like :
```
        <environment_variable name="GALAXY_URL">http://172.17.0.1:8080/</environment_variable>
```

An other option, like you said @mvdbeek , is to add some docker extra args, but I don't know how to do it. I remember it was used in GIE, but in GIT I've never tried.


In both cases, galaxy needs to listen on 0.0.0.0 so it is accessible from other networks.
```
uwsgi:
  http: 0.0.0.0:8080
```

In the case of a dev environment, I said to access the galaxy instance with http://localhost:8080 because when we will run the IT, the URL provided by galaxy will be :
subdomain.interactivetool.the_domain_name_you_are_using:port
but the problem is: if you access your galaxy instance with 127.0.0.1:8080, then you'll end up with the url:
 http://subdomain.interactivetool.127.0.0.1:8080/
 which is not a valid url.
 Thus, I always use `localhost:8080`, which is transformed in 
 http://subdomain.interactivetool.localhost:8080/
 
 That's why I suggested to use localhost.

PR #3173, opened 2022-02-16
fix tool ids in ref-based RNA-Seq tutorial

pavanvidem:
The tools from the toolshed have different parameters or at least different parameter names. Either we can just correct the tool names as I suggested in this PR or we can change the parameters. If we opt for the latter case, the tutorial will not match the recording on GTN.


hexylena:
So if I understand it correctly, the parameters don't match the TS tool? They match the ID you've corrected it to? Correcting the tool id to the builtin cut makes more sense to me here for sure :)

hexylena (approved):

shiltemann:
Thanks a lot @pavanvidem!

pavanvidem:
> 

Yes and yes, @hexylena

PR #3190, opened 2022-02-24
Reorganise admin tutorials into new subtopics

hexylena:
Clean up the admin area into some more useful subtopics. I think it just shows how much we were using tags for what could be subtopics. Also how poorly tagged our tutorials are, and k8s are the only two to use a "level", we should consider what levels are appropriate for the rest.

- ~should we tag everything that uses ansible? I think maybe.~
- ~should we tag the GAT course tutorials as "gat"? I think maybe.~
- ~should we tag git-gat tutorials? could be ok?~

Screenshot 2022-02-24 at 12-28-01 Galaxy Training Galaxy Server administration

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -1186,7 +1186,7 @@ erasmusplus:
     github: false
     funder: true
     funding_statement: |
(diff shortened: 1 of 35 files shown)


martenson:
I like it, maybe we could reorder the topics a bit? I suggest maintaining->monitoring->further learning

hexylena:
wanna approve @martenson ?

martenson (approved):

hexylena:
danke!

PR #3192, opened 2022-02-25
Fix some links in the read mapper overview slide

wm75:
Fixed:
- The GCSA link was identical to the HISAT2 link.
- The HISAT2 line was broken due to a missing closing parenthesis
- The MMP link was to an unrelated chemistry method with the same acronym, when really there is no separate link to MMP cause it's defined in the RNASTAR paper itself.


hexylena (approved):

hexylena:
Thanks @wm75

PR #3209, opened 2022-03-01
tutorial mode FAQ and add to transcriptomics tutorials

nomadscientist:
Transcriptomics tutorial tools are outdated, so need users to access older tool versions
Also changed name 'GTN in Galaxy' to 'Tutorial mode' because I'm a PR genius


hexylena (approved):
brilliant, visionary.

PR #3221, opened 2022-03-04
Fix link to advanced mdanalysis workflow

### LINKED ISSUE #3200: MDAnalysis workflow missing
In https://training.galaxyproject.org/training-material/topics/computational-chemistry/tutorials/analysis-md-simulations/tutorial.html#further-analysis the referenced workflow is missing. @simonbray prehaps you know which workflow this is referencing?

It wasn't caught because of the use of 'remote' URLs rather than local-to-training-material URLs. This could be replaced with `{% link topics/computational-chemistry/tutorials/analysis-md-simulations/workflows/advanced_workflow.ga %}` to ensure we catch that in the future.

simonbray:
Fix #3200, thanks for pointing that out @hexylena

### DIFF
--- a/topics/computational-chemistry/tutorials/analysis-md-simulations/tutorial.md
+++ b/topics/computational-chemistry/tutorials/analysis-md-simulations/tutorial.md
@@ -197,7 +197,7 @@ You can choose to use the tools one by one as described above, or alternatively 
 
 # Further analysis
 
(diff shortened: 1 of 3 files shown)


hexylena (approved):
With tests too?? :heart_eyes: awesome!

PR #3244, opened 2022-03-11
Stop deleting newlines within pre blocks

hexylena:
By manually replacing them with BRs before the newline deleter can get
to them

This fixes issues that are currently live:

Before | after
--- | ---
image | image
image | image


hexylena:
@shiltemann as you can see it's still not right :/ in the bottom right screenshot it's got *some* newlines but not all that it needs

shiltemann (approved):

PR #3254, opened 2022-03-14
Search and index FAQs

hexylena:
image

you can search just on FAQs, e.g. to find that pesky chrome problem

image


shiltemann:
this is awesome @hexylena !

shiltemann (approved):

PR #3272, opened 2022-03-17
Update ref-based tutorial.md

jennaj:
Reorganized an option to match the current 2.0.1+galaxy2 Featurecounts tool form

Questions:

1. This creates a mismatch between the tool version in the tutorial (2.0.1) and the current version (2.0.1+galaxy2) at usegalaxy.* servers. 
2. I didn't change the tool version link yet -- should that be done?
3. What else needs to be changed if we do this -- the workflow? anything else

Ok to reject this -- not sure how it is decided what is worth updating versus not


hexylena:
> I didn't change the tool version link yet -- should that be done?

yes, absolutely must.


> What else needs to be changed if we do this -- the workflow? anything else

yes, workflow version of the tool. That's it though.

hexylena:
@jennaj do you have time to update the workflow as well?

PR #3290, opened 2022-03-25
Making "Replace parts of text" explicit

nekrut:
Since tool search is not very specific simply searching for "Replace" will yield a lot of tool. Changing. this to "Replace parts of text" will make it a whole lot easier.


nekrut:
Wait .... now I'm not sure it is the right tool

nekrut:
Yes, that the one

shiltemann (approved):

PR #3296, opened 2022-03-28
merqury explanation clarification

nekrut:
Some clarification. Also removed "The fact that the primary haplotype shows a diploid genetic endowment can be explained as a result of the high divergence between the artificially generated haplotypes." is it not clear.


gallardoalba:
The training is going to be updated in order to avoid this problem, so probably this paragraph will be unnecesary.

PR #3305, opened 2022-03-28
RNA-seq reads counts: small update

mblue9:
This has just a couple of small updates
-updates Cutadapt parameter name to new version Minimum length -> Minimum length (R1)
-adds tool links for tutorial mode


hexylena (approved):
fantastic!

PR #3310, opened 2022-03-29
[FAQs] Added Finding Tools with Tools missing

garimavs:
The PR is a part of Issue: Move FAQs to GTN of the [galaxy_hub](https://github.com/galaxyproject/galaxy-hub) repository.

I have added the Finding tools-related steps with the tools missing one.
Kindly suggests improvements, if any.


faqs/galaxy/tools_missing.md
```diff
@@ -3,26 +3,46 @@ title: If a Tool is Missing
...
 
+To use the tools installed and available on the Galaxy server:
+
+1. At the top of the left tool panel, type in a tool name or datatype into the tool search box.
+2. Shorter keywords find more choices.
+3. Tools can also be directly browsed by category in the tool panel.
+
+If you can't find the tool you want:
+
+1. Try searching the [Tool Shed](http://toolshed.g2.bx.psu.edu/).
```
shiltemann, line 17: Thanks for your contribution @garimavs!

I don't think the second step here should be searching the Tool shed and talking about wrapping of tools. This FAQ was meant for GTN tutorials, so we know they should be installed on the server (if they made sure to use a compatible Galaxy), but the user is just not able to find it.

But maybe this could go into a separate FAQ, something like "adding new tools" orso? What do you think?

CONTRIBUTORS.yaml
```diff
@@ -12,825 +12,829 @@
 #
 # collection names should be equal to github username, if not, add github: false tag
 ---
-
 # two dummy entries so that autogenerated tutorials do not fail to build
 contributor1:
-    name: Contributor A (Please add contributors to the CONTRIBUTORS.yaml file)
```
shiltemann, line 18: I am guessing your text editor did this automatically, but please undo all the indentation changes here

faqs/galaxy/tools_missing.md
garimavs: Hello @shiltemann, I once considered that adding this content might not bear much co-relation, but then I felt that rather than only referring to GTN tutorials, we modify it such that every case of missing/can't find tools comes under this FAQ.

Though I will consider your insight and open a separate FAQ regarding "adding new tools."

shiltemann: @garimavs I understand your reasoning, but this FAQ is added to every individual tutorial FAQ pages, where people who are following a tutorial will go when they get stuck, and this is one of the most common questions, so it would be best if only the most relevant answers for that audience are here. 

Having a (different) FAQ as a landing page for all "missing tool" questions, like you suggest, could be a good idea, then that can link out to this and other FAQs, e.g. "Can't find a tool for a GTN tutorial?" (link to this one). as well as some of these other cases you describe, such as requesting installation of toolshed tools to a server, or adding new tools to the toolshed.

CONTRIBUTORS.yaml
garimavs: Yes, I believe my editor auto-formatted it. I will undo the changes. Sorry for the inconvenience.

faqs/galaxy/tools_missing.md
garimavs: > @garimavs I understand your reasoning, but this FAQ is added to every individual tutorial FAQ pages, where people who are following a tutorial will go when they get stuck, and this is one of the most common questions, so it would be best if only the most relevant answers for that audience are here.
> 
> Having a (different) FAQ as a landing page for all "missing tool" questions, like you suggest, could be a good idea, then that can link out to this and other FAQs, e.g. "Can't find a tool for a GTN tutorial?" (link to this one). as well as some of these other cases you describe, such as requesting installation of toolshed tools to a server, or adding new tools to the toolshed.

Yes, indeed, we can work on something along the same lines.

jennaj:
This version looks good to me

beatrizserrano (approved):

garimavs:
> thanks a lot @garimavs! (and @shiltemann and @jennaj of course!)

Thanks to you all also for assisting me throughout the process!

PR #3315, opened 2022-03-30
Moving FAQ - Updating user information

Melkeb:


jennaj:
This is good

Melkeb:
Thanks for the comments. Forgot Password is not included in this FAQ because it looked like a different issue that updating/ changing user preferences. I will add common phrases on the FAQ that people might search for.

Melkeb:
I updated the area, and included "email changes" in a sentence as well as information about updating current account at the bottom of the FAQ.

beatrizserrano (approved):
This PR looks good to me, so unless I'm missing something (@shiltemann @hexylena), we could merge :) Thank you @Melkeb 🎉

PR #3317, opened 2022-03-30
Moving FAQ - Moving data between Galaxy servers

Melkeb:


faqs/galaxy/datasets_moving_data_between_galaxy_servers.md
```diff
@@ -0,0 +1,12 @@
...
+title: Moving data between Galaxy servers
+area: datasets
+box_type: tip
+layout: faq
+contributors: [jennaj, Melkeb]
+---
+
+For individual datasets:
+1. Click on name of file (dataset) to expand the info.
+2. Right click on save icon {% icon galaxy-save %} > Copy Link. Note that for BAM files, clicking on save icon {% icon galaxy-save %} provides access to both alignment (BAM) and index (BAI) files. For BAM files use left mouse click to access BAM and index files. Move the cursor over the BAM file and use the right click to copy the link.
```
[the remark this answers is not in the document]
shiltemann: This Note about BAM files could probably just be it's own bullet point. If possible, could you test how/if the link button works for BAM files and how to get the index file? Since I am not sure about that myself

shiltemann (commented):
Thanks @Melkeb!

faqs/galaxy/datasets_moving_data_between_galaxy_servers.md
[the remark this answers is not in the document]
Melkeb: Okay I will do that. Thanks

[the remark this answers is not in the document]
jennaj: 1. The "copy" link for BAM data is just the single BAM file. Other Galaxy server's don't need the index, they make their own index when a BAM is loaded (uploaded, copies, etc)

2. The "disc" icon has two datasets inside it -- for downloading. The BAM and the BAM.bai index. People need both when moving out of Galaxy to command-line or other applications. Not needed for Galaxy-to-Galaxy.

The first used to have to be done with the second (copy the link(s) out of the disc icon). Now it is broken out and pushed up to the link icon (and is still nested under the disk, just takes more clicks..).  I suggested dropping all of the "BAM" specific content here since it isn't needed. Can address how BAM (and other data with indexes) when discussing download or in datatypes. Example: fasta has an index too -- we just don't save it to the history at all (created and deleted at runtime)-- not even available as a nested data unless you specifically create it and save it to the history (never needed in Galaxy, but in case it comes up later --- it happens under pencil > edit attributes > convert > create fa.fai index). 

Moving a single datasets is just using that "link" copy in the "from" galaxy, then a paste into the upload tool at the "to" galaxy. The limitations I know of. Could mention both, or just the first. 

1) cannot move HTML datasets (security reasons) -- can only download
2) cannot move SQLite datasets (the db under the dataset is not what is actually copied) -- and these don't download either afaik

Melkeb:
Hi @shiltemann is there a copy link icon that I can use for this FAQ or should I move on using the disc icon to fetch the link?

shiltemann:
@Melkeb yes, I will add to this PR to make the icon available, will update you

_config.yml
```diff
@@ -111,6 +111,7 @@ icon-tag:
   galaxy-home: fas fa-home
   galaxy-info: fas fa-info-circle
   galaxy-library: far fa-folder
+  galaxy-link: fas fa-link
```
shiltemann, line 114: ok, I have added the icon.

So now you can use {% icon galaxy-link %} to insert the link icon in your FAQ.

(In case you are wondering how this works, the `fas fa-link` refers to the id of the icon in [fontawesome](https://fontawesome.com) )

shiltemann: @Melkeb see above ^ . Let me know if you have any questions!

Melkeb: @shiltemann Thank you. I will add the new icon

Melkeb:
Added the copy link icon and added the limitations it has.

faqs/galaxy/datasets_moving_datasets_between_galaxy_servers.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Moving datasets between Galaxy servers
+area: datasets
+box_type: tip
+layout: faq
+contributors: [jennaj, Melkeb]
+---
+
+For individual datasets:
+1. Click on name of file (dataset) to expand the info.
```
beatrizserrano, line 10: would filename work better here?

beatrizserrano: also, before this first item, we could say something like:

`In the origin server:`

faqs/galaxy/datasets_moving_datasets_between_galaxy_servers.md
```diff
@@ -0,0 +1,14 @@
...
+2. Click on the **Copy link icon**{% icon galaxy-link %}.
+3. On another Galaxy server: click on **Upload data > Paste / Fetch Data** and paste the link. Select attributes, such as genome assembly, if required. Hit the **Start** button.
```
beatrizserrano, line 12:
```suggestion
On the destination Galaxy server:
3. Click on **Upload data > Paste / Fetch Data** and paste the link. Select attributes, such as genome assembly, if required. Hit the **Start** button.
```

faqs/galaxy/datasets_moving_datasets_between_galaxy_servers.md
Melkeb: I changed it to name of dataset to make it more clearer and added the first line about the origin server similar as the destination server text.

beatrizserrano: ok, thanks! that makes sense

beatrizserrano (approved):
thanks a lot @Melkeb!

PR #3318, opened 2022-03-30
Moving FAQ - Download Entire History

Melkeb:


jennaj:
This is good and simple. 

We might want to be very clear that this is just for _directly downloading_, does not cover how to move a history galaxy-to-galaxy (the first part is the same except the link is copied, but the next part about import history is missing -- happens under User > History > Import -- with three different options, and one is a link from another server).

beatrizserrano (approved):
thanks @Melkeb!

PR #3320, opened 2022-03-30
Moving FAQ - Tool doesn't recognize any input datasets

Melkeb:
When the tool doesn't recognize dataset due to the datatype format.


faqs/galaxy/tools_tool_doesn't_recognize_any_input_datasets.md
```diff
@@ -0,0 +1,16 @@
...
+box_type: tip
+layout: faq
+contributors: [jennaj, Melkeb]
+---
+
+
+This is usually a simple datatype assignment incompatibility between the dataset and the tool.
+Expected input datatype format is explained on the Tool form itself on the "Please provide a value for this option" section or on the information section below the Execute button.
+
+1. Click on the pencil icon {% icon galaxy-pencil %} found on the dataset to reach the **Edit Dataset Attributes** form.
```
shiltemann, line 13: Theses 4 steps are basically the same as the existing ["Changing datatype" FAQ](https://training.galaxyproject.org/training-material/faqs/galaxy/datasets_change_datatype.html)

The other common reason the tool won't suggest the desired input, is that they want to input a collection, but have not clicked on the button for that (https://training.galaxyproject.org/training-material/faqs/galaxy/tools_select_collection.html)

Maybe we can rewrite this FAQ to point to (or include directly) the existing datatype changing FAQ to avoid duplication, and same for the collection input

jennaj:
This needs some work and I'll adjust it. There are tips to determine what datatype an input field is expecting, to "redetect" a datatype to confirm it is correct (didn't find that anywhere yet), and changing the datatype with "convert" is usually a very bad idea (reassigning is better). 

Hold this for me please

shiltemann:
Thanks for looking at these @jennaj! I would lean towards making smaller FAQs for each of those points (converting, (re)detecting, changing datatype, collection input, etc), and then one bigger FAQ about what to do if your input isn't recognized that lists all these possibilities but just links out to them instead of having all the instructions for all those possibilities in it. What do you think?

jennaj (approved):

jennaj:
@Melkeb Thank you for following the advice for the change! I adjusted the content a bit to include more sub-FAQs. Is now a complete "all in one place" summary of the problem + common solutions.

@beatrizserrano -- waiting to see if the changes work. I'm not sure how to add in a link to an FAQ section instead of a specific FAQ.

Looks like that just failed. I'll come back to this -- maybe someone else could review and fix that link? I can learn from that example.

faqs/galaxy/tools_tool_doesn't_recognize_any_input_datasets.md
```diff
@@ -0,0 +1,22 @@
...
+
+The expected input datatype assignment is explained on the tool form. Review the input select areas and the help section below the Execute button. 
+
+Understanding datatypes [FAQ]({% link faqs/galaxy/datatypes_understanding_datatypes.md %}).
+
+No datasets or collections available? Solutions:
+1. Upload or Copy an appropriate dataset for the input into the active history.
+* To review options for getting data into Galaxy see [this FAQ]({% link faqs/galaxy/#data%20upload %}).
+* To copy datasets from a different history into the active history see [this FAQ]({% link faqs/galaxy/histories_copy_dataset.md %})
+3. Resolve a datatype assignment incompatibility between the dataset and the tool.
```
beatrizserrano, line 17: the numbers are not consecutive here :)

faqs/galaxy/tools_tool_doesn't_recognize_any_input_datasets.md
```diff
@@ -0,0 +1,22 @@
...
+contributors: [jennaj, Melkeb]
+---
+
+The expected input datatype assignment is explained on the tool form. Review the input select areas and the help section below the Execute button. 
+
+Understanding datatypes [FAQ]({% link faqs/galaxy/datatypes_understanding_datatypes.md %}).
+
+No datasets or collections available? Solutions:
+1. Upload or Copy an appropriate dataset for the input into the active history.
+* To review options for getting data into Galaxy see [this FAQ]({% link faqs/galaxy/#data%20upload %}).
```
beatrizserrano, line 15: Could be that this link is failing? This is the error:
`File contains invalid characters or sequences: faqs/galaxy/tools_tool_doesn't_recognize_any_input_datasets.md Valid syntax: {% snippet file.ext param='value' param2='value' %} in /home/runner/work/training-material/training-mateqrial/_layouts/faq-page.html`

What I understand from that is that `#data%20upload` should be a valid `.md` file

Melkeb: I'm not sure I understand how to fix this link tried adding `.md` at the end and also tried `dataupload.md` but it's still failing.

jennaj: The link is for the header of an FAQ section, not a specific FAQ.

Do we know how to link that? I wasn't sure and put notes in the prior changes about that.

jennaj (approved):
Ok, this is good. The general upload FAQ is enough -- people can find other options (FTP, etc) by searching or scrolling around .

jennaj:
Nope, that didn't work either. There isn't a dataupload.md -- there is this path: https://training.galaxyproject.org/training-material/faqs/galaxy/#data%20upload

We must have a way of internally referencing sections of the FAQs in other FAQs, yes? That's important. Not everything will be 1-1 except for items explicitly pointed to. 

I guess we could make one more FAQ and name it dataupload and list out all the known ways to upload that already have FAQs (some are in other sections. example: "datatypes" has a bunch of FAQs about fastq/fastq data upload.

Thoughts? This looks like the very last one.

jennaj (approved):
Test with fixed up links.

jennaj:
@beatrizserrano I made some more changes but this still fails. The two links this FAQ cannot find do exist. A rebase is probably the right choice.

beatrizserrano:
It is really strange because now we get errors in 2 files:
- faqs/galaxy/datasets_change_datatype.md (this was there before)
- faqs/galaxy/datasets_import_from_data_library.md (this is new)

There must be something else that we are missing, maybe something was broken in the PRs that I merged before this one. I think that the PR is correct.

beatrizserrano:
Not sure a rebase would fix it, because the supposedly missing file is in Melkeb's branch too: https://github.com/Melkeb/training-material/blob/Melkeb-Unrecognized-tool/faqs/galaxy/datasets_change_datatype.md

PR #3321, opened 2022-03-30
Made some minor changes to GTN FAQs

Melkeb:
- Changed Import an History to Importing History
- Removed the space to delete the unnecessary name info seen on the image
Capture

### DIFF
--- a/faqs/galaxy/collections_rename.md
+++ b/faqs/galaxy/collections_rename.md
@@ -4,7 +4,6 @@ area: collections
 box_type: tip
 layout: faq
 contributors: [delphine-l, martenson, shiltemann, hexylena]
-
 ---
 
 
--- a/faqs/galaxy/histories_import.md
+++ b/faqs/galaxy/histories_import.md
@@ -1,5 +1,5 @@
 ---
-title: Import an history
+title: Importing history
 area: histories
 box_type: tip
 layout: faq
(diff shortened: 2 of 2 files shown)


faqs/galaxy/collections_rename.md
```diff
@@ -4,7 +4,6 @@ area: collections
 box_type: tip
 layout: faq
 contributors: [delphine-l, martenson, shiltemann, hexylena]
-
```
shiltemann, line 7: huh, this empty line was causing the bad rendering? nice catch! We should probably add a check for that to the github testing framework.

faqs/galaxy/histories_import.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Import an history
+title: Importing history
```
shiltemann, line 2: I would make it either "Importing histori**es**" (plural), or "Importing **a** history"

Melkeb: Yeah to make it consistent with the other questions on the GTN history section I will use the second one. Thanks

faqs/galaxy/collections_rename.md
Melkeb: This should be tested though if it actually makes a difference and removes the unnecessary info. I didn't see anything wrong with it besides this inconsistency compared to the other files.

shiltemann: I just built the website locally and it is indeed fixed, so nice find :+1:

shiltemann (approved):

shiltemann:
Thanks @Melkeb! :tada:

PR #3379, opened 2022-04-13
[FAQ] Added Working with fastq dataset

garimavs:
This PR is to create an FAQ regarding _Working with huge fastq dataset_ and further link it with all of its occurrences.

Thanks to @jennaj for working on this.


jennaj (approved):
Perfect!

PR #3390, opened 2022-04-14
Fixing FAQ dataset not downloading at all

Melkeb:
Fixing the problem shown on the image (the first three steps).
Issue 1


beatrizserrano (approved):
that's a good catch, thanks for reviewing them!

PR #3394, opened 2022-04-15
Moving FAQ - Mismatched Chromosome identifiers and how to avoid them

Melkeb:
This FAQ is part of [Mismatched Chromosome identifiers (and how to avoid them)](https://galaxyproject.org/support/chrom-identifiers/). I separated the FAQ to different FAQs because the content was too long.

### DIFF
--- a/faqs/galaxy/datasets_chromosome_identifiers.md
+++ b/faqs/galaxy/datasets_chromosome_identifiers.md
@@ -11,11 +11,11 @@ contributors: [jennaj, Melkeb]
 
 - **If using a Custom Reference genome**, the methods below also apply, but the first step is to make certain that the Custom Genome is formatted correctly. Improper formating is the most common root cause of CG related errors.
 
-Method 1: [Find BAM dataset identifiers]({% link faqs/galaxy/datasets_BAM_dataset_identifiers.md %})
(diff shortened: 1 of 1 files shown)


jennaj (approved):
I think it is OK to have broken links to FAQs that will also be approved soon. At the very end of the process we can go through all and check/repair any broken crosslinks. Otherwise, we'll lose context and have much more to do later.

Melkeb:
Noted! Thank you @jennaj for the comments.

Melkeb:
Hi @annefou should I remove method 3 and 4 since its linked to the closed PR or should we close this one as well?

annefou:
I would. For me these "types" of FAQs are more for topics FAQs than for GTN FAQs.

beatrizserrano:
The broken link is really not broken, we merged before: https://github.com/galaxyproject/training-material/blob/main/faqs/galaxy/datasets_BAM_dataset_identifiers.md

So I think it's safe to merge this one :) Thank you @Melkeb and @jennaj

beatrizserrano:
ok, so I was very fast merging, the first link that was broken according to the tests was ok, but not the rest. @Melkeb could you edit this FAQ in another PR? thank you :)

Melkeb:
> ok, so I was very fast merging, the first link that was broken according to the tests was ok, but not the rest. @Melkeb could you edit this FAQ in another PR? thank you :)

Yeah sure I will

PR #3397, opened 2022-04-15
Moving FAQ - Adjusting identifiers for UCSC sourced data used with other sourced data

Melkeb:
This FAQ is part of [Mismatched Chromosome identifiers (and how to avoid them)](https://galaxyproject.org/support/chrom-identifiers/)

### DIFF
--- a/faqs/galaxy/datasets_adjust_identifiers_UCSC.md
+++ b/faqs/galaxy/datasets_adjust_identifiers_UCSC.md
@@ -18,7 +18,7 @@ For **bed** data:
 1. Tool **Add column**: add "chr" to the original dataset as a new column.
 2. Tool **Merge Columns**: merge "c7" with "c1".
 3. Tool **Cut**: cut "c8,c2,c3,c4,c5,c6" (replace c1 & c7 - with merged c8 - the new chrom identifier).
-4. Click on the pencil icon for the result dataset, then the tab for "Datatype". Assign "bed" and save. Allow the metadata to complete assignment (the "yellow" dataset state).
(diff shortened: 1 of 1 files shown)


faqs/galaxy/datasets_adjust_identifiers_UCSC.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Adjusting identifiers for UCSC sourced data used with other sourced data
+area: datasets
+box_type: tip
+layout: faq
+contributors: [jennaj, Melkeb]
+---
+
+
+- A GTF formatted dataset (potentially a "reference annotation dataset"), with Ensembl/UCSC/Other based chromosome identifiers, is to be used with a tool that also makes use of a different sourced reference genome or the reverse may be true, Ensembl/UCSC/Other sourced reference genome and a differnt sourced reference annotation.
```
beatrizserrano, line 10:
```suggestion
- A GTF formatted dataset (potentially a "reference annotation dataset"), with Ensembl/UCSC/Other based chromosome identifiers, is to be used with a tool that also makes use of a different sourced reference genome or the reverse may be true, Ensembl/UCSC/Other sourced reference genome and a different sourced reference annotation.
```

beatrizserrano (commented):
I don't understand this FAQ well. @jennaj or someone else, could you review it?

jennaj:
Do the same as this one: https://github.com/galaxyproject/training-material/pull/3398

Then we drop this and the hub section points to Galaxy FAQs (top level). Should count as a "done" FAQ. Removing old content is important, too, not all will be migrated but the old hub FAQs can be cleaned up with a redirect.

annefou:
This PR can be closed too. Thanks @Melkeb for your effort.

PR #3422, opened 2022-04-21
Moving Managing datasets FAQ

mruqqie:


jennaj:
Apologies, but this FAQ should not go into the GTN. Most of this content is already covered in other FAQs that are being migrated. This is a nice effort for an additional narrative style FAQ but might be dropped completely. We can leave the PR open -- will need heavy editing if retained later (likely not -- or perhaps a few sub-sections only).

jennaj:
Decided to close this. 

The area in the source FAQ can link in the hub to Galaxy FAQs in the GTN directly (but not a specific FAQ, instead just the top level). It should count as a "done" FAQ once the hub is updated.

PR #3426, opened 2022-04-22
Fixing FAQ - Mismatched Chromosome identifiers and how to avoid them

Melkeb:
I changed the first two links to the GTN FAQ links (Even though the first one was working changed both to make it consistent). The last two links added lead to the Galaxy Hub FAQ Page since it has no GTN FAQ link (If it shouldn't lead to the Galaxy Hub FAQ page then I can remove the link).


beatrizserrano (approved):
that was superfast @Melkeb, thanks a lot!

PR #3432, opened 2022-04-25
CYOA: Set a default, and allow taking it from a URL parameter

hexylena:
@bebatut @lldelisle it now supports a default, which is set for your tutorial, and we can choose which branch based on the URL parameter (probably good for linking to a specific version in a course)

### DIFF
--- a/_includes/cyoa-choices.html
+++ b/_includes/cyoa-choices.html
@@ -1,5 +1,4 @@
-
-<blockquote class="notranslate hands_on gtn-cyoa">
+<blockquote class="notranslate hands_on gtn-cyoa" id="gtn-cyoa">
   <h3><i class="fas fa-pencil-alt" aria-hidden="true"></i><span class="visually-hidden">hands_on</span> Hands-on: Choose Your Own Tutorial</h3>
 	<p>This is a "Choose Your Own Tutorial" section, where you can select between multiple paths. Click one of the buttons below to select how you want to follow the tutorial</p>
 	{% if include.text %}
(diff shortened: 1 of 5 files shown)


lldelisle:
Thank you so much

hexylena:
> Thank you so much

Feel free to approve if it's ok :)

lldelisle:
I am not granted for merging...
Is it expected to have this indentation?
image
Maybe I would remove some space but it is really minor.

hexylena:
>  Is it expected to have this indentation?

yes, otherwise it wouldn't fit in most browsers, and people would miss the last "text" parameter, so I separated it over multiple lines. Maybe there's a better way.

> I am not granted for merging...

I thought you were! If you do a review → approve, then it should let you. (And feel free to make suggestions as you like!)

image

lldelisle (approved):

lldelisle:
I thought I was only granted for epigenetics...

hexylena:
Nope! You have the power now :muscle: please use it wisely :smile:

lldelisle:
Should I delete the branch? Or do you usually keep them?

hexylena:
Deleting is fine for merged things!

lldelisle:
> Nope! You have the power now muscle please use it wisely smile

With great power comes great responsibility

lldelisle:
I tried to use titles but I did not managed it seems that the left panel is parsed differently.

hexylena:
> @hexylena Is it easy to add a flag in the left panel which tells you which option of the tutorial you are looking at? How the left panel is generated?
> Thanks

It is technically possible. The left panel is mostly JavaScript, but we could add elements to it dynamically when were in a CYOA tutorial.

Would you want to be able to change which one you're on? From the left panel? Or just see it, and maybe be able to scroll back to the section where you choose?

I've also thought we could highlight the different sections, my original implementation years ago had a coloured bar for any sections that were specific to a single track.

lldelisle:
If you can choose from the left panel this would be awsome but if you can see it and go back to the place where you choose it is already great.
I noticed that if you are using it from inside usegalaxy.eu with the non default, sometimes (maybe when you go to data libraries and come back) it put back the default without any notice.

lldelisle:
If it is easier that you only choose from the left panel it is also fine.

PR #3435, opened 2022-04-28
Ref-based RNA-seq try to fix cyot default

lldelisle:
Hi,
I noticed a "bug" (at least a non expecting behaviour).

If you 

1. go to galaxy
2. open the Ref-based tutorial
3. go to the Counting the number of reads per annotated gene.
4. Select the "STAR" flavour of the tutorial.
5. Click outside of the box, go to 'view all histories', come back to your current history, open back the tutorial.
6. It is back to the "featureCounts" flavour.

It seems that it stored the position where you were in the tutorial but not the choice.
This is probably what happened to: https://github.com/galaxyproject/training-material/issues/989#issuecomment-1110204833
And this happened to us during the training.

The small fix I propose here is to use the URL but I don't know if there is a way to hide it behind the "STAR" button of the Choose your own tutorial. Also I am not sure it will work, is there a way to test?
@hexylena if you can help...

Thanks,

### DIFF
--- a/topics/transcriptomics/tutorials/ref-based/tutorial.md
+++ b/topics/transcriptomics/tutorials/ref-based/tutorial.md
@@ -649,6 +649,15 @@ Therefore we offer a parallel tutorial for the 2 methods which give very similar
 
 In principle, the counting of reads overlapping with genomic features is a fairly simple task. But there are some details that need to be given to **featureCounts** or to the output of **STAR**, e.g. the strandness.
 
+<div class="STAR" markdown="1">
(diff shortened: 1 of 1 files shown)


hexylena:
@bebatut will teach this next week, I will try and get this fixed before then

lldelisle:
If she is teaching it with the featureCount flavour, she will probably not see any issue. The issue comes when we use the non-default tutorial.

bebatut:
Yes I will teach with default version

hexylena:
Fixing this in a more generic way in https://github.com/galaxyproject/training-material/pull/3451

PR #3437, opened 2022-04-28
Fixing FAQ Content

Melkeb:
Fixing - How can I reduce quota usage while still retaining prior work (data, tools, methods)? FAQ

Replaced the copy link since it did not address moving datasets or histories to another server instead it addressed copying datasets between histories on the same galaxy server and was the same as the the third bullet point of the [FAQ](https://training.galaxyproject.org/training-material/faqs/galaxy/reduce_quota_usage.html).


lldelisle (approved):
Great, thanks.

PR #3441, opened 2022-04-28
Ref-based RNA-seq: Minor changes post training

lldelisle:
- We noticed that the flatten collection was difficult to understand. I added a scheme. Also as it is a problem with multiQC it seems more logical to put it just before it.
- I noticed that the gene body coverage is really really slow so I do a subsampling before.


topics/transcriptomics/tutorials/ref-based/tutorial.md
```diff
@@ -166,13 +163,42 @@ Sequence quality control is therefore an essential first step in your analysis.
...
+>    > 
+>    > 
+>    > Desired organisation
+>    >                                       FastQC results
+>    >                                     /     /    \    \
+>    >                                 /       /        \      \
+>    >                             /         /            \        \  
+>    >                         /           /                \          \  
+>    >                     /             /                    \            \
+>    > GSM461177_.._forward   GSM461177_.._reverse   GSM461180_.._forward   GSM461180_.._reverse
```
hexylena, line 188: I hate to ask for more work, but what about an image/chart here, it can be something simple in inkscape or lucidchart or similar.

lldelisle: Yes good idea

hexylena (approved):

PR #3451, opened 2022-05-03
Store CYOA choice in localstorage

hexylena:
replaces https://github.com/galaxyproject/training-material/pull/3435

This *should* survive reloads, as it uses the same mechanism as themes (which do).

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -22,6 +22,8 @@ $("section.tutorial .hands_on").append('<p class="text-muted" style="text-align:
 // CYOA Support
 function cyoaChoice(text){
 	if(text !== undefined && text !== null){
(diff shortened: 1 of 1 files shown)


hexylena:
@lldelisle we can put it in the left menu, but I cannot seem to prevent it from scrolling, due to how the CSS is done, so unfortunately I cannot address this issue currently.

image

lldelisle:
> @lldelisle we can put it in the left menu, but I cannot seem to prevent it from scrolling, due to how the CSS is done, so unfortunately I cannot address this issue currently.
> 
> image

Thanks for trying.

lldelisle (approved):
This seems to work... Thanks @hexylena

PR #3459, opened 2022-05-09
Add a nicer stats display

hexylena:
Always bemoaned that our display wasn't as nice as grafana's

image

and also in slides

image

so the image is no longer constantly outdated.


shiltemann (approved):

shiltemann:
love it, thanks!!

PR #3466, opened 2022-05-15
[FAQ] Edited file names.

garimavs:
Through this PR, I have edited the name of files, as their _area name_ and _file name_ differ.


garimavs:
I think the last file name was used somewhere else. I need to change the name there, too, though I'm unable to spot where.
Kindly help @beatrizserrano @hexylena.

beatrizserrano:
> I think the last file name was used somewhere else. I need to change the name there, too, though I'm unable to spot where. Kindly help @beatrizserrano @hexylena.

I think that the issue may be that there are plenty of tutorials using `datasets_import_via_link.md`, so that would need to be replaced by `dataupload_import_via_link.md`

There seems to be 126 places where `datasets_import_via_link.md` is used. See here: https://github.com/galaxyproject/training-material/search?p=1&q=datasets_import_via_link

hexylena:
I do not think we want to rename these two. It's fine for them to not match, and it's not worth the huge number of changes we'd need to make, sorry @garimavs.

garimavs:
Sure, thanks for the help @beatrizserrano @hexylena!

PR #3471, opened 2022-05-20
Document all schemas

hexylena:
We've got a number of schemas (tutorial, topic, faq, contributor, slides) and they weren't really documented so if you wanted to use a feature, how would you know? You wouldn't. Now they're available as beautiful documentation, and integrated within our existing documentation as well, which is an improvement.

Please ignore the missing icons - train wifi.
image


hexylena:
ready to review if anyone's up for it.

shiltemann (approved):

PR #3477, opened 2022-05-31
Admin FAQs

hexylena:
- Add an FAQ for a custom welcome page (per discussion in matrix)
- Move some of our existing tips into FAQs for better visibility
- Add a checker script for incorrectly referenced `snippet`s since the jekyll message is useless ("could not find file ''")

### DIFF
--- a/.github/workflows/ci-symlinks.yml
+++ b/.github/workflows/ci-symlinks.yml
@@ -1,4 +1,4 @@
-name: Lint for Broken Symlinks
+name: Lint for Broken Symlinks, FAQs
 
 concurrency:
   group: ci-symlinks-${{ github.head_ref }}
@@ -13,6 +13,13 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
+
       - name: Check for broken symlinks
+        if: always()
         run: |
           bash bin/check-broken-symlinks.sh
+
+      - name: Check for broken symlinks
+        if: always()
+        run: |
+          bash bin/check-snippet-links.sh
--- a/.github/workflows/ci-symlinks.yml
+++ b/.github/workflows/ci-symlinks.yml
@@ -19,7 +19,7 @@ jobs:
         run: |
           bash bin/check-broken-symlinks.sh
 
-      - name: Check for broken symlinks
+      - name: Check for broken snippet links
         if: always()
         run: |
           bash bin/check-snippet-links.sh
--- a/topics/admin/faqs/ansible_define-once.md
+++ b/topics/admin/faqs/ansible_define-once.md
@@ -27,4 +27,4 @@ galaxy_config_templates:
     dest: "{{ galaxy_config.galaxy.datatypes_config_file }}"
 ```
 
-Practices like those shown above helps to avoid problems caused when paths are defined differently in multiple places. The datatypes config file will be copied to the same path as Galaxy is configured to find it in, because that path is only defined in one place. Everything else is a reference to the original definition! If you ever need to update that definition, everything else will be updated accordingly.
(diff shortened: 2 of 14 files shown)


.github/workflows/ci-symlinks.yml
```diff
@@ -13,6 +13,13 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
+
       - name: Check for broken symlinks
+        if: always()
         run: |
           bash bin/check-broken-symlinks.sh
+
+      - name: Check for broken symlinks
```
nsoranzo, line 22:
```suggestion
      - name: Check for broken snippet links
```

topics/admin/faqs/ansible_define-once.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Define once, reference many times
+area: ansible
+box_type: tip
+layout: faq
+contributors: [natefoo, shiltemann]
+---
+
+Using practices like those shown above helps to avoid problems caused when paths are defined differently in multiple places. The datatypes config file will be copied to the same path as Galaxy is configured to find it in, because that path is only defined in one place. Everything else is a reference to the original definition! If you ever need to update that definition, everything else will be updated accordingly.
```
nsoranzo, line 9: The first sentence doesn't probably make a lot of sense outside of the original tutorial, but not sure how to rephrase it.

hexylena: Good point. Thanks for looking at it!

hexylena: ok, rewrote a bit to be a bit more generic. Whatdya think?

topics/admin/faqs/ansible_define-once.md
```diff
@@ -6,4 +6,25 @@ layout: faq
...
+# and now we can re-use "{{ galaxy_config.galaxy.datatypes_config_file }}"
+# in other places!
+
+
+galaxy_config_templates:
+  - src: templates/galaxy/config/datatypes_conf.xml
+    dest: "{{ galaxy_config.galaxy.datatypes_config_file }}"
+```
+
+Practices like those shown above helps to avoid problems caused when paths are defined differently in multiple places. The datatypes config file will be copied to the same path as Galaxy is configured to find it in, because that path is only defined in one place. Everything else is a reference to the original definition! If you ever need to update that definition, everything else will be updated accordingly.
```
nsoranzo, line 30:
```suggestion
Practices like those shown above help to avoid problems caused when paths are defined differently in multiple places. The datatypes config file will be copied to the same path as Galaxy is configured to find it in, because that path is only defined in one place. Everything else is a reference to the original definition! If you ever need to update that definition, everything else will be updated accordingly.
```

topics/admin/faqs/ansible_define-once.md
```diff
@@ -6,4 +6,25 @@ layout: faq
 contributors: [natefoo, shiltemann]
 ---
 
-Using practices like those shown above helps to avoid problems caused when paths are defined differently in multiple places. The datatypes config file will be copied to the same path as Galaxy is configured to find it in, because that path is only defined in one place. Everything else is a reference to the original definition! If you ever need to update that definition, everything else will be updated accordingly.
+Using variables, either by defining them ahead of time, or simply accessing them via existing data structures that have been defined
```
nsoranzo, line 9: Looks like the sentence was truncated?

hexylena: that was the end, added an `e.g. :` to make clear it means the example that follows.

nsoranzo (approved):

hexylena:
merci!

PR #3511, opened 2022-06-24
Fix typo

wee-snufkin:
The question asks about dataset GSM461176, while the answer is for GSM461177.


hexylena (approved):
Great catch! Thank you!

PR #3512, opened 2022-06-24
Fix errors

wee-snufkin:
'Filter empty' image not loading properly.
'Keep first instance' described twice - the second description definitely refers to 'Keep last instance'


topics/galaxy-interface/tutorials/collections/tutorial.md
```diff
@@ -311,7 +311,7 @@ The tool allow extracting datasets based on position (**The first dataset** and
 
 This tool takes a dataset collection and filters out (removes) empty datasets. This is useful for continuing a multi-sample analysis when downstream tools require datasets to have content.
 
-.. image:: ${static_path}/images/tools/collection_ops/filter_empty.svg
+.. image:: ${static_path}/images/tools/collection_ops/filter_empty.svg 
   :width: 500
   :alt: Filtering empty datasets
```
hexylena, line 316: I don't think this image syntax is supported, it looks like restructured text and I'm not sure how it got in here. Would you mind 
- rewriting it to use the markdown image syntax `alt-text`
- I think the correct image path is `../../images/collections/filter_empty.svg`
- we don't need to retain the `width: 500` parameter.

hexylena (dismissed):

hexylena (approved):

PR #3515, opened 2022-06-27
replace uWSGI with Gunicorn

gmauro:
Currently, the training doesn't work, as it's using uWSGI settings for galaxy and Gunicorn settings for NGINX.
I updated it to be fully Gunicorn complaint.
As soon as https://github.com/galaxyproject/ansible-galaxy/issues/160 will be closed, I can finish updating it.


hexylena:
Yep, we got most of this in the last PR which updated everything but the roles. Thanks for finalising it! I'll have a look at this tomorrow

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -1752,15 +1769,14 @@ Galaxy is now configured with an admin user, a database, and a place to store da
...
 >    {: .tip}
 >
 > 6. Some things to note:
 >
->    1. Later, after NGINX is set up and we're ready to connect to Galaxy, refreshing the page while Galaxy is restarting will wait (spin) until the process is ready rather than produce an error message, a nice feature of uWSGI
->    2. Although the playbook will restart Galaxy upon config changes, you will sometimes need to restart it by hand, which can be done with `systemctl restart galaxy`
->    3. You can use `journalctl -fu galaxy` to see the logs of Galaxy
->    4. You may have noticed after setting up Galaxy in the previous section, that Ansible printed the message "RESTARTER NOT IMPLEMENTED - Please restart Galaxy manually. You can define your own handler and enable it with `galaxy_restart_handler_name`." You should no longer see this message on future playbook runs, as enabling systemd has automatically set `galaxy_restart_handler_name` for us - the role now knows how to restart Galaxy!
+>    1. Although the playbook will restart Galaxy upon config changes, you will sometimes need to restart it by hand, which can be done with `systemctl restart galaxy`
```
Slugger70, line 1777: Do we need to use gravity here? Or are we going to wrap gravity within systemd? 

I like using Gravity - the way I do it is:
1. `sudo su - galaxy`
2. `source /srv/galaxy/venv/bin/activate`
3. `gravityctl restart` or `gravityctl graceful`

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -1752,15 +1769,14 @@ Galaxy is now configured with an admin user, a database, and a place to store da
...
+>    2. You can use `journalctl -fu galaxy` to see the logs of Galaxy
```
Slugger70, line 1778: This then becomes `gravityctl follow`

topics/admin/tutorials/ansible-galaxy/tutorial.md
gmauro: I prefer to use systemd; it's a tool that people can know how to use before, can reuse after the training, can be used for other tools like NGINX.
In any case,  you can ever use gravityctl if you like but, in my opinion, the training should try to keep the access barrier low.

Slugger70: Fair enough. :)

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -1685,7 +1699,9 @@ Galaxy is now configured with an admin user, a database, and a place to store da
 >               - workflow-scheduler
 >    +
 >    +# systemd
->    +galaxy_manage_systemd: yes
+>    +galaxy_manage_systemd: true
+>    +galaxy_manage_gravity: "{{ false if __galaxy_major_version is version('22.01', '<') else true }}"
+>    +galaxy_systemd_mode: "{{ 'mule' if __galaxy_major_version is version('22.01', '<') else 'gravity' }}"
```
hexylena, line 1704: This shouldn't be user controlled, we don't want that. It should be in the role.

topics/admin/tutorials/ansible-galaxy/tutorial.md
hexylena: yeah I think I'm fine keeping this systemctl/journalctl, especially in the context of: pulsar, slurm, etc all using it.

topics/admin/tutorials/ansible-galaxy/tutorial.md
gmauro: the role currently default to 22.05. Should I pr the role then?

hexylena: hm I missed that it was 22.01 here, we want to use gravity starting with 22.05, not if it's <22.01. The playbook is right, any reason to override that here? You want to deploy 22.01 with mules/gravity?

I'm not sure we want to teach that in the training materials, I think these should be left default, and we're going to teach 22.05 this summer.

hexylena:
`0.10.1` is released.

hexylena (approved):

PR #3518, opened 2022-06-30
Reference-based RNA-Seq data analysis: fix DEG workflow

gallardoalba:
This PR includes the following changes:

- modifies the DESeq2 factor level groups in order to fix the following error:

`parameter 'groups': invalid options ('paired') were selected (valid options: untreat,treat,single.counts,paired.counts)
`
- changes the output format to *tabular* of the Concatenate output in order to fix the following error:

`
parameter 'field2': requires a value, but no legal values defined
`

### DIFF
--- a/topics/transcriptomics/tutorials/ref-based/workflows/deg-analysis.ga
+++ b/topics/transcriptomics/tutorials/ref-based/workflows/deg-analysis.ga
@@ -20,17 +20,17 @@
             "name": "Input dataset collection",
             "outputs": [],
             "position": {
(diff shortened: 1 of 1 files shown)


hexylena (approved):
I can't wait til the IWC comes up with nice ideas to make these easier to review :)

gallardoalba:
> I can't wait til the IWC comes up with nice ideas to make these easier to review :)

I could replace it with the text editor instead of using the workflow editor, sorry!

PR #3519, opened 2022-06-30
Update architecture tutotial slides (model+migrations)

jdavcs:
Ping @jmchilton @hexylena


hexylena:
@jmchilton how would you feel about the idea of splitting these up into separate slide decks, based on your existing YouTube splits?

PR #3520, opened 2022-06-30
Suggetions peaks2genes tutorial

teresa-m:
Added 3 small changes to the update that I found to be a bit different while doing the tutorial. Hope that helps and that they are in the correct format.


hexylena (approved):
These look like good changes, thanks @teresa-m !

PR #3539, opened 2022-07-12
edited VGP workflow tutorial

abueg:
removed parameters from hifiasm-hic hands-on that are no longer exposed in current workflows


gallardoalba (dismissed):

topics/assembly/tutorials/vgp_workflow_training/tutorial.md
```diff
@@ -207,9 +207,6 @@ After genome profiling, the next step is to run the **VGP HiFi phased assembly w
 >   - {% icon param-file %} "*HiC forward reads*": `3: Hi-C_dataset_F`
 >   - {% icon param-file %} "*HiC reverse reads*": `2: Hi-C_dataset_R`
 >   - {% icon param-file %} "*Genomescope summary dataset*": `19: Genomescope on data 13 Summary`
->   - "*K-mer length*": `31`
->   - "*Ploidy*": `2`
->   - "*Is genome large (>100Mb)?*": `No`
```
gallardoalba, line 212: We need to use this parameter for the yeast, so probably is not a good idea to remove it. Perhaps to include some information about why it is necessary in that case?

abueg: Hmm, how about I put those parameters mentioned in a sentence that they are set under the hood? I just tried to remove it because someone going through the tutorial was confused when the workflow input page did not have those options available anymore

gallardoalba: You are right, this is the workflow version; so I'll include this change, but for now let it open until the workflow is updated. Otherwise, the results will diverge.

abueg: OK! 👍🏼

gallardoalba:
I still need to make some changes in the workflow sorry. I'll try to work on it next week.

PR #3560, opened 2022-07-21
Add instructions to make workflow accessible in the creating tutorial tutorial

natefoo:
For completeness, technically you do have to publish it until galaxyproject/galaxy#14348 is deployed.


hexylena (approved):
Groundbreaking. Earth shattering. Fantastic work, 11/10.

PR #3567, opened 2022-07-28
Apollo for eukaryote genomes

abretaud:
Hey!
A new tutorial for Apollo, somewhat similar to the existing one except:
- uses a eukaryote genome (instead of bacteria), the same one as in the repeatmasker and funannotate tutorials
- it means different evident tracks, and isoforms/utr stuff
- I added a CYOA to make the galaxy <-> apollo interaction optional: sometimes I just want to teach Apollo for users and let them use the same org that I have precreated for them

If the CYOA works well I could probably do the same for the other apollo tutorial

A little question: there's no way currently to make the TOC adapt to CYOA choice? I managed to avoid the problem somehow, but it could be useful

I plan to test this new tuto for https://lepidoptera-2022.workshop.inrae.fr in the august 22-26 week

While I was at it, I've also reorganized the genome-annotation topic index page to make it tidier:

image

### DIFF
--- a/topics/genome-annotation/tutorials/repeatmasker/tutorial.md
+++ b/topics/genome-annotation/tutorials/repeatmasker/tutorial.md
@@ -32,7 +32,6 @@ follow_up_training:
   topic_name: genome-annotation
   tutorials:
     - funannotate
-    -
 subtopic: eukaryote
 priority: 1
 ---


abretaud:
Cool, thanks for the merge @bgruening!

hexylena:
Love the use of CYOA!

> A little question: there's no way currently to make the TOC adapt to CYOA choice? I managed to avoid the problem somehow, but it could be useful

Oh gosh, that is not a case I considered. The "agenda" toc is generated at compile time and hard to fix. The javascript toc, we can absolutely rebuild somehow once the cyoa path changes.

PR #3571, opened 2022-08-02
Add tutorial: Evaluating homophily/heterophily of complex networks

cumbof:
This PR defined a new topic (Network Analysis) with a tutorial about how to evaluate the homophilic nature of complex networks with the [honto](https://github.com/cumbof/honto) tool in Galaxy


shiltemann:
Hi @cumbof! Are you planning on creating more tutorials in this topic? We always want to avoid having a long list of topics that only have a few tutorials associated with them. We can also make subtopics on existing topics for example, maybe proteomics in your case?

P.S. At some point in the future we plan to rework the entire concept of topics and use ontologies to add metadata to tutorials and allow for multiple views/groupings on the homepage (by topic, domain, type of data, operation etc)

cumbof:
Hi @shiltemann,
I'm sorry for running late on replying your comment.

I'm not planning to create other tutorials about the same topic at the moment, and I agree with you that it doesn't make sense to have new topics for hosting single tutorials only.

Defining a subtopic under the proteomics topic would be great. Could you please advise on how to do that?
Thanks!

hexylena:
@cumbof please see the (quite short) subtopic documentation here: https://training.galaxyproject.org/training-material/topics/contributing/tutorials/create-new-topic/tutorial.html#adapt-the-metadata-for-your-topic

shiltemann:
@cumbof Just looking at this again, and sorry for the delay. I can help to put it under the proteomics topic in a subtopic, but since it is just a single tool run, and reads a bit like a promotion/announcement of the tool rather than a tutorial, would you like to maybe make it a news item instead? e.g. so that it shows up here: https://training.galaxyproject.org/training-material/news.html ?

Let me know what you think

Issue #3582, opened 2022-08-15
`bundle add webrick` needed for Jekyll on Ruby 3 (and thus Gitpod)


pvanheus:
`make gitpod-serve` fails with current Gitpod because it is Gitpod these days is running off Ruby 3. The error message is:

```
/home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/commands/serve/servlet.rb:3:in `require': cannot load such file -- webrick (LoadError)
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/commands/serve/servlet.rb:3:in `<top (required)>'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/commands/serve.rb:179:in `require_relative'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/commands/serve.rb:179:in `setup'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/commands/serve.rb:100:in `process'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/command.rb:91:in `block in process_with_graceful_fail'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/command.rb:91:in `each'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/command.rb:91:in `process_with_graceful_fail'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/jekyll-4.2.1/lib/jekyll/commands/serve.rb:86:in `block (2 levels) in init_with_program'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/mercenary-0.4.0/lib/mercenary/command.rb:221:in `block in execute'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/mercenary-0.4.0/lib/mercenary/command.rb:221:in `each'
        from /home/gitpod/.rvm/gems/ruby-3.1.2/gems/mercenary-0.4.0/lib/mercenary/command.rb:221:in `execute'
[...]

PR #3584, opened 2022-08-16
include webrick as a conditional dependency

martenson:
closes https://github.com/galaxyproject/training-material/issues/3582


martenson:
please review, ping @training-admin

bgruening (approved):
Seeing more and more people struggling with it I will merge it. Maybe @hexylena has a better idea when she is back.

PR #3591, opened 2022-09-02
replace <specs> by <setting>

bernt-matthias:
there is no specs but setting

https://github.com/galaxyproject/galaxy/blob/629a5134cbdf78365a642c960011b33ac3c83ab2/config/plugins/visualizations/example/config/example.xml#L17


hexylena (approved):
I think these are essentially unmaintained and outdated/now not correct, but happy to accept bugfixes.

bernt-matthias:
So you know where more up-to-date docs are found?

PR #3605, opened 2022-09-15
GTN intro for teachers

hexylena:
- new slide deck for teachers as a focus (will be tested this fall, then maybe can make a news post)
- updated stats page for our impressive FAQ count
- made charts more accessible by supporting `prefers-reduced-motion` (I forgot we used animations here during the last a11y pr)
- reorganises the create-new-content-markdown into Boxes and then Not Boxes features, rather than them being mixed.

image


bgruening (approved):
Great, thanks @hexylena

shiltemann:
Looks good, do we already have instructions for how to use quizzes? Would be nice to have that to link to if we mention them in these slides?

hexylena:
Nope, I'll write it up quickly.

shiltemann (approved):

PR #3613, opened 2022-09-20
Quick wireguard/tailscale tutorial

hexylena:


hexylena:
@natefoo since you've deployed already for the gx-it-proxy problem, is there anything you want to add?

natefoo (approved):
FWIW I ended up using tailscale instead of wireguard directly since its free FOSS plan is suitable for our needs. Firewalls were getting in the way of a simple wireguard setup and I just switched to tailscale before bothering trying to work out the STUN stuff. Which is all to say this looks great but my experience is not very useful here. :grin:

PR #3619, opened 2022-09-22
Add support for details type box for FAQs

shiltemann:
This should allow setting `box_type: details` on FAQs per @bebatut request


bebatut:
I think I already did it in the PR with the TtT material...

shiltemann:
are you sure? I don't see it?

bebatut:
My bad, it is the beer tutorial...

shiltemann:
ahh ok, we can wait until that one then, will close this again

PR #3623, opened 2022-09-23
Add a new tutorial about lncRNAs annotation (with FEELnc)

stephanierobin:
It's the first version of a new tutorial about lncRNAs annotation. In this tutorial, we will use Mucor mucedo genome that has been assembled in the 'Genome assembly using PacBio data' tutorial and annotated in the 'Genome annotation with Funannotate' tutorial.


abretaud (approved):
Merci Steph :tada:

PR #3632, opened 2022-09-28
clipseq workflow.ga cutadapt v1.6 -> v4.0

tuncK:
Strictly speaking, now the original version is also functional after Björn's fix.


shiltemann (approved):

shiltemann:
perfect, thanks a lot @tuncK!

PR #3640, opened 2022-10-04
fix galaxy server annotation

shiltemann:
something broke the annotations of supporting Galaxy servers, this seems to fix it @hexylena (fixes https://github.com/galaxyproject/training-material/issues/3607)


hexylena (approved):
It seems this is due to spaces in the filenames (UGH).

THis looks like a good fix.

PR #3659, opened 2022-10-11
Change default job runner to slurm in "Connecting Galaxy to a compute cluster tutorial"

simonbray:
Just going through some of the admin tutorials and it seems to me that this change is needed as well? But feel free to close if I'm wrong.


hexylena:
yeah this looks right if we want it to go there in the next step. Thanks!

hexylena:
We need to update the other diffs, but for some reason it's not working like normal. I'm not sure how to resolve this, but I'll merge this as soon as I've got that sorted.

hexylena (approved):

PR #3662, opened 2022-10-11
Correcting a typo

MarisaJL:
Deleted c in line 589 because it should be adding 1 not adding column 1


hexylena (approved):
Looks great, welcome to the GTN!

Issue #3636, opened 2022-09-30
From peaks to genes training: missing option


gallardoalba:
In the visualization section, the Sort tool requires the following options according with the training:

![Screenshot from 2022-09-30 16-16-54](https://user-images.githubusercontent.com/26678117/193290031-74cbc5a7-e2a7-43b9-a8d2-2674cdfcb88a.png)

However, the *Natural/Version sort (-V)* doesn't exist in any of the tool versions (at least I wasn't able to find it).

shiltemann:
If you use the tutorial mode (hat icon) in Galaxy, and click on the tool name, it opens the correct sort tool: 
![image](https://user-images.githubusercontent.com/2563865/193850315-87a49438-c76b-4cc5-8910-d81954f81b81.png)

PR #3673, opened 2022-10-19
[FAQ] new FAQ to address potential mixup of tools with similar names

shiltemann:
There are often mixups between tools for which many different tools with similar names exist (sort, cut, etc), and this FAQ recommends using tutorial more or checking of full tool name matches.

fixes https://github.com/galaxyproject/training-material/issues/3636 as reported by @gallardoalba 

image


shiltemann:
hmm, in this case two different tools have the exact same tool name (on EU), so point 2 still doesn't help, but in general it should

image

hexylena (approved):

gallardoalba:
Great :D thanks!

Issue #3288, opened 2022-03-24
Broken links in tutorial "Visualizations: generic plugins"


beatrizserrano:
While cleaning the Hub, I deleted outdated info that is pointed from different locations, one of those is the tutorial [Visualizations: generic plugins](https://training.galaxyproject.org/archive/2019-06-01/topics/dev/tutorials/visualization-generic/tutorial.html). Since the content is outdated and the links are just listed as "Additional documentation", I wonder if the way to go is simply to remove them (see [this issue](https://github.com/galaxyproject/galaxy-hub/issues/1252) to get more context).

PR #3678, opened 2022-10-19
[Dev] Generic Viz Plugins: remove broken links

shiltemann:
fixes https://github.com/galaxyproject/training-material/issues/3288

@beatrizserrano I removed the links that are not working. I left two that still looked ok. Do you think these are still helpful of should I just remove them too?

### DIFF
--- a/topics/dev/tutorials/visualization-generic/tutorial.md
+++ b/topics/dev/tutorials/visualization-generic/tutorial.md
@@ -47,12 +47,8 @@ However, for this tutorial we will keep it basic.
 
 Additional documentation about Galaxy visualizations can be found here:
 
-- [VisualizationsRegistry](https://galaxyproject.org/visualizations-registry)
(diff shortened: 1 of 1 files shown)


beatrizserrano:
> @beatrizserrano I removed the links that are not working. I left two that still looked ok. Do you think these are still helpful of should I just remove them too?

Thanks @shiltemann! I'm not sure how helpful they are nowadays, @jennaj has probably a better sense of it. I've pulled some numbers:
- [DataProviders](https://galaxyproject.org/data-providers) had 650+ pageviews in 2021, and similar number in 2022.
- [DataProviders/Cookbook](https://galaxyproject.org/data-providers/cookbook) had 25+ pageviews in 2021, and 10 in 2022. 

So I'm hesitant to remove them, we could keep them for now?

shiltemann:
thanks, sounds good, let's keep them :+1:

beatrizserrano (approved):

PR #3698, opened 2022-10-25
update coline's name

Marie59:


shiltemann:
Ah, looks like @yvanlebras made that same change in his branch half an hour ago too. 

Though it looks like his branch still has a mismatch in capitalisation for your name (between the tutorial and how it is listed in CONTRIBUTORS.yaml. it doesn't matter which you use, as long as it is consistently the same everywhere)

Thanks @Marie59! While you are there tho, you might want to update your own entry in CONTRIBUTORS.yaml? It only says "Marie" now

P.S. @yvanlebras: if you work on a branch in your own fork, you can give her and others the rights to write directly to your fork, so that they don't need to make PRs here and wait for us to review (we will then review only when you are ready to merge the branch into GTN main), just in case that is easier for you, but up to you of course

PR #3701, opened 2022-10-26
Build time improvements

hexylena:
Assigns are apparently expensive (ok that's a 324kb file, makes sense, don't want to allocate that for every page), especially if you're going to push data directly to a function/filter/etc.

```
{% assign fdbk = site.data['feebdack'] %}
```

1. We started at 88s to render JUST the extras.html partial. 
1. I replaced it initially with {% assign does_tuto_have_feedback %} IN the if/else case around the layout. That took it down to 18 seconds or so
2. Then replaced the where/group bys with a ruby function, down to 8s
3. and finally removed the feedback assignment as a variable, to get it down to **1s** :tada: 

Going to see if there's anything else I can improve but I suspect not.


hexylena:
The replacement of the modification time plugin further reduces from 65.252 seconds to 48.978

shiltemann (approved):

shiltemann:
wow!

PR #3702, opened 2022-10-26
[Slides] Also render author line for new "contributions:" syntax

shiltemann:
The author line on slides was not showing when the slide deck uses the fancier `contributions:` syntax

ping @hexylena


hexylena (approved):
Oh wow

PR #3723, opened 2022-10-28
Add tags

yvanlebras:


shiltemann:
@yvanlebras thanks for all the PR adding tags! (but feel free to edit multiple tutorials in a single PR in the future too)

Ecology topic is growing nicely, you may want to consider subtopics at some point? For example like the proteomics topic, to group tutorials on the topic page: https://training.galaxyproject.org/training-material/topics/proteomics/

shiltemann (approved):

PR #3730, opened 2022-10-31
Update Data Manipulation Olympics for Compute 2.0

natefoo:
See all the relevant discussion in galaxyproject/tools-iuc#4674.


wm75:
@shiltemann unfortunately, this update makes your entire last exercise, which is very instructive otherwise, superfluous since the new tool version makes everything possible in a single step (now part of the first BMI calculation exercise).
I removed this last part, but maybe you can come up with something similarly challenging?

shiltemann:
Nice! Thanks @natefoo and @wm75!

I'll think of a new "advanced" exercise to replace the BMI one, but I only added it in because I ran into the limitation and needed a bunch of steps to work around it, so it's definitely a good thing we have to remove it!

shiltemann (approved):

PR #3736, opened 2022-11-03
Automatically calculate image dimensions to decrease page shifting

hexylena:
Here we update the jekyll-figurify to use pre-calculated image dimensions. On first use the image's dimensions will be calculated and cached.

Note that this does not apply to *every* image, instead only to those used in markdown syntax that go through figurify. This is probably acceptable.


hexylena:
Not sure why i left this in draft, looks/works fine.

shiltemann (approved):

PR #3763, opened 2022-11-22
video.yml write permission fix part 2: Add recursive flag

fubar2:
Oh. Sorry @hexylena Hold the beer.
I forgot to add the recurse option - that works on my fork in an action that makes a single target as a sample.
Let's see how this goes?

### DIFF
--- a/.github/workflows/video.yml
+++ b/.github/workflows/video.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Build Videos
         run: |
           # Override permissions on _site to allow docker access.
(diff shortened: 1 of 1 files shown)


hexylena:
Thanks again for having bandwidth to take a stab at this.

fubar2:
Last run seems to have worked so I'd say it's fixed.

hexylena:
Fan freakin tastic @fubar2. I'm very happy and simultaneously very frustrated I was like 3 characters short of it working. :laughing:

PR #3769, opened 2022-11-24
Refine new AIV tutorial based on first training experience

wm75:


wm75:
Thanks @kciy for spotting the copy/paste error at the fastp step before the training!
Will fix more issues here as they get discovered.

shiltemann (approved):

wm75:
Thanks @shiltemann for the reminder. I had this last point on my list of fixes: bamtools split by ref messes up the order of the resulting collection if you run it with an explicit selection of refs as we discovered during the workshop.

I'm currently investigating how to fix this best at the tool level, but until then this is a simple fix for the tutorial.

wm75:
Guess these are all the hotfixes for now.

shiltemann (approved):

shiltemann:
wait, did you want to remove the `enable: false` here too?

wm75:
Lets leave it disabled for a little while still. I've also learnt quite a bit about typically asked questions in the field and will rephrase a couple of things in a separate PR.

PR #3784, opened 2022-12-14
WIP: Galaxy project guide for participants

fubar2:
Tried to explain Galaxy to anyone recently?

It's getting really hard to do, because the project has grown to a point where it is a full time job just to keep up to date with everything that's going on, and there is very little guidance on how new activities can be started.  Apologies for the very opinionated content, but this had to start somewhere. Clearly, no single person's view is enough. This is a big and tricky topic, but it should not be a rewrite of the Hub! Idea is to provide a more simple and logical structure that has pointers to the hub, where all the details live. 

It's far from complete - a lot of stub headings as placeholders, and the current structure seems to work ok for me - but suggested improvements would be welcomed. 

This proposed training topic is designed for anyone who wants a big picture overview of the project. In particular, it tries to show how to find interesting activities to contribute to. There is a module of "stories" about how things get started in the project to provide strategies for newcomers with good ideas - Tim has contributed the example and it could be useful if enough other examples can be contributed from other communities and individuals?


fubar2:
Not sure why check is failing so will close and recreate the fork.

PR #3788, opened 2022-12-19
Visual mode instructions in RMarkdown snippet

wee-snufkin:
It is way easier for the learner to read the rendered text instead of plain markdown; and I felt like it would be worth adding a short description of how to do it.


hexylena (approved):
oh, fantastic!

PR #3791, opened 2022-12-23
Creating cell cycle regression tutorial

MarisaJL:
Here is the completed and tested version of a tutorial to regress out the effects of the cell cycle from scRNA-seq data.


bgruening:
What a great present at the end of the year! Thanks a lot @MarisaJL for your contribution!

topics/single-cell/tutorials/scrna-case_cell-cycle/tutorial.md
```diff
@@ -0,0 +1,457 @@
...
+
+>    > <comment-title>Comment: When should we regress out the effects of the cell cycle?</comment-title>
+
+>    >  Cell cycle regression can be particularly important if we are planning to do trajectory analysis down the line or if we have a dataset that is very strongly influenced by the cell cycle {% cite Luecken2019 %}. However, it isn't always appropriate to remove the effects of the cell cycle - sometimes they can be useful for distinguishing between dividing and non-dividing cell types. When you are analysing your own data, you might need to try it both ways to determine if the effects of the cell cycle are helpful or not. You could also check whether the cell cycle genes are among the top scoring genes expressed by your cell clusters to get an idea of how strong the effects are. 
+>    {: .comment}
```
hexylena, line 120:
```suggestion
> <comment-title>Comment: When should we regress out the effects of the cell cycle?</comment-title>
>  Cell cycle regression can be particularly important if we are planning to do trajectory analysis down the line or if we have a dataset that is very strongly influenced by the cell cycle {% cite Luecken2019 %}. However, it isn't always appropriate to remove the effects of the cell cycle - sometimes they can be useful for distinguishing between dividing and non-dividing cell types. When you are analysing your own data, you might need to try it both ways to determine if the effects of the cell cycle are helpful or not. You could also check whether the cell cycle genes are among the top scoring genes expressed by your cell clusters to get an idea of how strong the effects are. 
{: .comment}
```

topics/single-cell/tutorials/scrna-case_cell-cycle/tutorial.md
```diff
@@ -0,0 +1,457 @@
...
+>    - {% icon param-file %} *"Input object in AnnData/Loom format"*: `CellCycle_Annotated` (output of **Inspect and manipulate** {% icon tool %})
+>    - *"Variables to regress out"*: `phase`
+>
+> 2. Rename the output `CellCycle_Regressed` 
+> 
+{: .hands_on}
+
+# Plotting the Effects of Cell Cycle Regression
+
+Your data is now ready for further analysis, so you can return to the [Filter, Plot and Explore Single-cell RNA-seq Data]({% topics/single-cell/tutorials/scrna-case_basic-pipeline/tutorial.md %}) tutorial and move on to the Preparing coordinates step there. Make sure that you use the `CellCycle_Regressed` dataset (you may want to rename it as `Use_me_Scaled` as that is the name used in the main tutorial). However, if you want to understand how cell cycle regression has affected your data then you can work through the following steps first.
```
hexylena, line 173:
```suggestion
Your data is now ready for further analysis, so you can return to the [Filter, Plot and Explore Single-cell RNA-seq Data]({% link topics/single-cell/tutorials/scrna-case_basic-pipeline/tutorial.md %}) tutorial and move on to the Preparing coordinates step there. Make sure that you use the `CellCycle_Regressed` dataset (you may want to rename it as `Use_me_Scaled` as that is the name used in the main tutorial). However, if you want to understand how cell cycle regression has affected your data then you can work through the following steps first. 
```

@nomadscientist informed me of this specific issue, so I added linting explicitly for this that will be available in the future https://github.com/galaxyproject/training-material/pull/3829/commits/09e74ca20d252df742dc8fe3960b76ce877570ba

MarisaJL:
I'm having an error about a link to the workflow file saying it doesn't exist, but isn't that because it hasn't been created yet? It's part of this pull request so I'm not sure how to deal with that!

hexylena:
It is because it is named `workflows/index` and not `workflows/index.md`, please rename the file and that *should* fix the error.

hexylena (approved):
Looks great, thanks so much @MarisaJL :)

Issue #3685, opened 2022-10-21
Self-host css/js/etc


hexylena:
Given that CDNs are now useless for efficiency reasons (thanks browser!), and @nomadscientist's request for running in air-gapped environments, let's bring in the css/js into the repo so it can run fully offline.

PR #3799, opened 2023-01-03
Accessibility fixes, another round

hexylena:
- Passes a few more WCAG things we missed earlier
- Passes more of the axe suite
- (nearly) passes html validation (which is v picky and doesn't reflect what browsers happily accept.)
- fixes https://github.com/galaxyproject/training-material/issues/3685

(just opening this old branch I worked on before the holidays and never finished.)


shiltemann (dismissed):

hexylena:
(just noting for us, this is/was WIP work that isn't finished, in addition to needing the conflicts resolved.)

hexylena:
Adds a new test: `axe` is run on a subset of pages, alongside `html-proofer`

hexylena:
I think this is ready to go @shiltemann if you have some time sometime to check that nothing looks very broken. it's mostly css / html structure changes, it should generally be invisible (except links now get underlined.)

shiltemann (dismissed):

PR #3804, opened 2023-01-05
Workaround for temporarily broken hub feed.json

hexylena:
If we don't find a solution by friday for https://github.com/galaxyproject/galaxy-hub/issues/1793, since @pvanheus's course starts on monday we can merge this friday evening.


dannon:
Hub's feed.json should be fixed now.  Thanks for noticing this and reporting it.

PR #3813, opened 2023-01-12
Work around jekyll build time issue with custom link

hexylena:
This does not support news, intentionally, because it's too much bother to deal with for something temporary, AND we only link to a news post in a single place and it wasn't really all that important of a link.


hexylena:
maybe we'll just merge the other PR and not worry about having it separate here.

PR #3830, opened 2023-01-20
Duplicate time key, I picked one randomly

hexylena:
@nomadscientist which is right? there were two time estimates. If not correct, please feel free to edit + merge it :)


nomadscientist:
1 hour is right!

nomadscientist (approved):

PR #3851, opened 2023-01-27
Update sequencing_nanopore.md

hrhotz:
adding blank spaces between "inelectrical", "Whennucleotides" etc.


hexylena:
ugh that's a bigger issue, ok. I think I've got a more general fix for this.

hexylena (approved):
as far as I can tell my fix doesn't break anything but I guess we'll find out :sweat:

hrhotz:
@hexylena Thank you for the right fix (instead of my quick and dirty one)

PR #3852, opened 2023-01-27
First Tutorial Update

EngyNasr:
that's the first update to the training, added citations in intro, correcting results in phylogenetic trees, quality retaining not improving, fastp instead of multiqc figure used

Please use this PR for all suggestions

Thanks a lot,
with love,
Engy


wm75:
@EngyNasr I've carried over most of this PR to #3853.
I dropped the fastp image and additional mention because it's not worth it without eliminating fastqc from the WF.
Also, I haven't added your two new links in the introduction yet - so once #3853 is merged you can open a new PR and continue there.

PR #3855, opened 2023-01-29
Icons in Automatic Jupyter Notebooks

wee-snufkin:
When icons are included in the text of the tutorial, there is an error with generated Jupyter Notebook which says: "The editor could not be opened due to an unexpected error: Unexpected token f in JSON at position XXXX".  I tested out that it's due to the icon symbols, so just wanted to point that out and save people's time on debugging :)


hexylena (dismissed):
Oof, yeah, did not consider people might include those. Good catch.

If we had a set of png copies of the icons in the gtn we could probably make a workaround but, short term, your solution works, letś just avoid them for now.

hexylena (approved):

hexylena:
Thanks @wee-snufkin !

PR #3866, opened 2023-02-01
1st of Feb 2023 Training updates for Foodborne pathogen detection

EngyNasr:
updating Zenodo version with different file names and the tutorial accordingly and updating Biolytix image


topics/metagenomics/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md
```diff
@@ -9,7 +9,7 @@ tags:
     - Heatmap
     - cyoa
 level: Introductory
-zenodo_link: "https://zenodo.org/record/7567733"
+zenodo_link: "https://zenodo.org/record/7593928"
```
wm75, line 12: I think you can simplify things by using [10.5281/zenodo.7564789](https://doi.org/10.5281/zenodo.7564789), i.e., the stable DOI that will always resolve to the latest version.

@hexylena is that also a good solution for the data-library.yaml file, or a bad idea because it could make testing non-reproducible?

wm75: Ah, strike this whole comment because the automatic resolution doesn't work for links to files. Stick to what you're proposing @EngyNasr :-)

PR #3883, opened 2023-02-09
ga4gh/trs/Write files ending with json, rewrite in nginx

hexylena:
My ugly hack to name the files `1?gtn=true` didn't work, sadly, I guess github strips query params off rather than saying "ah yes a file with exactly that name exists" like many other static file servers will do, allowing you to work around file + folder with the same name.

So we'll just end it in `.json` properly, and then rewrite in nginx. https://github.com/galaxyproject/infrastructure-playbook/pull/39


hexylena:
@bgruening could you hit merge for me?

hexylena:
oh nevermind I missed the conflict. need to fix first

natefoo (approved):

PR #3896, opened 2023-02-16
Remove notranslate

hexylena:
Based on analysis by @jurgigi, who discovered that for whatever reason the `<span>` tags around the notranslate words are causing those words to be merged together with surrounding words, as if the space wasn't there. The end result being they get translated as a block and result in extremely poor translations.

Between:
- This analysis
- The fact we haven't updated the list in some years
- And the improvements in the intervening years in machine translation

I think we may be better off dropping this.  I suspect e.g. workflow, downstream , mapping, reads are all getting automatically translated a bit better these days. BED probably not but as long as it is capitalised it should be fine.

cc @nomadscientist


shiltemann (approved):

nomadscientist:
I agree - I take the positive of the pilot showing that human translation just isn't an option for the sheer amount of hours required (let alone upkeep), BUT it gave us what we needed to get Jurgi to see about making stuff better more widely. 
Like this! Who knew?

PR #3913, opened 2023-02-27
Update tutorial.md - suggested rewording to match figure 5

hrukkudyr:
Figure 5 in this tutorial shows two genes each with two transcripts. gene 1 has UMI coloured pink and blue. Gene 2 has pink and green. The text, though, describes Gene 2 as being brown and green. This may be confusing to some students. The commit here changes the tutorial text so that Gene two is described as pink and green. The alternative would be to edit Figure 5?


mtekman:
@hrukkudyr Yep, I think at some point I wanted to change the colours to make them more red-green friendly (somehow?), and then I reverted back.

Good catch, and the PR looks good -- thanks for the feedback

mtekman (approved):

PR #3919, opened 2023-03-06
VGP assembly tutorials: add glossary, add abbreviations to wkflw training, CYOA for contigging

abueg:
hello! 👋🏼 

adding a glossary of commonly used terms to the beginning of the VGP assembly tutorial, I think it would be a useful resource to be able to refer to and for readers to have while going through the tutorial

PR is draft for now while i finish the glossary and also update some of the tutorial after recently re-running it


topics/assembly/tutorials/vgp_genome_assembly/tutorial.md
```diff
@@ -53,6 +55,49 @@ The G10K consortium launched the Vertebrate Genomes Project (VGP), whose goal is
...
+
+**Pseudohaplotype assembly**: An assembly that consists of long phased haplotype blocks that are separated by blocks where the haplotype cannot be distinguished (often homozygous regions). This can result in "switch errors", where a region of DNA might have a block of maternal sequence, then sequence the same in both haplotypes, and then a block of paternal sequence. These types of assemblies are usually represented by a primary assembly and an alternate assembly. (This definition largely taken from the [NCBI's Genome Assembly Model](https://www.ncbi.nlm.nih.gov/assembly/model/#asmb_def).)
+
+**Primary assembly**:
+
+**Alternate assembly**: The alternate assembly consists of sequences that are an alternate representation of loci in the primary assembly (such as at heterozygous loci). Usually this means the alternate haplotypic representation at a given region. These types of sequences are also referred to as haplotigs.
```
hexylena, line 66: not sure if it's relevant, since they're not acronyms, but we have something like this built in as a feature. https://training.galaxyproject.org/training-material/topics/contributing/tutorials/create-new-tutorial-content/tutorial.html#abbreviations, if it's useful in more than just this tutorial, we could expand the feature to also do glossary entries that aren't automatically abbreviated. E.g. https://training.galaxyproject.org/training-material/topics/admin/tutorials/tiaas/tutorial.html#frequently-asked-questions when TIaaS is used in the text sometimes it gets the "abbr" tag which lets us have an explanation on hover to help remind the reader of what the term means.

abueg: ‼️ ohh, that sounds like it could be a good idea! if i end up using it i'll probably have to make an abbreviated definition for the terms, though, as they're pretty lengthy right now. maybe a length definition up-front as a primer, and then an abridged glossary at the end? i'll keep that in mind though, thank you for the rec!

abueg: hi @hexylena , i have a quick question to clarify how the abbreviations tag works -- so, when the `{abbreviation}` is first used, it will render as the expanded abbreviation + the abbreviation, and then subsequent uses will be rendered as just the abbreviation + hover-text? just want to make sure ^^

also, checking out the rest of the options on the page and the choose your own adventure option actually might be a perfect fit for our pipeline that branches and then re-merges depending on data availability, thank you again for pointing me to that! 🙏🏼

abueg:
hi @gf777 i added a glossary of terms to the VGP longform tutorial, i think it would be a useful reference -- would you mind looking over my definitions to double-check if they are accurate/clear please? just github comment any edits you want and i'll incorporate them, thanks!

(can check the `Files changed` tab to comment on the added lines)

topics/assembly/tutorials/vgp_genome_assembly/tutorial.md
hexylena: > so, when the {abbreviation} is first used, it will render as the expanded abbreviation + the abbreviation, and then subsequent uses will be rendered as just the abbreviation + hover-text? just want to make sure ^^

yes, but every time you want the hover text definition, you must use `{abbreviation}`, it is not automatic (I discovered too many bugs that way.)

But more importantly, **if** you think this sort of thing is generally useful (more than just 1 tutorial?), we can also have a "glossary" somehow that does not do the automatic abbreviation. But the same feature of "tag the words you would like students to be able to look up more easily."

> might be a perfect fit for our pipeline

amazing ! very happy to hear this :)

abueg: hmm 🤔 for VGP in specific, we currently have two tutorials, and the one that is not this one is workflow-focused so the glossary might actually be more useful there, since that tutorial page won't have the in-depth concepts!

but if you meant if i think the feature is useful in general, then yes i would definitely think so! it feels to me like it would help by being somewhat like flashcard/repetition, where readers can mouse over and just re-solidify the definition in context multiple times over the training 🤔

abueg:
ack did i break something elsewhere in the code 😭

abueg:
ok there is still a broken link getting picked up during linting, but i'm not sure how to fix that because i didn't edit the `feedback.html` file as far as i know :(

other than that though, the text is ready to review! 🙏🏼

hexylena:
> ok there is still a broken link getting picked up during linting, but i'm not sure how to fix that because i didn't edit the feedback.html file as far as i know :(

no, not your fault. We can merge over this.

Delphine-L (approved):
I corrected a few typos and replace an image, if that works for you then the rest is good for me, awesome work ! :)

abueg:
looks good to me, thank you so much @Delphine-L !! <3

gallardoalba (approved):

PR #3927, opened 2023-03-09
imaging/DL/ai4life

hexylena:
@beatrizserrano per discussion today, here is an example topic using y'all's branding, we can start adding materials and resources next :) 

It's set to draft mode so it won't appear on the home page, but otherwise contributors could start contributing.

When merged it'll be https://training.galaxyproject.org/training-material/topics/ai4life/

Topic | Slides
--- | ---
image | image


hexylena:
Ok @beatrizserrano https://training.galaxyproject.org/training-material/topics/ai4life/ it's live

beatrizserrano:
Very cool and super fast!! Thank you very much @hexylena ❤️

Issue #3939, opened 2023-03-14
Separate out db servers group


hexylena:
Couple more things:

- Maybe should consider creating a `dbservers` group and putting postgres stuff in `group_vars/dbservers.yml`. It might be confusing up front but it will help people who later deploy on their own infra and have a separate DB server.
- Should we create `galaxy_user_name` and `galaxy_db_name` vars to avoid duplication in `postrgresql_objects_*` and `database_connection`?

_Originally posted by @natefoo in https://github.com/galaxyproject/training-material/issues/2583#issuecomment-863747333_

PR #3942, opened 2023-03-14
Updates to *Installing Galaxy with Ansible*

natefoo:
Still WIP but I wanted to get it out there for comment. This tutorial is the foundation for basically all the other admin tutorials and some changes could be controversial, such as:

- The job conf is now inline instead of in a separate file
- Diff mode by default means lots of verbosity (but also should make mistakes after the first run more apparent)

I probably broke git-gat stitching please don't hate me @hexylena.

Depends on as-yet unmerged changes to the role in https://github.com/galaxyproject/ansible-galaxy/tree/gravity-systemd. My plan is also to update this for the monolithic role so there is a 23.0 version using the stuff that everyone already knows and loves, and then I'll update it to use [the collection](https://github.com/galaxyproject/ansible-collection-galaxy), which I think will be the best way forward (removing all the remote_user/become/become_user stuff entirely).

TODO: #3939

### DIFF
--- a/topics/admin/tutorials/ansible-galaxy/tutorial.md
+++ b/topics/admin/tutorials/ansible-galaxy/tutorial.md
@@ -346,7 +346,7 @@ We have codified all of the dependencies you will need into a YAML file that `an
 >    ```diff
 >    --- /dev/null
 >    +++ b/ansible.cfg
->    @@ -0,0 +1,7 @@
+>    @@ -0,0 +1,11 @@
 >    +[defaults]
 >    +interpreter_python = /usr/bin/python3
 >    +inventory = hosts
@@ -354,6 +354,10 @@ We have codified all of the dependencies you will need into a YAML file that `an
 >    +stdout_callback = yaml
 >    +# Use the stdout_callback when running ad-hoc commands.
 >    +bin_ansible_callbacks = True
+>    +
+>    +# Show diffs of changes
+>    +[diff]
+>    +always = True
 >    {% endraw %}
 >    ```
 >    {: data-commit="Add ansible.cfg"}
@@ -760,7 +764,7 @@ For a normal Galaxy instance there are a few configuration changes you make very
 - Configuring the admin user list
 - Changing the "brand"
 
-Additionally we'll go ahead and set up the production-ready [Gunicorn + Webless](https://docs.galaxyproject.org/en/master/admin/scaling.html#deployment-options) deployment, which separates Galaxy's web and job handling into separate processes.
+Additionally we'll go ahead and set up the production-ready [Gunicorn + Webless](https://docs.galaxyproject.org/en/master/admin/scaling.html#deployment-options) deployment, which separates Galaxy's web and job handling into separate processes. This is done by configuring Galaxy's process manager, [Gravity](https://gravity.readthedocs.io).
 
 Finally, best admin practices are to not run Galaxy as a user with `sudo` access, like your login user probably has. Additionally, it is best to install the Galaxy code and configs as a separate user, for security purposes. So we will instruct the `galaxyproject.galaxy` role to create a new user account specifically to run Galaxy under.
 
@@ -873,7 +877,7 @@ The configuration is quite simple thanks to the many sensible defaults that are 
 >    ```diff
 >    --- a/group_vars/galaxyservers.yml
 >    +++ b/group_vars/galaxyservers.yml
->    @@ -26,3 +26,15 @@ galaxy_force_checkout: true
+>    @@ -26,3 +26,12 @@ galaxy_force_checkout: true
 >     miniconda_prefix: "{{ galaxy_tool_dependency_dir }}/_conda"
 >     miniconda_version: 4.12.0
 >     miniconda_channels: ['conda-forge', 'defaults']
@@ -886,9 +890,6 @@ The configuration is quite simple thanks to the many sensible defaults that are 
 >    +    file_path: /data
 >    +    tool_data_path: "{{ galaxy_mutable_data_dir }}/tool-data"
 >    +    id_secret: "{{ vault_id_secret }}"
->    +    job_config:
->    +      handling:
->    +        assign: ['db-skip-locked']
 >    {% endraw %}
 >    ```
 >    {: data-commit="Configure galaxy config"}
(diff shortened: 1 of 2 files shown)


topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -763,24 +764,18 @@ For a normal Galaxy instance there are a few configuration changes you make very
...
 The configuration is quite simple thanks to the many sensible defaults that are provided in the Ansible roles.
 
 > <hands-on-title>Minimal Galaxy Playbook</hands-on-title>
 >
 > 1. Open `galaxy.yml` with your text editor and set the following:
 >
->    - Amend the [package installation](https://docs.ansible.com/ansible/2.9/modules/package_module.html#package-module) pre-task to install some additional necessary dependencies: `bzip2`, `git`, `make`, `tar`, and `virtualenv`.
->    - Add the roles `geerlingguy.pip`, `galaxyproject.galaxy` and `uchida.miniconda` (in this order) at the end, with `uchida.miniconda` run as the `galaxy` user.
+>    - Amend the [package installation](https://docs.ansible.com/ansible/2.9/modules/package_module.html#package-module) pre-task to install some additional necessary dependencies: `bzip2`, `git`, `make`, `tar`, `python3-venv`, and `python3-setuptools`.
+>    - Add the roles `geerlingguy.pip`, `galaxyproject.galaxy` and `galaxyproject.miniconda` (in this order) at the end, with `uchida.miniconda` run as the `galaxy` user.
```
lldelisle, line 778:
```suggestion
>    - Add the roles `geerlingguy.pip`, `galaxyproject.galaxy` and `galaxyproject.miniconda` (in this order) at the end, with `galaxyproject.miniconda` run as the `galaxy` user.
```

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -441,6 +438,10 @@ For this tutorial, we will use the default "peer" authentication, so we need to
 >    ```
 >    {: data-commit="Add initial group variables file"}
 >
+>    > <comment-title>Regarding `expanduser`</comment-title>
+>    > The `expanduser` filter turns `~postgres` into the absolute path to the `postgres` user's home directory. However, you should be warned that the `expanduser` filter, like all Ansible filters, runs *locally*. This works in the training because Ansible is executed from the Galaxy server, but if you run Ansible over SSH, you would need to set the full path in `postgresql_backup_local_dir` directly.
```
hexylena, line 442: this has bitten me before. I don't know if I like it. Perhaps we should just explicitly set to the full path? I expect 1 out of 20 students remembers this comment when they follow this training and later experience it.

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -881,29 +870,25 @@ The configuration is quite simple thanks to the many sensible defaults that are
...
 >    +
 >    +galaxy_config:
 >    +  galaxy:
 >    +    brand: "🧬🔬🚀"
 >    +    admin_users: admin@example.org
 >    +    database_connection: "postgresql:///galaxy?host=/var/run/postgresql"
 >    +    file_path: /data
->    +    check_migrate_tools: false
 >    +    tool_data_path: "{{ galaxy_mutable_data_dir }}/tool-data"
->    +    object_store_store_by: uuid
```
hexylena, line 906: is this default now or so?

hexylena:
> the job conf is now inline instead of in a separate file

yeah, that's probably fine for most folks ig.

> Diff mode by default means lots of verbosity (but also should make mistakes after the first run more apparent)

+1

> I probably broke git-gat stitching please don't hate me @hexylena.

s'fine, i'll fix it at the end.

hexylena:
> > I probably broke git-gat stitching please don't hate me @hexylena.
> s'fine, i'll fix it at the end.

actually better idea @natefoo , you can try the tutorial I wrote on re-stitching https://training.galaxyproject.org/training-material/topics/contributing/tutorials/admin-knitting/tutorial.html and let me know if it explains it well enough or if I need to update it. Just to reduce bus factor and all.

lldelisle:
I proposed a PR: https://github.com/natefoo/training-material/pull/1 for the tmp_dir param.

topics/admin/tutorials/ansible-galaxy/tutorial.md
natefoo: Ok, I agree with that (we're already reaching the point of oversaturation of tips/comments in this tutorial anyway).

Kinda silly that Ansible still has no built in mechanism for this. You have to write a series of tasks that run the getent module and extract the attributes you want to facts. So here we can just hardcode the path.

topics/admin/tutorials/ansible-galaxy/tutorial.md
natefoo: Acutally - I thought I verified that it was, but now I'm not sure:

>          What Dataset attribute is used to reference files in an ObjectStore implementation,
>          this can be 'uuid' or 'id'. The default will depend on how the object store is configured,
>          starting with 20.05 Galaxy will try to default to 'uuid' if it can be sure this
>          is a new Galaxy instance - but the default will be 'id' in many cases. In particular,
>          if the name of the directory set in <file_path> is `objects`, the default will be set
>          to 'uuid', otherwise it will be 'id'.

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -1706,108 +1756,62 @@ The configuration is quite simple thanks to the many sensible defaults that are
...
->    > {: data-cmd="true"}
->    {: .code-in}
->
-> 3. Log in and check the status with `sudo systemctl status galaxy`
+> 1. Log in and check the status with `sudo galaxyctl status`
 >
 >    > <code-in-title>Bash</code-in-title>
 >    > ```bash
->    > sudo systemctl status galaxy
+>    > sudo galaxyctl status
```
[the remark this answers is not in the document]
natefoo: I went back and forth over this, the Gravity syntax is a bit shorter, the equivalent systemd syntax would be `systemctl status 'galaxy*'`, likewise instead of `galaxyctl status gunicorn` you have `systemctl status galaxy-gunicorn`. But I'm totally fine with pushing everyone to just interact with systemd/journald natively.

In the changes to the role (unpublished) I drop a wrapper around `galaxyctl` into `/usr/local/bin` that sets `$GRAVITY_CONFIG_FILE` appropriately so that you can call it from anywhere. Seemed like a useful convenience.

natefoo:
> actually better idea @natefoo , you can try the tutorial I wrote on re-stitching https://training.galaxyproject.org/training-material/topics/contributing/tutorials/admin-knitting/tutorial.html and let me know if it explains it well enough or if I need to update it. Just to reduce bus factor and all.

Yep, I've used it before, works great, I can do this once I'm finished with the updates.

lldelisle:
Sorry, I can't believe that I misspell my username...

topics/admin/tutorials/ansible-galaxy/tutorial.md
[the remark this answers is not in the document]
hexylena: Both? Both is good. Show galaxyctl first, and then show the systemd units that get created? I think it's useful to show those as well for e.g. things like memory limits? And just to show the connection a bit? (We could just look at one, maybe, instead of all)

hexylena:
I am told this is ready, so I open it for review.

hexylena (approved):

PR #3960, opened 2023-03-22
Update email address

abretaud:
Forgot about that old address...

Thinking about it, there's no mastodon support yet?


hexylena:
Ahh you're right! We should add that :D

hexylena (approved):

PR #3962, opened 2023-03-23
Add fediverse support since we have twitter

hexylena:
For @abretaud. (Please feel free to merge it :) )


hexylena:
Of course! I would join you but I still can't decide how "work-friendly" mine :laughing: and whether or not I should 'advertise' it like this

PR #3968, opened 2023-03-27
Remove any mention of "For all dataset selected..." in FAQ and different tutorials

lldelisle:
With the new galaxy interface 23.0, when multiple datasets are selected the button is not "For all datasets selected..."  but "n of N selected" or "All N selected". So I changed it wherever I could find it.


lldelisle:
Just a question/remark should we keep `Operations on multiple datasets` while the info box say `Select Items`?

hexylena:
> Just a question/remark should we keep Operations on multiple datasets while the info box say Select Items?

no we should attempt to match the titles where possible

hexylena (approved):

hexylena (approved):

PR #3972, opened 2023-03-29
Change to import history text

hrukkudyr:
I'm currently proofing this tutorial for Smorgasbord 2023 and experiencing difficulties in getting the dataset into galaxy.org.
- Zenodo transfers are failing on the transfer files
- importing a history from usegalaxy.eu is complicated 
Workaround: propose additional link to a newly created history on galaxy.org This holds the data from @nomadscientist usegalaxy.eu history including the #tags


hexylena (approved):
Seems fine to me, but yea zenodo working would be ideal here.

PR #3975, opened 2023-03-30
Add draft beacon tutorial

hexylena:
untested, wip


hexylena:
From @mira-miracoli 

> I see a problem with postgres maybe, the [role](https://github.com/Paprikant/ansible-role-beacon) wants to run postgres in docker and uses port 5432 as default. The [importer-role](https://github.com/Paprikant/ansible-role-beacon_importer) seems to assume that the db is not on the same node. the [importer-role](https://github.com/Paprikant/ansible-role-beacon_importer/blob/d4535923401596c2f600b303eb4586a4bbcfe043/tasks/main.yml#L10) which runs the import script on cron, wants to install Python 3.9, but that should be fine, I guess?

hexylena:
> run postgres in docker and uses port 5432 as default

Looks like we can change that, glad it's dockerized: `postgres_external_binding: 5432` / `beacon_db_port: 5432`

> seems to assume that the db is not on the same node

their example in the readme uses `localhost` so maybe it's fine? I'll test this in one of the VMs for sure.

> wants to install Python 3.9, but that should be fine, I guess?

that is  not great, but, ok, we can work with that. at least ubuntu is sensible about coexisting python installations, as long as you're really careful to use the associated pip.

hexylena:
And, oof, the importer exposes the password on the CLI, nice /s

PR #3985, opened 2023-04-10
More pulsar tutorial updates for GAT2023

cat-bro:
Add consumer_timeout to rabbitmq_config and make sure rabbitmq role is run after nginx. There are still problems as the container continually restarts and the container log complains about not being able to read `/etc/ssl/user/privkey-rabbitmq.pem` which is owned by root with 640 permissions. After changing permissions on that file to 644, the docker container is able to stay up, but I imagine the proper solution is for the file to be owned by somebody else (rabbitmq?)

TODO: job conf diffs will be different as this takes place after the job destinations tutorial.

### DIFF
--- a/topics/admin/tutorials/pulsar/tutorial.md
+++ b/topics/admin/tutorials/pulsar/tutorial.md
@@ -145,11 +145,11 @@ Firstly we will add and configure another *role* to our Galaxy playbook - a comm
 >       version: 0.0.2
 >     - src: galaxyproject.slurm
 >       version: 0.1.3
->    +- name: geerlingguy.docker
+>    +- src: geerlingguy.docker
 >    +  version: 6.1.0
->    +- name: usegalaxy_eu.rabbitmqserver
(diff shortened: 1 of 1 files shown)


hexylena:
> After changing permissions on that file to 644, the docker container is able to stay up, but I imagine the proper solution is for the file to be owned by somebody else (rabbitmq?)

that key is supposed to be owned by rabbitmq, but I guess we need it owned with the permisisons of the user inside the container.

cat-bro:
> TODO: job conf diffs will be different as this takes place after the job destinations tutorial.

>> Ok, this looks fine. We can resolve the diffs at a later stage.

By the end of the tpv tutorial, the default destination/environment is `tpv_dispatcher` and all other destinations have been moved into the tpv config file. I'll need to rewrite this a bit so that the pulsar destination also goes into tpv config instead of galaxy_job_config.

hexylena:
Sorry for merging this sooner, but, I need to resolve this N pr mess we have currently and get you a better base to work on

PR #4003, opened 2023-04-13
Update telegraf role to use a specific commit version

sanjaysrikakulam:
This should fix the issues with keys during installation.


topics/admin/tutorials/monitoring/tutorial.md
```diff
@@ -780,7 +781,7 @@ You can run the playbook now, or wait until you have configured Telegraf below:
 >    +    privs: SELECT
 >    +    objs: ALL_IN_SCHEMA
 >    +
->
```
hexylena, line 783: these whitespace changes are not ideal, they're in the commit blocks which may break the export process. Could you please restore the whitespace? I would recommend turning off any "remove trailing whitespace" plugin for the GTN repository as here it is meaningful in the admin diffs.

PR #4004, opened 2023-04-13
Update telegraf role to use a specific commit

sanjaysrikakulam:
This should fix the issues with keys during installation.


hexylena (approved):
looks good, thanks for fixing the whitespace

PR #4023, opened 2023-04-17
Repeatmasking: pre-smorg review

abretaud:
Reviewed the repeat masking tutorial before re-recording it. Mainly small cosmetic changes + added a "details" block about repeatmodeler.

Ran it successfully on .fr and .eu

### DIFF
--- a/topics/genome-annotation/tutorials/repeatmasker/tutorial.md
+++ b/topics/genome-annotation/tutorials/repeatmasker/tutorial.md
@@ -46,11 +46,11 @@ When you assemble a new genome, you get its full sequence in FASTA format, in th
 - Tandem repeats: small sequences (<60 base pairs) repeated next to each other, found in many places in the genome, in particular centromeres and telomeres
 - Interspersed repeats: sequences repeated in distant positions, including transposons, {SINEs} or {LINEs}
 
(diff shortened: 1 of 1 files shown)


hexylena:
if you ran it - did you have a workflow invocation? If so we could write a test for this, and test it automatially across the other N usegalaxy.*s

abretaud:
Good idea!

Also just saw the answer in https://github.com/galaxyproject/galaxy/issues/15945 + there are new repeatmasker/modeler releases => will see if I can update these before merging. So switching to draft mode

hexylena:
awesome!

shiltemann (approved):

abretaud:
Should be good now :crossed_fingers:

faqs/gtn/gtn_workflow_testing.md
```diff
@@ -51,4 +51,16 @@ contributors: [hexylena]
    ```
 
 9. You will need to check the `-tests.yml` file, it has some automatically generated comparisons. Namely it tests that output data matches the test-data exactly, however, you might want to replace that with assertions that check for e.g. correct file size, or specific text content you expect to see.
-10. Contribute all of those files to the GTN in a PR.
+
+10. If the files in test-data are already uploaded to Zenodo, to save disk space, you should delete them from the `test-data` dir and use their URL in the `-tests.yml` file, as in this example:
```
hexylena, line 55: :yellow_heart:

hexylena (commented):
Looks great!!! workflow tests are fantastic

hexylena (approved):

PR #4042, opened 2023-04-18
keep local_env in galaxy_job_config when tpv_dispatcher is added

cat-bro:
When tpv is added, environments are removed, including local_env.

the mapping:
```
>       tools:
>         - class: local # these special tools that aren't parameterized for remote execution - expression tools, upload, etc
>           environment: local_env
```

needs to be removed at this point or data fetch jobs fail. Another mapping (testing tool to tpv_dispatcher) is added later in the tutorial.

Making this a draft PR as I'm not sure how to handle the diffs here and later on.  Can we leave the tools section empty like so?
```
tools:
```


cat-bro:
That sounds like the safest solution.  I've edited this so that both the mapping between class: locals and local_env is kept, and local_env is also kept.

natefoo:
Looks good to me, I'll knit git-gat.

PR #4070, opened 2023-04-27
Update some tools in tuberculosis variant analysis tutorial

wm75:


wm75:
@pvanheus the linter complains about the WF not specifying a creator and a license. I'm hesitant to add a license in particular since it's not my WF.

pvanheus:
The WF was meant to be the steps of the tutorial itself. So can it fall under whatever license covers the tutorial?

wm75:
The content of the tutorials is CC BY 4.0-licensed, but you could argue that workflows are "code" more than tutorial content, and for that MIT would be roughly the equivalent (and also what the repo here uses).

pvanheus:
Ok then I'm happy to go with MIT license.

bgruening (approved):

PR #4076, opened 2023-04-27
Update tutorial to match Alevin 1.9

wee-snufkin:
@nomadscientist Generating a single cell matrix using Alevin tutorial updated!
- names of the output files in the tutorial changed accordingly 
- tools IDs changed to the newest version of the tool
- workflow using Alevin 1.9 added


topics/single-cell/tutorials/scrna-case_alevin/workflows/Galaxy-Workflow-Generating_a_single_cell_matrix_using_Alevin_1.9.ga
```diff
@@ -0,0 +1,1216 @@
...
+    "annotation": "This workflow generates a single cell matrix using Alevin v 1.9. \nUpdated April 2023",
+    "creator": [
+        {
+            "class": "Person",
+            "identifier": "0000-0002-8170-8806",
+            "name": "Wendi Bacon"
+        },
+        {
+            "class": "Person",
+            "identifier": "",
```
hexylena, line 12: do you have an orcid yet @wee-snufkin ? if so could you add it? We're also missing a license for this WF.

And if you have 15 minutes, and [could generate a test case](https://training.galaxyproject.org/training-material/faqs/gtn/gtn_workflow_testing.html), that would be incredible and we can then test all three usegalaxy.*s and generate example histories on each.

hexylena (approved):
linting issues discussed out of band, going with this for now until we can resolve a testing issue.

Issue #4021, opened 2023-04-17
Job conf - we should not set gravity handlers for jobs/workflows before we set the job config


hexylena:
Reported by @torfinnnome We setup

```
+    handlers:                                 
+      handler:                                
+        processes: 2                          
+        pools:                                
+          - job-handlers                      
+          - workflow-schedulers               
```

pretty early in the tutorial, before we actually need job handlers. As a result it is possible to see an error message for 

> dynamic handlers are configured in gravity, but galaxy is not configured to assign jobs to handlers dynamically

We should move the handlers setup down to the section on job configuration, it's probably a better time to discuss this topic as well

PR #4091, opened 2023-05-04
Typo/minor hiccups in ansible galaxy/IT

tuncK:
I noticed a few typos while following.

In addition, I observed some warnings etc. that I tried to clarify.


hexylena:
xref https://github.com/galaxyproject/training-material/issues/4021

PR #4093, opened 2023-05-04
Update the collection tutorial with screen capture when the interface changed

Delphine-L:
Updated images and gifs when the interface changed: tool forms for bwa-mem and collapse collection, and select all in the history


hexylena (approved):
in the future we definitely want to decrease the number of these images since the accessibility is generally quite poor, but for now this works. Maybe I can write some better alt text before smorgasbord.

PR #4100, opened 2023-05-10
Updated workflow with new version of Fasta Stats

AnnaSyme:
Updated the version of Fasta Stats in the workflow (in two places), and re-joined to required inputs.


shiltemann (approved):

shiltemann:
thanks @AnnaSyme!

PR #4102, opened 2023-05-10
Add versions of a few ML tools 

anuprulez:
Add versions of a few ML tools to point to correct tools


hexylena (approved):
thanks for 'upgrading' this tutorial to add that support!

bgruening:
Thanks @anuprulez. If you like feel free to add those to other trainings as well :)

PR #4109, opened 2023-05-12
Updates to assembly tutorial for large genome

AnnaSyme:
Two commits - one for user to check input datatypes, and one to recommend using either of Galaxy Eu or Au servers.

### DIFF
--- a/topics/assembly/tutorials/largegenome/tutorial.md
+++ b/topics/assembly/tutorials/largegenome/tutorial.md
@@ -141,6 +141,9 @@ We are also using a reference genome *Arabidopsis thaliana* for a later comparis
 >
 >    {% snippet faqs/galaxy/datasets_import_via_link.md %}
 >    {% snippet faqs/galaxy/datasets_import_from_data_library.md %}
+>    
+> 3. Check that the datatypes for the three files of sequencing reads are `fastq.gz`, not `fastqsanger.gz` and change datatype if needed.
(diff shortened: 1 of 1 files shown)


topics/assembly/tutorials/largegenome/tutorial.md
```diff
@@ -25,6 +25,8 @@ contributors:
 
 # Introduction
 
+*Note: We recommend running this tutorial on either the Galaxy Europe or Galaxy Australia servers. Other servers (such as Galaxy main) have not yet been configured fully for all the tools in this analysis.*
```
hexylena, line 28: If this note is specific to Smörgåsbord, I don't think we need to include it. Under every video we'll shown a list of which servers are supported (we check every tutorial against every tool version that's installed on each server.)

hexylena:
There's a linting error where you've skiped a heading level, could I trouble you to check into that?

topics/assembly/tutorials/largegenome/tutorial.md
AnnaSyme: Thanks @hexylena - sounds good, yes let's not include this.

AnnaSyme:
> There's a linting error where you've skiped a heading level, could I trouble you to check into that?

Yes, thanks, have fixed this in a new commit.

hexylena (approved):

PR #4120, opened 2023-05-16
Update 10x preprocessing tutorial

pavanvidem:
- uses latest STARSolo
- different order of the parameters for STARSoloe
- Slightly different numbers from the previous tutorial
- Summary stats from the latest version has a different naming scheme 
- Rank barcode plot y-axis title was changed


hexylena (commented):
Looks good, could you fix the [heading level issues](https://github.com/galaxyproject/training-material/pull/4120/files#diff-3dc441b5992a5366bbac835d3f33f92e0bcf112c33c9bc320896790ad7285b52R142) and then we can merge :)

hexylena (approved):

hexylena:
We should probably update the associated workflow at some point soon as well.

pavanvidem:
thanks, @hexylena

PR #4135, opened 2023-05-17
[Metagenomics] Enable click-to-run workflows for 16S tutorial

shiltemann:
it was already enabled for the first workflow in the tutorial, but not the others, want to show off this super nice feature for the smorg video


hexylena:
sorry I made the first one as a test and then when it worked, forgot to apply it further

PR #4150, opened 2023-05-18
16S short: update workflows

shiltemann:
couple more workflow fixes


hexylena (approved):

hexylena:
if you have histories where you've run these, then we can extract workflow tests (but let's save that for next week)

PR #4155, opened 2023-05-21
Update ToolFactory introductory module. Add new docker installation option

fubar2:
New installation instructions and various small clarifications,

Backend has been substantially simplified - no more Planemo - using a local test script instead. Without new deps, a tool can be built and installed within about 10-15 seconds now.

Revised the docker image so it's far simpler, for a fast but non-persistent installation compared to a normal desktop dev server.


fubar2:
needs fixing - header levels :(

PR #4156, opened 2023-05-21
Updates to both ToolFactory topics to fix broken installation instructions

fubar2:
TF has been simplified by removing planemo from the testing part. Now uses galaxy-tool-test so is much faster and more stable as a result. Works well with sqlite so that's the default.

A non-persistent docker image has been created using a simple overlay to a standard 23.0 git checkout.
Instructions for building a local desktop version take longer but give a normal persistent dev server.


fubar2:
I must lint first
I must lint first...

PR #4158, opened 2023-05-21
Updates to both ToolFactory topics to fix broken installation instructions.

fubar2:
TF code has been greatly simplified by removing planemo from the testing part. Now uses **galaxy-tool-test** instead, so is much faster and more stable. 
Works well with sqlite - defined as happily running the workflow to regenerate the samples, so that's the default.
No real changes to the underlying tool form.

A new non-persistent docker image has been created using a simple overlay to a standard 23.0 git checkout.
Instructions for building a local desktop version take longer but give a normal persistent dev server.


fubar2:
Interesting - the Accessibility linting failed with a message about someone else's material so I guess that needs fixin' too but not by me.....
```
Run bundle exec htmlproofer \
Running 3 checks (Links, Images, Scripts) in ["./_site"] on *.html files...


Checking 7135 internal links
Checking internal link hashes in 675 files
For the Links > Internal check, the following failures were found:

* At ./_site/training-material/news/2023/05/15/isoform-usage-training.html:244:

  internally linking to /training-material/topics/transcriptomics/tutorials/differential-isoform-expression/tutorial.md, which does not exist

* At ./_site/training-material/topics/one-health/index.html:1984:

  internally linking to /training-material/hall-of-fame/tklingstrom/, which does not exist
```

hexylena (approved):
looks good overall!

if possible, in the future you might consider updating an existing PR rather than making a new one. If you didn't want the commits on that branch you can also use force push (not the star wars kind) to fix a branch. (ask us if you need guidance on doing that)

I've fixed the issues that aren't your fault, I caused them on friday and forgot about them. Thanks for reminding me about them :)

fubar2:
Thanks @hexylena - the night before Smörgåsmas too! 
Managed to screw up that old PR for the galaxy project tutorial - will resubmit now this is in and I don't have to screw with git branches to keep PR's separate :(

I got a question reminding me that docker was broken. so changed to use galaxy-tool-test instead of that big ol' planemo hammer.
Docker/sqlite is stable running the workflow that generates all the example tools too - partly because only only one TF build job will run at a time so conda does not get all confused. 
Best wishes for next week - looking forward to seeing you cuddling a koala!

PR #4163, opened 2023-05-22
Update pox-virus-tiled-amplicon-ref-masking.ga

wm75:
Make the input label in the WF match the name used in the WF test.
Should fix one of the items in @natefoo's https://gist.github.com/natefoo/2c02814fffc7384e833c5df0e8f28b00#missing-inputs.


bgruening (approved):

hexylena:
This is surely something we can lint on, whether labels mentioned in WF tests match something in the workflow.

PR #4192, opened 2023-06-06
Show off GTN stats in more places

shiltemann:
Added some more stats cards to the contributors section on the homepage and the hall of fame page itself

image


hexylena (approved):
Looks fantastic!

PR #4193, opened 2023-06-08
New tutorial: Scanpy Parameter Iterator

wee-snufkin:
A wee tutorial falling under Tips and Tricks single-cell category. An extension of [Filter, Plot and Explore tutorial](https://training.galaxyproject.org/training-material/topics/single-cell/tutorials/scrna-case_basic-pipeline/tutorial.html).

Here it is @nomadscientist!


wee-snufkin:
hm, is the output testing mandatory now?

hexylena:
> hm, is the output testing mandatory now?

yep, changed that probably last month as we noticed a couple where we didn't verify outputs at all, and we're trying to (slowly) push folks to best practices.

wee-snufkin:
> yep, changed that probably last month as we noticed a couple where we didn't verify outputs at all, and we're trying to (slowly) push folks to best practices (sorry)

ok ok, good to know. But now when running planemo workflow_test_init it automatically returns 'outputs: {}' (it didn't last month) - how to work around that?

hexylena:
> But now when running planemo workflow_test_init it automatically returns 'outputs: {}' (it didn't last month) - how to work around that?

oh that's unexpected that that behaviour changes. What version of planemo are you using? I'm on planemo, version 0.75.7 which still produces outputs properly

wee-snufkin:
I'm also running version 0.75.7. But apparently it's the problem with the workflow itself - just realised it didn't produce anything in 'outputs' tab when I checked it in 'Invocations' in Galaxy. Weird as the older workflows have 'outputs' displayed normally. I'll look into the workflow again and try to fix it. 
But when it eventually works as expected, where should the output test files be uploaded? Should I add test-data folder (large files ;/) or upload the output files to Zendo and add links as with the inputs? Apologies if it's already described somewhere that I missed

hexylena (approved):

PR #4197, opened 2023-06-09
FAIR intro: fix typos

abretaud:
Fixed two minor typos.

I also noticed that the "edit on github" link is broken on https://training.galaxyproject.org/training-material/topics/fair/tutorials/fair-intro/tutorial.html, there's a `/` missing


hexylena:
huh how is the check passing, there's no alt text.

PR #4205, opened 2023-06-14
Fix word duplication in collections tutorial

hexhowells:


hexylena (approved):
Great catch, thanks @hexhowells and welcome to the GTN!

PR #4209, opened 2023-06-14
Fix typo in workflow-editor tutorial

hexhowells:


shiltemann:
good catch, thanks @hexhowells!

PR #4217, opened 2023-06-19
implement simple JS bundling

hexylena:
It smushes the JS files together at the end of the build. dead simple.

also moves some of the default JS we run on every page into `main.js` which makes a bit more sense? and preloads the bundle.


hexylena:
I think this is fine to go. It's pretty simple and will make requests like, a tiny bit more efficient in prod.

shiltemann (approved):

PR #4219, opened 2023-06-21
add a Webfinger endpoint, allowing mentioning fedi users by GTN handle

hexylena:
This can make it easier to announce news on the fedi, as we can write things like "Thanks @abretaud@training.galaxyproject.org" and it'll mention @abretaud's account :)

requires https://github.com/galaxyproject/infrastructure-playbook/pull/44


shiltemann (approved):

abretaud:
Nice! feeling sooo social :D

PR #4243, opened 2023-06-29
In Identifying tuberculosis - Changed Concatenate datasets to Collapse Collection on data

T0ront0:
Updating Identifying tuberculosis transmission links: from SNPs to transmission clusters, 
Analysis Part 2 under "Clean up the table",
In step 1. Replace Text Tool: 

“File to process”: Single file (output of Concatenate datasets tool)  should refer to the previous step, which was "Collapse Collection" so it would clear up any confusion.


wm75 (approved):
Very well spotted, thanks!

bgruening:
Cool thanks a lot @T0ront0

PR #4248, opened 2023-06-30
Add two more Mastodon accounts (and a matrix id)

wm75:
cause who wouldn't want to profit from https://mstdn.science/@gtn/110628391201479353 :-)


hexylena (approved):
hah :D wondered if it would be motivating

hexylena:
@hrhotz you're also welcome to join in

hrhotz:
Thanks for the heads-up
I will add my accounts within the next few days

hrhotz:
done, see: https://github.com/galaxyproject/training-material/pull/4256

PR #4258, opened 2023-07-07
Change transparent image in Clustering Tutorial so it is readable in dark mode

hrukkudyr:
Tutorial has a transparent svg image - this is not viable when viewed in dark mode. The svg image has been converted to png with a white background and the file uploaded. Markdown changed to link to new file.


hexylena:
a white background could be added to the SVG with the same effect right? it would mean faster loading for users and maybe that's preferable.

hrukkudyr:
Hi – I looked into that but couldn’t see an easy (for me) way of doing that using the markdown. It’s normally done using a CSS style to put a white background onto the containing element. I couldn’t see a way of doing that without changing the main StyleSheet?Somebody who knows the markdown/CSS better than me may have a solution for that? From: HelenaSent: 08 July 2023 08:34To: galaxyproject/training-materialCc: ***@***.***; AuthorSubject: Re: [galaxyproject/training-material] Change transparent image in Clustering Tutorial so it is readable in dark mode (PR #4258) a white background could be added to the SVG with the same effect right? it would mean faster loading for users and maybe that's preferable.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

hexylena:
oh, no, not in the CSS but directly in the SVG, I think it would be easy to maybe just put a large white rectangle behind the drawing or so? I don't think we want to change the CSS currently

shiltemann:
Hi @hrukkudyr! I applied @hexylena suggestion but could not push the changes to your PR here, so I made a new PR here: https://github.com/galaxyproject/training-material/pull/4636

Thanks a lot for noticing this and creating the PR!

PR #4279, opened 2023-07-13
User Preferences

hexylena:
This moves user preferences into a new page just for them.

- Change theme preferences
- Enable opting out of data collection¹
- See what data we collect

Previously our themes were hardcoded as dark or light, depending on the theme. This splits out the important axes relevant to users:

- Theme: Light/Dark/System (defaults to system, need to double check this.)
- Light theme preference: pure white, or yellow-ish
- Dark theme preference: pure black, or a dark grey
- Theme preference: rainbows/bhm/'straya/etc
- Contrast preference: low/medium/high

I wish github did the same, letting me be colorblind + high contrast but we're better than them ;) It is a tad hard to manage but I think it's worth it.

TODO:

- [x] ~~double check system theme behaviour, also without JS~~  does not work without JS
- [x] ~~replace the dropdowns with small pictures~~ too much work
- [x] ensure every combination of every theme looks right everywhere. (yikes)
- [x] respect DNT

¹ I know some folks prefer opt-in, but given the data we collect (nothing private, no ability to correlate), our need for this data for continued funding outweighs the non-existent privacy concerns.


hexylena:
We will be ignoring the a11y complaint on the tools from scratch tutorial due to the inability to properly check if we're in a markdown code block or not currently.

This could use a pair of eyes if anyone has some free time.

hexylena:
I've added licenses for all the fonts and exposed them at a licenses.md page in the root folder as well. Additionally there was a variable weight version of yanone kaffeesatz so we could replace ours with that and yeah. This should be ready to be tested (the linting failure in the one tutorial is expected.)

shiltemann:
thanks @hexylena !

PR #4280, opened 2023-07-13
Small update to cron job for cleanup

pauldg:
`gxadmin galaxy cleanup` requires some python scripts which require dependencies installed in the virtual env unless these are all covered by pre-task? If the latter is the case please ignore and close. If not, then it's possible better to move this to one of the `galaxyproject.galaxy` tasks?


hexylena (approved):
yeah nah looks good mate

PR #4292, opened 2023-07-17
Update exome-seq tutorial/WFs to latest tool versions

wm75:
- Replace outdated samtools filter tool with Samtools view
- Provide tool links
- Make sure users end up with a dbkey on their VCF in all cases


bgruening (dismissed):
LGTM, thanks or adding the tool IDs.

bgruening (approved):

PR #4302, opened 2023-07-21
use igraph instead of python-igraph, as the latter name is deprecated for "pip install"

szhorvat:
Please see https://github.com/igraph/python-igraph/issues/699 for why this is necessary.


mtekman (approved):
It seems I can install both `python-igraph` and `igraph`, but the latter appears to have a higher version number, so it seems more legitamite.

PR #4309, opened 2023-08-01
Remove missing tool

paulzierep:
# This tool is not available on EU or Org and cannot be found in the toolshed.

* Users were requesting this tool, but the source is unknown.
* Additional not to the authors: all hands-on boxes still have the "Task description" label, this should be replaced with a meaningful text.

### DIFF
--- a/topics/transcriptomics/tutorials/full-de-novo/tutorial.md
+++ b/topics/transcriptomics/tutorials/full-de-novo/tutorial.md
@@ -201,20 +201,6 @@ Why do we need to correct those?
 {: .hands_on}
 
 # Assembly assessment / cleaning
-
-## Checking of the assembly statistics
-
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Probably this one https://toolshed.g2.bx.psu.edu/view/anmoljh/trinitystats/8fd7b0c6ec0e

But maybe there are better?

paulzierep:
thanks @bernt-matthias; I am checking with the authors if it can be added

paulzierep:
Fixed with: https://github.com/galaxyproject/tools-iuc/pull/5434
But I would suggest to use:  `{% tool [Porechop](toolshed.g2.bx.psu.edu/repos/iuc/porechop/porechop/0.2.4) %}` instead of `{% icon tool %}` for all tools.

PR #4310, opened 2023-08-03
testing_parameter_iterator_updates

nomadscientist:
I'm testing the parameter iterator tutorial as a user and providing edits as I go

### DIFF
--- a/topics/single-cell/tutorials/scanpy_parameter_iterator/tutorial.md
+++ b/topics/single-cell/tutorials/scanpy_parameter_iterator/tutorial.md
@@ -25,7 +25,7 @@ requirements:
         - scrna-case_alevin-combine-datasets
         - scrna-case_basic-pipeline
 
-time_estimation: 1H
+time_estimation: 2H
 
 key_points:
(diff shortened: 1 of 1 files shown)


nomadscientist:
Other notes

- [ ] I didn't understand the final comment in the tutorial
- [ ] A couple of tools function oddly, and it would be good to make issue reports on them in addition to the tutorial workarounds included (messaged author directly)
- [ ] Requested image to help with dataset workaround (https://github.com/galaxyproject/training-material/issues/4311)
- [ ] Data is not in the data library

Other than these minor points, this tutorial was super fun and I imagine very popular whenever students ask about which parameters to use! "We have a tutorial for that"! Woo!

wee-snufkin:
Thanks for testing and your feedback! I'll have a look and will try to fix the points you mentioned :)

wee-snufkin (approved):
Thanks for your edits @nomadscientist ! 
I'm approving all your changes and I've removed some stuff discussed as unnecessary / not applicable.

Issue #4305, opened 2023-07-25
Automated notebooks don't include images


nomadscientist:
We're making a bunch of automated notebook tutorials and have just realised that images embedded (and successfully rendering in the GTN site) don't appear in the notebooks. 
Is there a way to make them pull from their GitHub repo like in the tutorial?
![Screenshot 2023-07-25 at 13 10 39](https://github.com/galaxyproject/training-material/assets/44605769/669f2b34-f6f6-4d9f-b8c3-9a9ae18dcdf6)

@hexhowells found this

hexylena:
they're definitely supposed to, the images are encoded as a base64 PNG file within the notebook intentionally so they work without network access. I'm not sure why they aren't showing up, which tutorial is this so I can reproduce it?

nomadscientist:
https://training.galaxyproject.org/training-material/topics/single-cell/tutorials/scrna-case_JUPYTER-trajectories/tutorial.html

nomadscientist:
https://training.galaxyproject.org/training-material/topics/single-cell/tutorials/scrna-case_JUPYTER-trajectories/tutorial.html

PR #4315, opened 2023-08-10
New Filter, Plot and Explore (Jupyter Notebook) tutorial

hexhowells:
Built the Filter, Plot and Explore tutorial in a Jupyter notebook, my first tutorial contribution.

---

The tutorial is largely complete but set as a draft as there are two issues found with the automated conversion to a Jupyter notebook:
- Images from the tutorial don't load in the notebook as its trying to fetch from a absolute filepath, this is already a raised issue here: https://github.com/galaxyproject/training-material/issues/4305#issue-1820195125
- Question boxes containing multiple solution dropdown menus don't seem to render correctly in the notebook causing part of the cell to be hidden. A quick work around to this can be to create a separate question box for each question. I can raise this as a separate issue if needed.


hexhowells:
The question box in the tutorial causing issues in the notebook can be found [here](https://github.com/galaxyproject/training-material/pull/4315/files#diff-5c475fb0a863c46fc667f6166bbc812000739bc18e78ec2449b0376f15ad282aR224)

nomadscientist:
Images must be indented to match text or numbering restarts

diff --git a/topics/single-cell/tutorials/scrna-case-jupyter_basic-pipeline/tutorial.md b/topics/single-cell/tutorials/scrna-case-jupyter_basic-pipeline/tutorial.md
index c5874d6e5e43..e81069717ffe 100644
--- a/topics/single-cell/tutorials/scrna-case-jupyter_basic-pipeline/tutorial.md
+++ b/topics/single-cell/tutorials/scrna-case-jupyter_basic-pipeline/tutorial.md
@@ -230,7 +230,7 @@ Now that we've assessed the differences in our samples, we will look at the li
 > > <solution-title></solution-title>
 > >
 > > 1. Any plot with `log1p_n_genes_by_counts` would do here, actually! Some people prefer scatterplots to violins.
-> > Scatter-genesxmito")
+> >    Scatter-genesxmito")
 > >
 > > 2. In `Scatter - mito x genes` you can see how cells with `log1p_n_genes_by_counts` up to around, perhaps, `5.7` (around 300 genes) often have high `pct_counts_mito`.
 > >   - You can plot this as just `n_counts` and see this same trend at around 300 genes, but with this data the log format is clearer so that's how we're presenting it.
@@ -247,7 +247,7 @@ Now that we've assessed the differences in our samples, we will look at the li
 > > <solution-title></solution-title>
 > >
 > > 1. As before, any plot with `log1p_total_counts` will do! Again, we'll use a scatterplot here, but you can use a violin plot if you wish!
-> > Scatter-countsxmito")
+> >    Scatter-countsxmito")
 > >
 > > 2. We can see that we will need to set a higher threshold (which makes sense, as you'd expect more UMI's per cell rather than unique genes!). Again, perhaps being a bit aggressive in our threshold, we might choose `6.3`, for instance (which amounts to around 500 counts/cell).
 > >   - In an ideal world, you'll see a clear population of real cells separated from a clear population of debris. Many samples, like this one, are under-sequenced, and such separation would likely be seen after deeper sequencing!
@@ -262,7 +262,7 @@ Now that we've assessed the differences in our samples, we will look at the li
 > > <solution-title></solution-title>
 > >
 > > 1. Any plot with `pct_counts_mito` would do here, however the scatterplots are likely the easiest to interpret. We'll use the same as last time.
-> > Scatter-countsxmito")
+> >    Scatter-countsxmito")
 > >
 > > 2. We can see a clear trend wherein cells that have around 5% mito counts or higher also have far fewer total counts. These cells are low quality, will muddy our data, and are likely stressed or ruptured prior to encapsulation in a droplet. While 5% is quite a common cut-off, this is quite messy data, so just for kicks we'll go more aggressive with a `4.5%`.
 > >    - In general, you must adapt all cut-offs to your data - metabolically active cells might have higher mitochondrial RNA in general, and you don't want to lose a cell population because of a cut-off.

nomadscientist:
My deepest apologies

diff --git a/_plugins/notebook.rb b/_plugins/notebook.rb
index f210167c6ad8..056d42bcca52 100644
--- a/_plugins/notebook.rb
+++ b/_plugins/notebook.rb
@@ -449,7 +449,7 @@ module GTNNotebooks
     notebook
   end
 
-  def self.renderMarkdownCells(site, notebook, metadata, _page_url, _dir)
+  def self.renderMarkdownCells(site, notebook, metadata, _page_url, dir)
     seen_abbreviations = {}
     notebook['cells'].map do |cell|
       if cell.fetch('cell_type') == 'markdown'
@@ -522,30 +522,30 @@ module GTNNotebooks
         # love it) or we'll link to the production images and folks can live
         # without their images for a bit until it's merged.
 
-        # if cell['source'].match(/<img src="\.\./)
-        #   cell['source'].gsub!(/<img src="(\.\.[^"]*)/) do |img|
-        #     path = img[10..]
-        #     image_path = File.join(dir, path)
-        #
-        #     if img[-3..].downcase == 'png'
-        #       # puts "[GTN/Notebook/Images] Embedding png: #{img}"
-        #       data = Base64.encode64(File.binread(image_path))
-        #       %(<img src="data:image/png;base64,#{data}")
-        #     elsif (img[-3..].downcase == 'jpg') || (img[-4..].downcase == 'jpeg')
-        #       # puts "[GTN/Notebook/Images] Embedding jpg: #{img}"
-        #       data = Base64.encode64(File.binread(image_path))
-        #       %(<img src="data:image/jpeg;base64,#{data}")
-        #     elsif img[-3..].downcase == 'svg'
-        #       # puts "[GTN/Notebook/Images] Embedding svg: #{img}"
-        #       data = Base64.encode64(File.binread(image_path))
-        #       %(/, '<pre style="color: inherit; background: transparent">')

nomadscientist:
The Violin - genotype - log") image is incorrect, it should look like this: 
Screenshot 2023-08-22 at 14 54 19

nomadscientist:
On that vein, I strongly recommend remaking the data outputs. They are slightly different due to code / formatting, so it would be better to use those outputs rather than the ones from the GUI

nomadscientist (approved):
Great job Morgan!
I tested the tutorial and it is working great :)

bgruening (approved):
Very cool. Thanks a bunch, @hexhowells - this will be very useful.

PR #4334, opened 2023-09-04
Add eye-slash icon

wee-snufkin:
Adding eye-slash icon and updating the Parameter Iterator tutorial which needed this


shiltemann:
perfect! thanks @wee-snufkin!

PR #4339, opened 2023-09-11
Set up pip_virtual_command with "--copies"  in group_vars/galaxyservers.yml (admin tutorial)

vazovn:
# Thank you for contributing to the GTN! :yellow_heart:

## Please check that:

- [ ] Your pull request has a **good title**, e.g.
    - "Fix typo in ansible-galaxy tutorial"
    - "Add new transcriptomics tutorial covering a new sequencing technology"
- [ ] Please **replace this entire pull request message** with a *description of your changes*, it will help us review your pull request faster
- [ ] Check that your images are allowed to be re-hosted by the GTN

Once these are done, you're ready to go!

<details>
    <summary>Pip virtual command setup</summary>

Version 23.01. Running pip_virtual_command without "--copies"  created an issue which was difficult to debug - jobs, both local and remote (cluster) would fail to start. They would only start if the galaxy framework is restarted (galaxyctl restart) and would throw an error `lib/galaxy/model/__init__.py: ModuleNotFoundError: No module named 'sqlalchemy'` Actually venv fails to install properly and the jobs from the cluster would ignore the venv's packages which do contain sqalchemy. Deleting the venv and restarting the ansible playbook with "--copies" switch solved the issue.

</details>


topics/admin/tutorials/ansible-galaxy/group_vars/galaxyservers.yml
```diff
@@ -1,6 +1,6 @@
 ---
 # Python 3 support
-pip_virtualenv_command: /usr/bin/python3 -m virtualenv # usegalaxy_eu.certbot, usegalaxy_eu.tiaas2, galaxyproject.galaxy
+pip_virtualenv_command: /usr/bin/python3 -m virtualenv --copies # usegalaxy_eu.certbot, usegalaxy_eu.tiaas2, galaxyproject.galaxy
```
hexylena, line 3: Ah this is unfortunate, this file hasn't existed in some years. Could you please instead

0. Update your fork first, the folder should look like this: https://github.com/galaxyproject/training-material/tree/main/topics/admin/tutorials/ansible-galaxy
1. Add it to topics/admin/tutorials/ansible-galaxy/tutorial.md *as* a tip box, instead of an actual patch/change for everyone, as I guess smaller galaxies could benefit from the disk space savings of *not* using that flag, right?
2. In that tip box describe when this would be useful, maybe title it something like "Running a Large Galaxy on a Cluster? Copies can help" to make it clear it's for those use cases? (please change accordingly), and suggest adding it to `pip_virtualenv_command`, if you could also include some of the errors/failure modes/error messages that you saw that would be helpful to people searching for that specific text, to find your tip.

I think that would achieve all of our goals: instruct both small and large galaxy users, and teach them when you need specific flags like this rather than just setting some magic.

PR #4344, opened 2023-09-12
Adding banner to show that FTP is no longer supported on UseGalaxy.org

natwhitaker:
Since FTP is no longer supported on UseGalaxy.org, we are adding banners to let users know and direct them to alternatives.

### DIFF
--- a/faqs/galaxy/datasets_upload_ftp.md
+++ b/faqs/galaxy/datasets_upload_ftp.md
@@ -3,27 +3,31 @@ title: Upload many files (>10) via FTP
 area: data upload
 box_type: tip
 layout: faq
-contributors: [bebatut]
+contributors: [bebatut, shiltemann]
 ---
 
-<div class="alert alert-success trim-p" role="alert">
-UPDATE: FTP upload is no longer supported on UseGalaxy.org. For more information and to learn about alternative options, please see https://help.galaxyproject.org/t/the-usegalaxy-org-ftp-service-will-be-decommissioned-on-august-12-2022/8318.
-</div>
+Some Galaxies offer FTP upload for very large datasets.
 
-1. Make sure to have an FTP client installed
+1. Check that your Galaxy supports FTP upload and look up the FTP settings.
+   - **Galaxy EU** (usegalaxy.eu): [FTP settings](https://usegalaxy-eu.github.io/ftp/)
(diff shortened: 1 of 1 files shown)


faqs/galaxy/datasets_upload_ftp.md
```diff
@@ -6,6 +6,10 @@ layout: faq
 contributors: [bebatut]
 ---
 
+<div class="alert alert-success trim-p" role="alert">
+UPDATE: FTP upload is no longer supported on UseGalaxy.org. For more information and to learn about alternative options, please see https://help.galaxyproject.org/t/the-usegalaxy-org-ftp-service-will-be-decommissioned-on-august-12-2022/8318.
+</div>
+
```
hexylena, line 12: this is a very .. broad place to put this, I'm not sure it's the best targeting for your users? If we were to include it, it would need to look like:

```suggestion
> <tip-title>FTP Support on UseGalaxy.org</tip-title>
> FTP upload is no longer supported on UseGalaxy.org. For more information and to learn about alternative options, please see [the documentation](https://help.galaxyproject.org/t/the-usegalaxy-org-ftp-service-will-be-decommissioned-on-august-12-2022/8318)
{: .tip}
```

but again I feel a bit odd about this, given that this is simply an FAQ about FTP, which is seen by users on .eu, .org.au, and hundreds of smaller galaxies.

No matter what we do, I suggest removing usegalaxy.org below, I'd suggest a change but you can't do that for unmodified lines in the github ui.

natwhitaker: I understand and agree! This was on just the list, but not a problem; I will let them know!

shiltemann: We could also just add a new step 1 "Check that your Galaxy supports FTP uploads", then potentially mention which of the big 3 do and don't?

shiltemann: @natwhitaker I reworked it a bit, still should make it clear that FTP is unavailable on Main.

@cat-bro: I could only find a post from 2019 mentioning FTP settings for Galaxy Australia, is this still accurate? is there a better place to link to where the latest FTP settings can always be found (if you still have it enabled)?

shiltemann (approved):

PR #4350, opened 2023-09-19
Neoformit patch 1

neoformit:
In the [Nanoplot section of the sequence quality tutorial](https://training.galaxyproject.org/training-material/topics/sequence-analysis/tutorials/quality-control/tutorial.html#assess-quality-with-nanoplot---long-reads-only) the plots shown are for a different dataset that the one used in the tutorial (`m64244_210612_174252` is shown instead of `m64011_190830_220126`). This is confusing because users will wonder why the results in Galaxy look different to those shown in the tutorial.

This PR replaces those plot images with files corresponding to the correct dataset, and makes a minor correction to the tutorial text to match the corrected plot images:

Instructions in GTN:
image

Output shown in GTN:
image

Output observed in Galaxy:
image


shiltemann (approved):

shiltemann:
Thanks @neoformit!

PR #4355, opened 2023-09-22
Add python memory leak debugging faq

hexylena:
@mvdbeek this is based on your message in the admins channel, is this OK for you?

### DIFF
--- a/topics/admin/faqs/python-memory-leaks.md
+++ b/topics/admin/faqs/python-memory-leaks.md
@@ -26,6 +26,8 @@ It might also be useful to just check what the process is doing with py-spy dump
 
 You can follow web workers in gunicorn with
 
-    memray run --follow-fork -o the_dump gunicorn 'galaxy.webapps.galaxy.fast_factory:factory()' --timeout 600 --pythonpath lib -k galaxy.webapps.galaxy.workers.Worker -b localhost:8082 --config python:galaxy.web_stack.gunicorn_config -w 1 --preload
+```console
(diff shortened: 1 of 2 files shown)


topics/admin/faqs/python-memory-leaks.md
```diff
@@ -0,0 +1,27 @@
...
+
+```console
+memray flamegraph --leaks --temporal the_dump -o the_dump.html
+```
+
+would then produce a report that shows allocation made but not freed over time.
+
+It might also be useful to just check what the process is doing with py-spy dump.
+
+If you're debugging a Galaxy process, and a user is splitting a dataset into a million element collection that could use some memory, however we have some pretty stringent limits for this now and haven't had a problem in a while.
```
mvdbeek, line 27:
```suggestion
```
I guess that's not very relevant and taken out of context

mvdbeek, line 27:
```suggestion
You can follow web workers in gunicorn with

memray run --follow-fork -o the_dump gunicorn 'galaxy.webapps.galaxy.fast_factory:factory()' --timeout 600 --pythonpath lib -k galaxy.webapps.galaxy.workers.Worker -b localhost:8082 --config python:galaxy.web_stack.gunicorn_config -w 1 --preload

the traced app will run on port 8082, you can then for instance in an upstream nginx section direct a portion of the traffic to your profiled app.
```

I guess that's not very relevant and taken out of context, instead this is how to follow gunicorn workers.

mvdbeek (dismissed):
Thanks, good idea to record it here.

topics/admin/faqs/python-memory-leaks.md
hexylena: Sure, that makes sense!

hexylena:
If you could re-approve @mvdbeek we'll get this merged, thanks for the additional command, that'll be useful

mvdbeek (approved):

PR #4361, opened 2023-09-25
Fix for tutorial nav back button

hujambo-dunia:
- Quick fix for overlapping icons on GTN.
- Related to [this issue](https://github.com/galaxyproject/galaxy/issues/16634) on the main Galaxy repo.
- Putting PR in "draft" until @shiltemann has time to preview.
- Preview of suggested change below:

screenshot_Before_After

### DIFF
--- a/assets/css/slides.scss
+++ b/assets/css/slides.scss
@@ -184,7 +184,7 @@ body.remark-container {
 		font-size: 10pt;
 		position: absolute;
 		left: 15px;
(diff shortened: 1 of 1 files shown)


hexylena (approved):
Thank you for the quick fix, that's already an improvement! We can fix it properly but even this is a big improvement now.

Also, thanks for the preview, that's so so helpful!

PR #4364, opened 2023-09-27
Workflow testing display / UI

hexylena:
- [x] workflow page listing
- [ ] topic listing
- [ ] by-server listing

### DIFF
--- a/_layouts/topic.html
+++ b/_layouts/topic.html
@@ -46,8 +46,12 @@ layout: base
     {{ content }}
 
     {% if topic.type == "use" %}
-        <h2 id="docker_image">Galaxy instances</h2>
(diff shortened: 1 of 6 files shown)


hexylena:
Closing this as I no longer have time/availability to implement it sadly. It would be nice to do in the future so I'll tag it appropriately.

hexylena:
there are a lot of components in place, `bin/workflow-test.rb` works today, but it suffers from very long execution times as it runs on real servers. https://github.com/galaxyproject/planemo/issues/1290 https://github.com/galaxyproject/planemo/issues/1368 remain relevant for our use case, it would be really good to be able to launch workflows quickly, and then a day later check in to see if they completed, we don't need to be wasting compute time doing `curl` on a loop.

PR #4308, opened 2023-07-28
Proposed changes to job-destinations and ansible-galaxy admin tutorials to prevent ansible-playbook errors.

Edmontosaurus:
Following the job-destinations and ansible-galaxy instructions as they currently are, will both result in failed ansible playbook runs. See the details about the problems and my proposed fixed below. If my proposed fixes are not ideal, then feel free to reject them and just use this pull-request as informational messages about current problems in these tutorials:

* ansible-galaxy tutorial: Following this tutorial will result in ansible playbook errors at line 926 where the variable "miniconda_version: 4.12.0" is defined. This tutorial used to work without errors but recently the miniconda role fails when miniconda_version 4.12 is defined. The cause of this is because the miniconda role by default attempts to install the latest conda version and then downgrades to the version defined in "miniconda_version". The latest miniconda versions can't be downgraded to 4.12 because a 'UnsatisfiableError' error will occur because of python version constraints. (the latest conda version being installed by the role uses python 3.11 which causes the version constraint error.) One workaround for this problem is to define extra variables, as can be seen in the diff. Two other possible solutions are to have the playbook install the latest conda version or to modify the code of the miniconda_role.

* job-destinations tutorial: In this tutorial, the student is instructed to run the ansible playbook after having added the usegalaxy_eu.tpv_auto_lint role. However running this playbook at this point will also result in ansible errors. This happens because TPV itself has not yet been installed. Skipping the playbook run at this point and only running it after the "Configuring TPV" section proved to be a workaround for me. As mentioned in the tutorial, adding TPV as a runner will cause Galaxy to automatically install TPV. The tpv_auto_lint role will only work at this point because it requires TPV to be installed. My proposed fix would be to remove the instruction to run the ansible-playbook at this point and run it after the "Configuring TPV" section.

# Thank you for contributing to the GTN! :yellow_heart:

## Please check that:

- [ ] Your pull request has a **good title**, e.g.
    - "Fix typo in ansible-galaxy tutorial"
    - "Add new transcriptomics tutorial covering a new sequencing technology"
- [ ] Please **replace this entire pull request message** with a *description of your changes*, it will help us review your pull request faster
- [ ] Check that your images are allowed to be re-hosted by the GTN

Once these are done, you're ready to go!

<details>
    <summary>Consider sustainable computing with Draft Mode :green_heart:</summary>

Is your pull request finished? 100% ready to be merged? Or do you want some time to work on it first (which we encourage! It's great to have visibility)

[...]

Issue #4322, opened 2023-08-22
Add Galaxy version information to the GTN admin materials


hechth:
Currently it is quite difficult to figure out whether a feature covered in the admin training is actually supported by a Galaxy version and I think some info in the beginning of the admin training saying which galaxy version this training is targeting would be very useful.

hexylena:
Generally you can check the ansible-galaxy tutorial for what the 'current' version is.  Would you expect this on every GAT tutorial?

hexylena:
@hechth if you have more input on this let me know :) (just where you would expect to see a notice/which tutorials/etc.)

PR #4370, opened 2023-10-03
Fix fatal errors in GxIT and ansible tutorials

tuncK:
I noticed that the IT tutorial (missing config file) and the ansible galaxy tutorial (conda cannot resolve) do not work out of the box anymore. This PR proposes necessary corrections and version updates so that the tutorials work with the current version of Galaxy.

### DIFF
--- a/topics/admin/tutorials/ansible-galaxy/tutorial.md
+++ b/topics/admin/tutorials/ansible-galaxy/tutorial.md
@@ -275,7 +275,7 @@ We have codified all of the dependencies you will need into a YAML file that `an
 >    @@ -0,0 +1,13 @@
 >    +# Galaxy, Postgres, Nginx
 >    +- src: galaxyproject.galaxy
->    +  version: 0.10.15
+>    +  version: 0.10.14
 >    +- src: galaxyproject.nginx
 >    +  version: 0.7.1
 >    +- src: galaxyproject.postgresql
@@ -894,7 +894,7 @@ The configuration is quite simple thanks to the many sensible defaults that are 
 >    `galaxy_layout`                 | `root-dir`                                                       | This enables the `galaxy_root` Galaxy deployment layout: all of the code, configuration, tools, and mutable-data (like caches, location files, etc.) folders will live by default beneath `galaxy_root`. User data is stored under `file_path`, a variable we will set later.
 >    `galaxy_root`                   | `/srv/galaxy`                                                    | This is the root of the Galaxy deployment.
 >    `galaxy_user`                   | `{name: "{{ galaxy_user_name }}", shell: /bin/bash}`             | The user that Galaxy will run as.
->    `galaxy_commit_id`              | `release_23.1`                                                   | The git reference to check out, which in this case is the branch for Galaxy Release 23.1
+>    `galaxy_commit_id`              | `release_23.0`                                                   | The git reference to check out, which in this case is the branch for Galaxy Release 23.0
 >    `galaxy_force_checkout`         | `true`                                                           | If we make any modifications to the Galaxy codebase, they will be removed. This way we know we're getting an unmodified Galaxy and no one has made any unexpected changes to the codebase.
 >    `miniconda_prefix`              | {% raw %}`"{{ galaxy_tool_dependency_dir }}/_conda"`{% endraw %} | We will manually install conda as well. Normally Galaxy will attempt to auto-install this, but since we will set up a production-ready instance with multiple handlers, there is the chance that they can become deadlocked.
 >    `miniconda_version`             | `23.9`                                                           | Install a specific miniconda version, the latest one at the time of writing that was tested and working.
@@ -920,7 +920,7 @@ The configuration is quite simple thanks to the many sensible defaults that are 
 >    +galaxy_layout: root-dir
 >    +galaxy_root: /srv/galaxy
 >    +galaxy_user: {name: "{{ galaxy_user_name }}", shell: /bin/bash}
->    +galaxy_commit_id: release_23.1
+>    +galaxy_commit_id: release_23.0
 >    +galaxy_force_checkout: true
 >    +miniconda_prefix: "{{ galaxy_tool_dependency_dir }}/_conda"
 >    +miniconda_version: 23.9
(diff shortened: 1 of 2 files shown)


topics/dev/tutorials/interactive-tools/tutorial.md
```diff
@@ -554,10 +554,9 @@ Let's check this integration on your machine. You can use a VM if you prefer not
 > cd ~/GxIT/galaxy/config
 > # Create custom config files
 > cat galaxy.yml.interactivetools > galaxy.yml
-> cat job_conf.xml.interactivetools > job_conf.xml
 > cat tool_conf.xml.sample > tool_conf.xml
 > ```
-> In `galaxy.yml`, add the `galaxy_infrastructure_url` parameter to the galaxy section:
+> In `galaxy.yml`, ensure that the `galaxy_infrastructure_url` parameter os present under the galaxy section:
```
bgruening, line 559: is instead of os

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -2082,7 +2082,7 @@ For this, we will use NGINX (pronounced "engine X" /ˌɛndʒɪnˈɛks/ EN-jin-EK
 >
 >    > <details-title>Running this tutorial <i>without</i> SSL</details-title>
 >    >
->    > In your `galaxy.j2` in the above step, you should change the `listen` parameter:
+>    > Ignore the https redirection by omitting `templates/nginx/redirect-ssl.j2`. In your `galaxy.j2` in the above step, you should change the `listen` parameter:
```
hexylena, line 2085:
```suggestion
>    > In your `galaxy.j2` in the above step, you should change the `listen` parameter:
```

This is covered in the other "Running this tutorial without SSL details box.

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -894,10 +894,10 @@ The configuration is quite simple thanks to the many sensible defaults that are
 >    `galaxy_layout`                 | `root-dir`                                                       | This enables the `galaxy_root` Galaxy deployment layout: all of the code, configuration, tools, and mutable-data (like caches, location files, etc.) folders will live by default beneath `galaxy_root`. User data is stored under `file_path`, a variable we will set later.
 >    `galaxy_root`                   | `/srv/galaxy`                                                    | This is the root of the Galaxy deployment.
 >    `galaxy_user`                   | `{name: "{{ galaxy_user_name }}", shell: /bin/bash}`             | The user that Galaxy will run as.
->    `galaxy_commit_id`              | `release_23.0`                                                   | The git reference to check out, which in this case is the branch for Galaxy Release 23.0
+>    `galaxy_commit_id`              | `release_23.1`                                                   | The git reference to check out, which in this case is the branch for Galaxy Release 23.1
```
hexylena, line 897:
```suggestion
>    `galaxy_commit_id`              | `release_23.0`                                                   | The git reference to check out, which in this case is the branch for Galaxy Release 23.0
```

The entire GAT has been tested with Release 23.0 only. Our goal is to be sure these tutorials keep working for admins, and given that we skipped 22.05 due to infrastructure changes, I feel we need to be a bit more cautious here.

xref https://github.com/galaxyproject/training-material/issues/4322

topics/admin/tutorials/ansible-galaxy/tutorial.md
```diff
@@ -275,7 +275,7 @@ We have codified all of the dependencies you will need into a YAML file that `an
 >    @@ -0,0 +1,13 @@
 >    +# Galaxy, Postgres, Nginx
 >    +- src: galaxyproject.galaxy
->    +  version: 0.10.14
+>    +  version: 0.10.15
```
hexylena, line 278: Is there a necessary update in 0.10.15 that fixes your fatal error?

topics/admin/tutorials/ansible-galaxy/tutorial.md
tuncK: It does not work with 23.0, either. Same conda error.

hexylena: Well that's a massive issue then, could you share in what way they're not resolving? Was there a bugfix in 23.1 that we should backport?

topics/admin/tutorials/ansible-galaxy/tutorial.md
tuncK: OK (as part of the upcoming commit).

tuncK: OK (part of the upcoming commit)

topics/admin/tutorials/ansible-galaxy/tutorial.md
tuncK: Regardless, reverted back to 23.0

topics/admin/tutorials/ansible-galaxy/tutorial.md
tuncK: I saw the second box, but it does not fix my problem. If I follow both boxes, I get a 502 error.

topics/admin/tutorials/ansible-galaxy/tutorial.md
hexylena: Can you please share the conda error? I think that's the single most important part of this error is what's going wrong with conda. (Were there discussions in matrix I missed or so?)

tuncK: > Well that's a large issue then, could you share in what way they're not resolving? Was there a bugfix in 23.1 that we should backport?

Sorry, I should have been less concise: The conda resolution fails for both 23.0 and 23.1 for me. And that is because of a conflict between conda versions and the latest python that it fetches. Both are defined by the tutorial, and both 23.0 and 23.1 can be gotten to work by specifying a more recent conda version as I did.

tuncK: > Can you please share the conda error? I think that's the single most important part of this error is what's going wrong with conda. (Were there discussions in matrix I missed or so?)

```
      - conda=4.12.0 -> python[version='>=3.10,<3.11.0a0|>=3.7,<3.8.0a0|>=3.9,<3.10.0a0|>=3.8,<3.9.0a0']
    Your python: python=3.11
```

PS: I see your messages with a lag.

hexylena: Ahhh it's a py3.11 issue. Ok, then the solution is maybe instead to pin the specific python version required? @natefoo's miniconda role enables that: https://github.com/galaxyproject/ansible-miniconda/blob/main/defaults/main.yml#L17

that said i'm surprised it fails, since the default galaxy conda environment is supposed to be created with ancient 3.6 https://github.com/galaxyproject/ansible-miniconda/blob/main/defaults/main.yml#L43

topics/admin/tutorials/ansible-galaxy/tutorial.md
hexylena: That's quite odd! 502 indicates an error reaching the Galaxy backend, if you're seeing that it usually isn't related to nginx, nor ssl configuration, but instead because galaxy isn't started properly (maybe the conda issue?)

The first box includes precisely this change, removing the redirect from deployment. If you followed the tutorial as written, then you saw the first box which included this change:

image

that would prevent the redirect file from ever being sent to the server where galaxy is being set up. The template will still exist in your ansible, but, it isn't included in the configuration.

topics/admin/tutorials/ansible-galaxy/tutorial.md
tuncK: Yes, it seems so. Though, what is the advantage of downgrading python rather than upgrading conda?

hexylena: So I can definitely agree with upgrading conda! I don't *want* to downgrade python, it's just an unexpected result given the error message.

From your error message, for conda 4.12.0,  it's erroring on Python 3.11. Based on the role Python 3.6 _should_ have been installed, but for some reason I imagine it's using system python? So this is I guess down to your specific operating system, that python 3.11 is installed by default which doesn't support conda 4. 

So, yes, we should update conda. Hopefully that fixes it all.

hexylena (approved):
Thanks for the conda vesion bump @tuncK! With this, ansible-galaxy works for you, correct? Nothing else needed, just the conda bump?

mvdbeek:
Does this close https://github.com/galaxyproject/training-material/pull/4308/files then ?

hexylena:
Ah it closes half of it, the other half, yeah ok we should've merged that earlier. THanks @mvdbeek I'd forgotten about that.

tuncK:
> Thanks for the conda vesion bump @tuncK! With this, ansible-galaxy works for you, correct? Nothing else needed, just the conda bump?

Thanks a lot for merging, @hexylena. This indeed should solve the installation error, but not the 502. Though, let me re-visit this latter with a fresh installation, and file a separate issue or PR if still troubling.

PR #4379, opened 2023-10-05
Topic embeds, topics by tag

hexylena:
@nomadscientist what do you think? This something you could put in an iframe on the singlecell homepage?

It uses the GTN's existing layout engine of course so you get the Tags (do you hate those?) and it won't fit *as* nicely since we use different fonts/etc (but you could match our font.), but otherwise this might give you something that would let you keep https://usegalaxy-eu.github.io/index-singlecell.html updated

Not wide | wider 
-- | ---
Screen Shot 2023-10-05 at 14 26 28 | Screen Shot 2023-10-05 at 14 26 35


nomadscientist:
This is fucking amazing

nomadscientist:
I like the tags (and plan to organise the tags on the single cell tutorials soon anyway) so no, this is good, and particularly the narrow looks great

hexylena:
The workflow search

Screenshot 2023-10-05 at 16-25-31 Galaxy Training Cross UseGalaxy  Workflow Search

Also available as an embed

```
/workflows/embed.html?query=...
```

hexylena:
Ok, should be ready to go. Would appreciate that whoever merges this tests it briefly (do topic pages load right? by-tag topic pages like covid?)

hexylena:
@nomadscientist you wanna hit merge? I think I've tested enough locally

nomadscientist (approved):

PR #4382, opened 2023-10-06
News Update: allow sub-selecting news to send to different matrix channels

nomadscientist:
I have no idea how to test this

Goal: any tutorials/slide decks put onto the single cell page, notifies the user matrix channel.

I removed the 'news' pull because we don't have #singlecell tagged on news so it doesn't work


hexylena (approved):
Thanks for starting this @nomadscientist I rewrote it a bit and I think it should be good to go now :) We'll have to figure out the news tagging at a later date

PR #4387, opened 2023-10-06
Trying to add the bar

nomadscientist:
There used to be a button/bar around the 'user requests' section, but I think I deleted it...


nomadscientist:
IT WORKED! I AM A GENIUS

mtekman (approved):

PR #4390, opened 2023-10-07
Single cell page - Fix sc tags, change Gitter links

nomadscientist:


mtekman:
It's a lot of commits, I'm going to squash them and clean it up a bit

mtekman:
Will close and @nomadscientist can re-open this in a new PR

hexylena:
> It's a lot of commits, I'm going to squash them and clean it up a bit

The GTN has no requirement for clean history. As long as you aren't adding + removing images (which thus will be preserved forever) then we really don't mind many small commits.

PR #4398, opened 2023-10-09
Add workflowhub to cross .* workflow list page

hexylena:
And update the metadata file weekly.

Not sure if you have feelings about this @shiltemann, half a megabyte file, do we keep it as part of the build process, or bake it into the GTN. It takes about 1.25 minutes to collect the data. It would be fresher, but, it's already a slow deploy process.

### DIFF
--- a/bin/workflows-fetch.rb
+++ b/bin/workflows-fetch.rb
@@ -4,16 +4,22 @@ require 'net/http'
 require 'uri'
 require 'yaml'
 
-# Get the list of workflows
-def fetch_workflows(server)
-  uri = URI.parse("#{server}/api/workflows/")
+def request(url)
+  uri = URI.parse(url)
   request = Net::HTTP::Get.new(uri)
+  request['Accept'] = 'application/json'
   req_options = {
     use_ssl: uri.scheme == 'https',
   }
   response = Net::HTTP.start(uri.hostname, uri.port, req_options) do |http|
     http.request(request)
   end
+  response
+end
+
+# Get the list of workflows
+def fetch_workflows(server)
+  response = request("#{server}/api/workflows/")
 
   JSON.parse(response.body).map do |w|
     w['server'] = server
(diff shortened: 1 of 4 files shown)


bin/workflows-fetch.rb
```diff
@@ -4,34 +4,88 @@
...
+  data = JSON.parse(response.body)
+  if !data['links']['next'].nil?
+    puts "ERROR: Cannot yet handle multiple pages"
+    exit 42
+  end
+  puts "INFO: Fetching #{data['data'].length} workflows from WorkflowHub"
+  data['data'].map.with_index { |w, i|
+    # {"id"=>"14", "type"=>"workflows", "attributes"=>{"title"=>"Cheminformatics - Docking"}, "links"=>{"self"=>"/workflows/14"}}
+    wf_info = JSON.parse(request("https://workflowhub.eu#{w['links']['self']}").body)
+    creator0 = wf_info['data']['attributes']['creators'][0]
```
wm75, line 41: I think `<div class="rli-project-list hidden-xs">` would be good to parse, too, since it has the Teams information, e.g. iwc.

hexylena:
> Could also be good to ask workflowhub people for an "embed" view or whatever you want to call it, which could be non-paginated and without the surrounding header/sidebar stuff?

yeah that could be possible. Here we're mixing the results of .* + wfh together so folks can embed a listing of "all galaxy workflows", potentially filtered on a specific search term.

bin/workflows-fetch.rb
hexylena: I don't have easy access to that in the endpoint I'm working with currently. We do see this, but that requires additional HTTP requests to fetch. So currently we look only at creators + other_creators which are mostly filled out.

```
      "creators": [],
      "other_creators": "Wolfgang Maier",
...
    "relationships": {
      "creators": {
        "data": []
      },
      "submitter": {
        "data": [
          {
            "id": "112",
            "type": "people"
          }
        ]
      },
      "people": {
        "data": [
          {
            "id": "112",
            "type": "people"
          }
        ]
      },
      "projects": {
        "data": [
          {
            "id": "33",
            "type": "projects"
          }
        ]
      },
```

hexylena: It's wildly denormalised, everything takes 10 requests for some reason. But yes! that's a good suggestion, that's only 1 then.

Shall I just comma-separate that with whatever is in `creators` and `other_creators` and the `relationships.projects.data[]` then? There are so many fields for attribution here that it seems silly somehow to stuff so many names in the page (especially compared to Galaxy workflows with only a single owner)

hexylena: Thanks, good suggestions, will implement.

hexylena: Implemented.

nomadscientist (approved):

PR #4401, opened 2023-10-09
fix sending empty news

hexylena:
@nomadscientist merge please :)


hexylena:
Or anyone else reading this, if it gets merged in the next 10 minutes we won't get an empty news post again

bebatut (approved):

PR #4403, opened 2023-10-09
remove single cell tag from tutorials

nomadscientist:
I've learned it's bad practice to label a tag the same as the actual page, so let's remove it! Woo!

### DIFF
--- a/_plugins/jekyll-jsonld.rb
+++ b/_plugins/jekyll-jsonld.rb
@@ -185,7 +185,7 @@ module Jekyll
         url: "#{site['url']}#{site['baseurl']}#{page['url']}",
         name: page['title'],
         headline: page.excerpt[0..100].gsub(/\n/, ' '), # TODO: remove html tags.
-        keywords: page.fetch('tags', []),
+        keywords: page['tags'] || [],
         description: page.excerpt[0..100].gsub(/\n/, ' '), # TODO: remove html tags
         articleBody: page.content, # TODO: remove html tags
         datePublished: page.date,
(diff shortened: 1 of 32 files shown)


pavanvidem:
Also, remove from the Spanish version? There are some linting issues, mainly due to incorrect levels of headings. Otherwise, looks good.

hexylena (commented):
Made more sense when they were in transcriptomics, now that they're in their own category this looks like an improvement!

Would you mind fixing some of the heading level issues while you're here?

nomadscientist:
> Also, remove from the Spanish version? There are some linting issues, mainly due to incorrect levels of headings. Otherwise, looks good.

Only because we aren't upkeeping them (still waiting on that grant to come through :-P)
But I'm not killing the pages so Spanscriptomics still lives...

But also Jurgi will create a better automated system (hopefully) so all hope is not lost for the Spanish translations

nomadscientist:
> Made more sense when they were in transcriptomics, now that they're in their own category this looks like an improvement!
> 
> Would you mind fixing some of the heading level issues while you're here?

Done!

nomadscientist:
@hexylena I can't figure out the latest linting error...

hexylena (approved):

hexylena:
>  I can't figure out the latest linting error...

ähm, you find the best failure modes, ¿thank you? for flagging them for us to fix lol

hexylena:
Ok it's due to having `tags:` rather than `tags: []` but that should be caught earlier in the linting process, not this late. But our schema validation library apparently doesn't check types properly, so `nil` was allowed for a `sequence`. So I guess we'll be more careful about how we handle tags access and that'll fix it.

hexylena (approved):

nomadscientist:
I am a magic beast at finding new ways to screw up
Thank you fantastic GTN overlord!

PR #4408, opened 2023-10-11
Separate out funders, organisations from GTN Contributors

hexylena:
Affiliations are now tracked (if users add them)
image

and we add a short name for them since everyone puts 60 words in their grant titles:
image

The homepage logo soup is now auto-generated and updated from org/funder list

image


hexylena:
todo: add affiliation tracking, show on profile.

hexylena:
xref https://github.com/galaxyproject/training-material/issues/4420

@nomadscientist this will in theory let you have a 'community' group that people mark themselves as affiliated with

If you had such a 'homepage' for a specific group in the GTN, do you have any thoughts what you'd like it to look like?

Right now it will look like a standard contributor page with the addition of 'affiliated users' being listed (unsorted)

Would you also want to be updating this page? Would you want custom markdown blocks where you can write about the org? Or is that just more work and you'd like it to point to the hub so you can de-duplicate? Not sure what's best for you

nomadscientist:
Interesting!
Ok, so I like separating out affiliation from funding, I think that's important. Too much free labour, you know?

I personally treat the single cell training page as the page for the SC-CoP. So I wouldn't want yet another page to a) keep updated and b) potentially trip up users overwhelmed by so many places to go. I was slightly reticent about learning pathways because of that (I sort of saw each training page as a learning pathway) but then saw it useful for cherry-picking courses in a time-limited fashion (i.e. we have 1 week to run a course on X and way too many tutorials, let's pick some for our pathway). So it has a clear use

I don't see a clear use for a group page in the GTN on top of the topics themselves (unless some topics are really just that massive? Do they need a divorce like the single cell / transcriptomics)?

It would be nice to have anybody who is in the SC-CoP have that on their contributors page. But, you make a fair point - the Hub currently has those groups on it, and furthermore there's no active list of those groups. But I guess we could manually put it in like we do affiliations, couldn't we, and then it's up to individual users how much they put there? And it could just point to the hub, like you're saying? That seems like a good idea.

hexylena:
> I don't see a clear use for a group page in the GTN on top of the topics themselves (unless some topics are really just that massive? Do they need a divorce like the single cell / transcriptomics)?

that's a great result to have, no use for group pages on the GTN, ok, one less thing to do :)

> It would be nice to have anybody who is in the SC-CoP have that on their contributors page. But, you make a fair point - the Hub currently has those groups on it, and furthermore there's no active list of those groups. But I guess we could manually put it in like we do affiliations, couldn't we, and then it's up to individual users how much they put there? And it could just point to the hub, like you're saying? That seems like a good idea.

yeah, so, manual for now and up to individuals how much they add. I'll see about adding a link to the hub when we create a 'single cell CoP organisation' on the gtn side (maybe let's revisit that later)

hexylena:
ok, think this is ready to ship before it gets any more out of date with contributors. (i've rebased on main)

shiltemann (approved):

Issue #2799, opened 2021-09-30
Suggestion for Annotating clusters section scrna-seq-basic-pipeline


swbioinf:
Hi,

Firstly thanks to @nomadscientist and others for an excellend scRNAseq processing tutorial in scrna-seq-basic-pipeline. Planning to use it for a workshop.

Had one suggestion around the annotating clusters section. Currently it suggests copying a metatdata column, renaming it, and then adding celltype names. I think this could be simplified by copying the column, using the 'keep original' option in ManipulateAnndata (1.8.1+1+galaxy1).  And also adding a note that the cell type order could be different if you have different clustering.

Afraid I don't know enough about the tutorial structure and building e.t.c yet to put this in as a proper pull request. I've put a suggestion for the 'Annotating clusters' section of the markdown (https://github.com/galaxyproject/training-material/blob/main/topics/transcriptomics/tutorials/scrna-seq-basic-pipeline/tutorial.md) below though. In case someone happens to be working on that one?

Low priority.

Thanks!

`
### Annotating Clusters

> ### {% icon hands_on %} Hands-on: Annotating clusters
>
> 1. {% tool [AnnData Operations](toolshed.g2.bx.psu.edu/repos/ebi-gxa/anndata_ops/anndata_ops/1.6.0+galaxy1) %} {% icon tool %} with the following parameters:
>    - {% icon param-file %} *"Input object in hdf5 AnnData format"*: `Final object`
>    - *"Change field names in AnnData observations"*
>    - {% icon galaxy-wf-new %} *"Insert Change field names in AnnData observations"*
>    - *"1: Change field names in AnnData observations"*
>    - *"Original name"*: `louvain`
>    - *"New name"*: `cell_type`
>    - *"Keep original"*: {% icon history-share %} `Yes`
>    - This makes a copy of the louvain column, which you can change without altering the original
>
> 2. {% tool [Manipulate AnnData](toolshed.g2.bx.psu.edu/repos/iuc/anndata_manipulate/anndata_manipulate/0.7.5+galaxy0) %} with the following parameters:
>    - {% icon param-file %} *"Annotated data matrix"*: (output of **AnnData Operations** {% icon tool %})
>    - *"Function to manipulate the object"*: `Rename categories of annotation`
>    - *"Key for observations or variables annotation"*: `cell_type`
[...]

PR #4424, opened 2023-10-13
Faq common sc text

nomadscientist:
I'm making the tutorials and pages more consistent
With turning common text into FAQ and fixing some formatting issues

### DIFF
--- a/topics/single-cell/index.md
+++ b/topics/single-cell/index.md
@@ -13,6 +13,6 @@ Check out workflows shared by users like you!
 
 If you want to help us behind the scenes, from testing workflows and tutorials to building tools, join our Galaxy Single Cell Community of Practice!
 
- - <i class="fa fa-hand-o-right" aria-hidden="true"></i> [Community of Practice](https://galaxyproject.org/projects/singlecell/)
- - <i class="fa fa-comments-o" aria-hidden="true"></i> [Matrix Chat Forum](https://matrix.to/#/#usegalaxy-eu_single-cell-workflows:gitter.im)
+ - <i class="fa fa-hand-o-right" aria-hidden="true"> </i> [Community of Practice](https://galaxyproject.org/projects/singlecell/)
+ - <i class="fa fa-comments-o" aria-hidden="true"> </i> [Matrix Chat Forum](https://matrix.to/#/#usegalaxy-eu_single-cell-workflows:gitter.im)
  - <i class="fa fa-envelope" aria-hidden="true"> </i> [Mailing List](https://lists.galaxyproject.org/lists/single-cell-cop.lists.galaxyproject.org/)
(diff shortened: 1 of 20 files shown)


topics/single-cell/index.md
```diff
@@ -13,6 +13,6 @@ Check out workflows shared by users like you!
 
 If you want to help us behind the scenes, from testing workflows and tutorials to building tools, join our Galaxy Single Cell Community of Practice!
 
- - <i class="fa fa-hand-o-right" aria-hidden="true"></i> [Community of Practice](https://galaxyproject.org/projects/singlecell/)
- - <i class="fa fa-comments-o" aria-hidden="true"></i> [Matrix Chat Forum](https://matrix.to/#/#usegalaxy-eu_single-cell-workflows:gitter.im)
+ - <i class="fa fa-hand-o-right" aria-hidden="true"> </i> [Community of Practice](https://galaxyproject.org/projects/singlecell/)
+ - <i class="fa fa-comments-o" aria-hidden="true"> </i> [Matrix Chat Forum](https://matrix.to/#/#usegalaxy-eu_single-cell-workflows:gitter.im)
  - <i class="fa fa-envelope" aria-hidden="true"> </i> [Mailing List](https://lists.galaxyproject.org/lists/single-cell-cop.lists.galaxyproject.org/)
```
hexylena, line 18: `{% icon .. %}` should function here, let's add support for whatever icons you need to `_config.yml` (just give them nice human names) to make it a bit more readable for you and others?

```suggestion
 - <i class="fa fa-hand-o-right" aria-hidden="true"> </i> [Community of Practice](https://galaxyproject.org/projects/singlecell/)
 - {% icon feedback %} [Matrix Chat Forum](https://matrix.to/#/#usegalaxy-eu_single-cell-workflows:gitter.im)
 - <i class="fa fa-envelope" aria-hidden="true"> </i> [Mailing List](https://lists.galaxyproject.org/lists/single-cell-cop.lists.galaxyproject.org/)
```

nomadscientist:
Also addressed this issue: https://github.com/galaxyproject/training-material/issues/2799

topics/single-cell/index.md
nomadscientist: like this?   point-right: fa fa-hand-o-right

hexylena: yep! fas / far are 'solid' and 'regular', whichever one is fitting better + free (if you check the fontawesome website/search)

nomadscientist:
removed the FPE fix to another branch since I have to deal with workflow testing...

nomadscientist:
Yes I think that's an awesome idea! Idk how to make the subsections without examples yet but I think that's a great call
It goes hand-in-hand with us needing to change the help forum to switch from scrna as the only option to single cell as well

pavanvidem:
Then we should also merge all the markdowns into 3 files to fit the sections if that's possible.

PR #4432, opened 2023-10-16
add zenodo api url linting, fix broken zenodo links

hexylena:
since many of the `zenodo.org/api/` URLs are now broken/not resolving, at least we can prevent it going forward


shiltemann (approved):

shiltemann:
thanks!

PR #4443, opened 2023-10-20
small fixes to Reference-based RNA-Seq data analysis tutorial

hrhotz:
several small fixes as discussed yesterday with Lucille


lldelisle (approved):
Thanks. If you are motivated enough. There are still 2 places where [Filter](Filter1) could be replaced by [Filter data on any column using simple expressions](Filter1). This way it would be more homogeneous.

lldelisle (approved):

hexylena:
Ooof. Didn't consider how many places would need updates to tool names.

Maybe we should automate this by keeping a cache of `tool_id: tool_name` so you can just provide the tool ID.

Thanks for the fixes!

hexylena:
@lldelisle if tests pass, you're always welcome to merge it directly! :)

PR #4445, opened 2023-10-20
Use subworkflows in Ref-based Transcriptomics GTN

lldelisle:
Update workflow that follows the tutorial to have steps in the same order as in the tutorial


hexylena (approved):

lldelisle:
There is no workflow test?

mvdbeek:
Just in case, https://github.com/mvdbeek/iwc-consumer/blob/main/.github/workflows/workflow_test.yml is an example of how you can use the IWC test workflow in an external repo.

mvdbeek:
The workflows should have minimal test data, so it shouldn't take that long ? For most IWC workflows it's on the other of 10 to 15 minutes. But you could of course use a slash command or something like that when the workflow is ready.

PR #4446, opened 2023-10-20
NEW LESSON Making clinical datasets FAIR

kkamieniecka:
Making clinical datasets FAIR: discusses how to make clinical datasets FAIR and why this is important
PR update @hexylena I think this one is fixed now, thanks


bin/lint.rb
```diff
@@ -684,21 +684,6 @@ def self.check_bolded_heading(contents)
...
-        idx: idx,
-        match_start: selected.begin(1),
-        match_end: selected.end(1) + 1,
-        replacement: nil,
-        message: 'The Zenodo.org/api URLs are not stable, you must use a URL of the format zenodo.org/record/..., apologies we cannot fix automatically.',
-        code: 'GTN:032'
-      )
-    end
-  end
-
```
hexylena, line 701: really sorry but it looks like something has gone weirdly in the merge here; we just added these lines recently in a PR, but here they're being removed and I suspect it isn't just a "github display the PR weirdly" issue, but part of the commits. Is there any chance you could extract just the files you want from this PR?

kkamieniecka: Hi @hexylena I was just writing to you when I spot this comment. Yes I think something went wrong as before the disappearing branches (on our side) PR was linting correctly

PR #4454, opened 2023-10-23
 NEW LESSON Making clinical datasets FAIR

kkamieniecka:
Making clinical datasets FAIR: discusses how to make clinical datasets FAIR and why this is important; PR fix


mtekman (commented):
I can't comment on the config or linting changes, but why are the Zenodo links changed from

`https://zenodo.org/record/<blah>/files/` to `https://zenodo.org/api/files/<new_id>` ? 

Are these links generated directly because of a new version change from Zenodo, or are they the same files and it's simply better to access the files directly via the API?

kkamieniecka:
@hexylena can you advise me how to fix this lint? I feel like it is keep merging in a weird way

hexylena:
@mtekman no, the /API route doesn't work, there are some commits on this branch that revert a previous commit to move towards the /records route which does work

> Are these links generated directly because of a new version change from Zenodo, or are they the same files and it's simply better to access the files directly via the API?

@kkamieniecka the linting errors come from all of the 'unexpected' files added, these trigger linting on those files, and those files have existing linting errors. We really need to remove the unexpected file changes (bin/lint.rb, etc.) anything not intended to be here. (you don't need to open a new PR, let's just use this one,) but we need to rewrite the commits to remove those unexpected changes. I'll see what I can do

hexylena:
So for these, since I can't push commits to your branch, and authorship information is already lost, I would recommend the following procedure.

```
git checkout main
git checkout -b DASH-1211-b
git reset DASH-1211 -- topics/fair/
git checkout -- topics/fair/
```

and this should restore *just* the topics/fair/ changes without all of the other tutorials touched. Then you can 'edit' this PR to use that branch as the source.

PR #4455, opened 2023-10-23
fix broken PR 

kkamieniecka:
NEW LESSON Making clinical datasets FAIR #4454
should restore just the topics/fair/ changes without all of the other tutorials touched. Then you can 'edit' this PR to use that branch as the source thanks, @hexylena


hexylena:
Oh this looks much better!! thanks so much @kkamieniecka

hexylena (approved):

PR #4468, opened 2023-11-01
Add tool version to tool buttons

hexylena:
cc @nomadscientist @lldelisle  we've discussed this in the past but it continues to be an issue so time to finally do it. It's a bit more verbose, but tutorials are already verbose, and reproducibility of using the correct tool version is more important.

image


nomadscientist:
This is such a good idea

shiltemann (approved):

lldelisle:
@hrhotz following DESeq2 issue...

Issue #4313, opened 2023-08-09
Error 137 while installing requirements


kostrykin:
The [GTN tutorial on viewing the site locally](https://training.galaxyproject.org/training-material/topics/contributing/tutorials/running-jekyll/tutorial.html) basically says that one should run
1. `make create-env`
2. `make install`

after installing Conda.

I have a fresh local clone of the repository. When running `make create-env` (step 1), an out of memory error occurs (error 137):

> (base) user@hostname:~/Documents/galaxy-training-material$ make create-env
> find: ‘_site/training-material’: No such file or directory
> find: ‘_site/training-material/*/*/slides/*’: No such file or directory
> find: ‘_site/training-material’: No such file or directory
> if ~/.anaconda3/bin/conda env list | grep '^galaxy_training_material'; then \
>     ~/.anaconda3/bin/conda env update -f environment.yml; \
> else \
>     ~/.anaconda3/bin/conda env create -f environment.yml; \
> fi
> Collecting package metadata (repodata.json): done
> Solving environment: \ bash: line 4: 3332886 Killed                  ~/.anaconda3/bin/conda env create -f environment.yml
> make: *** [Makefile:43: create-env] Error 137

The same happens when I try running `make install` (step 2) after creating the environment manually using `mamba env create -f environment.yml`:

> (galaxy_training_material) user@hostname:~/Documents/galaxy-training-material$ make install
> find: ‘_site/training-material’: No such file or directory
> find: ‘_site/training-material/*/*/slides/*’: No such file or directory
> find: ‘_site/training-material’: No such file or directory
> if ~/.anaconda3/envs/galaxy_training_material/bin/conda env list | grep '^galaxy_training_material'; then \
>     ~/.anaconda3/envs/galaxy_training_material/bin/conda env update -f environment.yml; \
> else \
>     ~/.anaconda3/envs/galaxy_training_material/bin/conda env create -f environment.yml; \
> fi
> Retrieving notices: ...working... done
> Collecting package metadata (repodata.json): done
> Solving environment: \ bash: line 4: 3341561 Killed                  ~/.anaconda3/envs/galaxy_training_material/bin/conda env create -f environment.yml

So how to install the requirements?

kostrykin:
So in case someone else stumbles into this issue. Running `make install --dry-run` gave me the necessary hints to install the requirements manually:

1. `gem update --no-document --system`
[...]

PR #4511, opened 2023-11-13
fixes Makefile, prefers mamba over conda if available

kostrykin:
The [GTN tutorial on viewing the site locally](https://training.galaxyproject.org/training-material/topics/contributing/tutorials/running-jekyll/tutorial.html) basically says that one should run
1. `make create-env`
2. `make install`

after installing Conda. I have a fresh local clone of the repository. Both steps yield an out of memory error on my system (see #4313). When using mamba instead of conda, everything works.

This PR proposes to use mamba in Makefile if it is available, and conda otherwise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,12 @@ else
 	ENV_FILE=environment.yml
 endif
 
-CONDA=$(shell which conda)
+CONDA=$(shell which mamba)
 ifeq ($(CONDA),)
-	CONDA=${HOME}/miniconda3/bin/conda
+    CONDA=$(shell which conda)
+    ifeq ($(CONDA),)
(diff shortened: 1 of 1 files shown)


hexylena (approved):

hexylena:
Thanks @kostrykin

hexylena:
The other thing to note is that mamba has been merged into conda, one just needs to enable the resolver. I guess at some point it will also become the default resolver

abretaud:
(The mamba resolver is the default now in the latest release)

kostrykin:
> (The mamba resolver is the default now in the latest release)

Doing a `conda update conda` still gives me conda version 22.9, but the mamba resolver is default only since 23.10.

PR #4515, opened 2023-11-14
Implement nicer categories for toolshed install instructions

### LINKED ISSUE #4513: Gives 'better' tool_panel_section_label in the tool YAML
First of all, this is super cool to have the `Galaxy Administrators: Install the missing tools` section :heart: .
I think it would be even better if we could improve the tool_panel_section_label.
My idea would be to use https://github.com/usegalaxy-eu/usegalaxy-eu-tools and/or https://github.com/galaxyproject/usegalaxy-tools to attribute the good section label.
@hexylena proposed to:
```
generate a list of ids → sections, we could pull from the big 3 and use an algorithm like "if 2 agree, use that."
```

hexylena:
cc @lldelisle
fixes #4513 

depends on:

- [x] https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/648
- [x] https://github.com/galaxyproject/usegalaxy-tools/pull/655
- [ ] https://github.com/usegalaxy-au/usegalaxy-au-tools/pull/988

Demo using data from those three PRs, for the atac-seq tutorial:

image

quite a bit nicer than all going to "Epigenetics"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,6 +235,7 @@ annotate: ## annotate the tutorials with usable Galaxy instances
 	wget https://github.com/hexylena/toolshed-version-database/raw/main/guid-rev.json -O metadata/toolshed-revisions.json && \
 	python bin/supported-fetch.py
 	bin/workflows-fetch.rb
+	bin/fetch-categories.rb
 .PHONY: annotate
 
 rebuild-search-index: ## Rebuild search index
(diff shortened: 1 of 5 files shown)


lldelisle:
You are a genius! I love it.

shiltemann (approved):

PR #4532, opened 2023-11-19
limits deploy workflow to `galaxyproject` repo

kostrykin:
Forks of this repo regularly run deployment workflows to GTN. Such workflows usually fail, since deployment is usually not configured in the forked repo. This regularly produces notifications and emails of failed workflows. To overcome this issue, I suggest limiting the deployment workflow to the original GTN repo (i.e. this repo).


hexylena (approved):
Thanks @kostrykin, we have that in a few other workflows, but it definitely should be here. Cheers.

PR #4581, opened 2023-12-11
Update account_update_preference.md

jennaj:
Add in newer preferences.

### DIFF
--- a/faqs/galaxy/account_update_preference.md
+++ b/faqs/galaxy/account_update_preference.md
@@ -8,7 +8,7 @@ contributors: [jennaj, bernandez]
 
 
 1. Log in to Galaxy
-2. Navigate to **User -> Preferences** on the top menu bar
+2. Navigate to [**User -> Preferences**](https://my.galaxy.training/?path=/user) on the top menu bar
 3. Here you can update various preferences, such as:
    - {% icon pref-info %} Manage Information (change your registered email addresses or public name)
    - {% icon pref-password %} Change Password (change your login credentials)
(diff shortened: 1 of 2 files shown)


jennaj:
Whoever reviews -- please help me to check the new icons I added in to match the UI (palette and list). Guessing about what font-awesome translates to for these.

hexylena:
To check the icon names you can use the browser's inspect tool to check, since we both use fontawesome

fa

fa-palette was the correct choice!

I'm separately making a PR to give better linting messages with an unknown icon.

hexylena (approved):

faqs/galaxy/account_update_preference.md
```diff
@@ -8,19 +8,19 @@ contributors: [jennaj, bernandez]
 
 
 1. Log in to Galaxy
-2. Navigate to **User > Preferences** on the top menu bar.
+2. Navigate to **User -> Preferences** on the top menu bar
```
hexylena, line 11:
```suggestion
2. Navigate to [**User -> Preferences**](https://my.galaxy.training/?path=/user) on the top menu bar
```

We've setup the 'my.galaxy.training' service to allow redirecting to the user's preferred galaxy, it let's the enter the domain and then go directly to their galaxy, with galaxy-independent URLs (like to /user). It doesn't work for e.g. workflows given their different IDs but it might be nice to use it in this one place, in case the user can't find it

hexylena (approved):
wish we didn't have to keep this in sync; thanks for the hard work @jennaj !

hexylena (approved):

PR #4586, opened 2023-12-12
Update reference_genomes_custom_genomes.md

jennaj:
Completely redone with cross-links. Easier to give a user one FAQ for this specific problem with full context re usual followup. Shows up nicely in forums.

### DIFF
--- a/faqs/galaxy/reference_genomes_custom_genomes.md
+++ b/faqs/galaxy/reference_genomes_custom_genomes.md
@@ -24,7 +24,7 @@ There are two options for reference genomes in Galaxy.
 There are five basic steps to use a **Custom Reference Genome**, plus one optional.
 1. Obtain a FASTA copy of the target genome. See tip 2.
 2. Upload the genome to Galaxy and to add it as a dataset in your history.
-3. [Clean up the format0({% link faqs/galaxy/datasets_working_with_fasta.md %}) with the tool **NormalizeFasta** using the options to wrap sequence lines at 80 bases and to trim the title line at the first whitespace.
+3. [Clean up the format]({% link faqs/galaxy/datasets_working_with_fasta.md %}) with the tool **NormalizeFasta** using the options to wrap sequence lines at 80 bases and to trim the title line at the first whitespace.
 4. Make sure the [chromosome identifiers]({% link faqs/galaxy/datasets_chromosome_identifiers.md %}) are a match for other inputs.
 5. Set a tool form's options to use a custom reference genome from the history and select the loaded genome FASTA.
 6. (Optional) Create a [custom genome build's database]{% link faqs/galaxy/analysis_add_custom_build.md %}) that you can [assign to datasets]({% link faqs/galaxy/datasets_change_dbkey.md %}).
--- a/faqs/galaxy/reference_genomes_custom_genomes.md
+++ b/faqs/galaxy/reference_genomes_custom_genomes.md
@@ -31,4 +31,4 @@ There are five basic steps to use a **Custom Reference Genome**, plus one option
 
 {% icon tip %} TIP 1: Avoid [assigning a native database]({% link faqs/galaxy/datasets_change_dbkey.md %}) to uploaded data unless you confirmed the data are based on the [same exact genome assembly]({% link faqs/galaxy/datasets_chromosome_identifiers.md %}) or you [adjusted the data to be a match]({% link topics/introduction/tutorials/data-manipulation-olympics/tutorial.html %}) **first**!
 
-{% icon tip %} TIP 2: When choosing your reference genome, consider [choosing your reference annotation]{% link faqs/galaxy/analysis_differential_expression_help %}) at the same time. Standardize the format of both as a preparation step. Put the files in a dedicated "reference data" history for easy resuse.
+{% icon tip %} TIP 2: When choosing your reference genome, consider [choosing your reference annotation]{% link faqs/galaxy/analysis_differential_expression_help.md %}) at the same time. Standardize the format of both as a preparation step. Put the files in a dedicated "reference data" history for easy resuse.
--- a/faqs/galaxy/reference_genomes_custom_genomes.md
+++ b/faqs/galaxy/reference_genomes_custom_genomes.md
@@ -7,12 +7,28 @@ contributors: [jennaj, Nurzhamalyrys]
 ---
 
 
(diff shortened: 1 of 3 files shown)


faqs/galaxy/reference_genomes_custom_genomes.md
```diff
@@ -7,12 +7,28 @@ contributors: [jennaj, Nurzhamalyrys]
...
+4. Make sure the [chromosome identifiers]({% link faqs/galaxy/datasets_chromosome_identifiers.md %}) are a match for other inputs.
+5. Set a tool form's options to use a custom reference genome from the history and select the loaded genome FASTA.
+6. (Optional) Create a [custom genome build's database]{% link faqs/galaxy/analysis_add_custom_build.md %}) that you can [assign to datasets]({% link faqs/galaxy/datasets_change_dbkey.md %}).
+
+{% icon tip %} TIP 1: Avoid [assigning a native database]({% link faqs/galaxy/datasets_change_dbkey.md %}) to uploaded data unless you confirmed the data are based on the [same exact genome assembly]({% link faqs/galaxy/datasets_chromosome_identifiers.md %}) or you [adjusted the data to be a match]({% link topics/introduction/tutorials/data-manipulation-olympics/tutorial.html %}) **first**!
+
+{% icon tip %} TIP 2: When choosing your reference genome, consider [choosing your reference annotation]{% link faqs/galaxy/analysis_differential_expression_help %}) at the same time. Standardize the format of both as a preparation step. Put the files in a dedicated "reference data" history for easy resuse.
```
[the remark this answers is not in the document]
hexylena:
```suggestion
{% icon tip %} TIP 2: When choosing your reference genome, consider [choosing your reference annotation]{% link faqs/galaxy/analysis_differential_expression_help.md %}) at the same time. Standardize the format of both as a preparation step. Put the files in a dedicated "reference data" history for easy resuse.
```

faqs/galaxy/reference_genomes_custom_genomes.md
```diff
@@ -7,12 +7,28 @@ contributors: [jennaj, Nurzhamalyrys]
...
 
-1. Obtain a FASTA copy of the target genome.
-2. Use FTP to upload the genome to Galaxy and load into a history as a dataset.
-3. Clean up the format with the tool **NormalizeFasta** using the options to wrap sequence lines at 80 bases and to trim the title line at the first whitespace.
-4. Make sure the chromosome identifiers are a match for other inputs.
-5. Set a tool form's options to use a custom reference genome from the history and select the loaded genome.
+There are five basic steps to use a **Custom Reference Genome**, plus one optional.
+1. Obtain a FASTA copy of the target genome. See tip 2.
+2. Upload the genome to Galaxy and to add it as a dataset in your history.
+3. [Clean up the format0({% link faqs/galaxy/datasets_working_with_fasta.md %}) with the tool **NormalizeFasta** using the options to wrap sequence lines at 80 bases and to trim the title line at the first whitespace.
```
hexylena, line 27:
```suggestion
3. [Clean up the format]({% link faqs/galaxy/datasets_working_with_fasta.md %}) with the tool **NormalizeFasta** using the options to wrap sequence lines at 80 bases and to trim the title line at the first whitespace.
```

hexylena:
@jennaj 

> Shows up nicely in forums.

I'm guessing it's not what you meant, but, we *could* make the GTN show up more nicely in discourse if it's helpful to you:

1. iframes are supported, we could allow-list the GTN domain
2. apparently we can support the 'oembed'  API to maybe have a richer preview that includes more of the FAQ for users, if it's helpful. 

https://meta.discourse.org/t/create-rich-link-previews-with-onebox/98088

faqs/galaxy/reference_genomes_custom_genomes.md
jennaj: Great, thanks!

jennaj:
That extra would be very good. Right now I mean the FAQ title, but pushing the content itself would be fabulous. Could you help to set this up? 

Related:
1. Tutorial titles also show up already, but an extra peek would be nicer.
2. Searches at the site render as a default `GalaxyTraining!` but including the search terms directly would be nicer. 
3. Title on a full topic or pathway view would also be nice to add in.

hexylena:
I think short term we can improve the title/preview, but all of those previews will be rather limited in length due to that field not supporting longer/richer texts.

long term, the oEmbed API doesn't look too bad to implement, but it'll take some more work to provide a nice experience. We'll have to decide how each type of page should look in the preview which is going to be some work, and maybe we need a dynamic page at least for the search results to render nicely which is more work, but, probably worth doing.

Doing some testing here: https://help.galaxyproject.org/t/gtn-link-preview-testing-ideas/11321

> Tutorial titles also show up already, but an extra peek would be nicer.

unfortunately we are limited by what discourse will display, I've removed the truncation because [some platforms seem to show longer descriptions](https://ogtester.com/blog/what-is-maximum-length-of-og-title-and-og-description).

here's the longest discourse will natively display, so
image

> Searches at the site render as a default GalaxyTraining! but including the search terms directly would be nicer.

Again unfortunately since it's a completely static site, we can't change the title field that gets rendered to include search terms. I've got a commit adding them dynamically when you're searching, it'll show up for the user, but that doesn't solve your problem at all.

> Title on a full topic or pathway view would also be nice to add in.

Oh, yes, we can improve this at least!

jennaj (commented):
updated to remove spaces

jennaj (commented):
Removed the export to file download/remote server options (on purpose). These have changed a ton and need a new FAQ that we can cross link into this one. The work is still in progress.. so maybe at next full release 24.N. I'll make a ticket as a reminder and assign myself.

jennaj:
Failing the check for the fa-link icon. Could we add it? Could also be added here https://github.com/galaxyproject/training-material/blob/main/faqs/galaxy/datasets_icons.md

Or, am I getting the format wrong? Not sure what the `::before` part means, is that the angle? Screenshot may help

Screen Shot 2023-12-14 at 2 02 13 PM

jennaj (commented):
Test link -- that seems to have been enough!

jennaj:
Let's also push this. I'd like to start using these for Q&A and see what else related remains are unclear to users out in the wild.

hexylena:
> Could we add it?

yes, any icons you want, just add to `_config.yml`, it looks like we have `galaxy-link` already in the config. Also visible [on this page](https://training.galaxyproject.org/training-material/topics/contributing/tutorials/create-new-tutorial-content/tutorial.html#icons).

I'm not sure how it works with just `{% icon link %}`, maybe our icon plugin is doing more than I thought, well, this works :)  

> Not sure what the ::before part means, is that the angle

it's a css selector, you just want `link`

PR #4588, opened 2023-12-13
Fix for broken link to Docker image missing Topic parameter in URL

hujambo-dunia:


hujambo-dunia:
Hi @hexylena it appears that the Topic Name got detached from an actual value and was outputting a blank string which created a 404-error when users tried to click the "Available on these Galaxies" >> "Docker image" link. Please check on your local that my suggested fix works. Open to edits.

hexylena (approved):
Thanks so much @hujambo-dunia for noticing this!

We'll probably deprecate the docker containers long term since the lack of updates to the base container mean they're not going to provide as nice of an experience as using one of the public servers, but for now great to have this corrected!!

hujambo-dunia:
> Thanks so much @hujambo-dunia for noticing this!
> 
> We'll probably deprecate the docker containers long term since the lack of updates to the base container mean they're not going to provide as nice of an experience as using one of the public servers, but for now great to have this corrected!!
> 
> (I've made an additional commit to use one of the already-included variables, since accesses to site.data incur a small performance hit on building, but that's a very esoteric bit of jekyll performance trivia.)

@hexylena Thanks for your correction - appreciated.

PR #4599, opened 2023-12-18
Blog post: user story

wee-snufkin:
I heard we don't have many insights from the users published, so encouraged by @nomadscientist I've created a wee blog post about how useful Galaxy was for me as a user. 

PS. Single cell blog post is on its way as well!


hexylena (approved):
:purple_heart:

PR #4627, opened 2024-01-09
Port more SCSS variables to CSS variables

ElectronicBlueberry:
Ports some of the missing SCSS variables from #3876 to CSS variables, allowing them to be easily adjusted by themes.


hexylena (approved):
lgtm.

though, for these colours we had not envisioned themes modifying them, in order to retain a coherent branding/colour-meaning throughout, but i guess that's a discussion to have in the context of someone actually wanting to change them, rather than now.

hexylena:
thanks @ElectronicBlueberry !

PR #4628, opened 2024-01-09
Update reference_genomes_custom_genomes.md

jennaj:
Fix two typos


jennaj (commented):
md 2 html !

hexylena:
The conversion of md to html on extensions is not required. In fact it *should* be triggering failures on our linting, it's a bug that it isn't.

you're right, the linting error is due to anton's new tutorial, and not related.

hexylena:
Ok, linting does work.

hexylena (approved):

PR #4631, opened 2024-01-10
Update bebatut info

bebatut:
Update with recent job info


hexylena (approved):

hexylena:
maybe we should have a 'former' associations role, since avans is no longer accurate for me

PR #4639, opened 2024-01-16
[CONTRIBUTORS] Support BlueSky contact option (& update my own contact details)

shiltemann:
also adds a full example user at top of file with all options


bgruening:
Thanks @shiltemann

PR #4641, opened 2024-01-16
Update microbiome pURLs

shiltemann:
the short URLs weren't hitting the jekyll redirects, so updating the targets here


natefoo (approved):

hexylena:
this is definitely a bug, we need to investigate it.

PR #4657, opened 2024-01-19
Changed question mark to under bar in test data file name to allow clone on Windows

martinwolst:
? is an allowable character in Linux filenames, but not in Windows. This prevents a clean clone on a Windows system so I have replaced it with _.


hexylena (approved):
Oh wow, didn't know that was an issue on windows. Ok, we'll add some linting for that in the future.

PR #4667, opened 2024-01-22
per topic rss feeds

hexylena:
it works. Per-topic RSS feeds announcing new (only, not updated) materials.

We could add more for e.g. updated materials per topic, all material changes, just let us know what's useful.

image


hexylena:
Fixed this up in support of #4890  which will need a new RSS feed as well.

shiltemann (approved):

PR #4675, opened 2024-01-24
update the control freec tutorial

khaled196:
The pull request is to fix the workflow and provide a workflow test. 

also fixed the Indentation in the markdown file


bgruening (approved):
Thanks @khaled196 looks good to me.

PR #4702, opened 2024-02-05
Automated dev terminal recordings

hexylena:
not sure if we should ship this as is

it's a nice improvement, since it gives the user somehing slightly more interactive to play with. but it also should be made more general. And perhaps it's a slight downgrade from xml syntax highlighting? Alternatively also should run the planemo command itself (using more of the auto-gat machinery, rather than just 'typing' the first file.)

asci


hexylena:
it's a nice feature, but, no time.

PR #4707, opened 2024-02-07
Created learning pathways and renamed the modules 

subinamehta:
I created a learning pathway and also renamed the modules so that they are easy to interpret.


hexylena (approved):
Looks great @subinamehta thanks so much!

PR #4726, opened 2024-02-14
Add short link to tpv shared db

nuwang:
Add a shorter and more memorable link to the TPV shared database, that can be transparently redirected if necessary.


hexylena (approved):
lgtm

Issue #4730, opened 2024-02-16
Odd compression on the R2.bz2 file for mrsa-illumina tutorial


jennaj:
Tutorial: https://training.galaxyproject.org/training-material/topics/assembly/tutorials/mrsa-illumina/tutorial.html

Zenodo: https://zenodo.org/records/4534098

Ok --- https://zenodo.org/record/4534098/files/DRR187559_1.fastqsanger.bz2
Fails "as truncated data" with FastQC --- https://zenodo.org/record/4534098/files/DRR187559_2.fastqsanger.bz2

The second file R2 in bz2 format is tossing errors with **FastQC** at both usegalaxy.org and usegalaxy.eu.

If that same file is uncompressed but otherwise unchanged, it works fine with FastQC.

If that file is left bz2 compressed, it fails just FastQC but works with the tutorial's workflow for the other steps.

Compression problem? Doesn't seem to be actually truncated. 

ORG https://usegalaxy.org/u/jen-galaxyproject/h/genome-assembly-of-mrsa-using-illumina-miseq-data-1
EU didn't run myself but it was reported here https://help.galaxyproject.org/t/fastqc-fails-in-mrsa-genome-assembly-tutorial/11703

hexylena:
Ugh this was always a nightmare, had so many issues with the .bz2 files. Incredibly frustrating, and the worfklow was always very difficult about the format, and I never understood why. 

a test of the file exits without error

```
$ bzip2 -t DRR187559_2.fastqsanger.bz2                             
bzip2 -t DRR187559_2.fastqsanger.bz2  7.43s user 0.01s system 96% cpu 7.723 total
```

and agreed not truncated

```
$ cat DRR187559_2.fastqsanger | tail
+DRR187559.451780 451780 length=209
CCCCCGGGGGGGGGGGGF@FGGGGGGGF9FFFGGGGGGFGGGGGGGGGGGGFGGGGGGGGGFFGGEFGGFD7FFGEFFGGGGGGGFGGGGGGGFFFGG<FGGGGGGGGFG9E+A<EF95=,FFFGGGGCFDA,@;D,,=B=,4@>=DFFA=EGC>8@FC,@FFC6,6@9DF@F@+0@,DFGA?,@,8?FF+68=>C9;FCFFFCCF**0
@DRR187559.451781 451781 length=165
[...]

PR #4732, opened 2024-02-16
Fix odd bzip2 record

hexylena:
Fixes https://github.com/galaxyproject/training-material/issues/4730


bebatut (approved):
Thanks a lot @hexylena for the fix

PR #4751, opened 2024-02-22
fixes broken RO Crate documentation links

SandyRogers:
There were two typos in the RO Crate Intro tutorial, in URLs for the RO Crate docs webpages.


SandyRogers:
> We currently don't do external link checking because there is so much link rot over time, but perhaps we can opt-in some tutorials like the RO-Crate ones where we know it's maintained, to ensure the links are checked. It's the second time we've had this situation so, clearly time to fix it :)

Thanks @hexylena !
Perhaps [an existing github action for markdown link checking](https://github.com/marketplace/actions/markdown-link-check) could be a fairly quick win ?

hexylena:
@SandyRogers it's an interesting suggestion! We have an existing pipeline step that checks internal links *post build* (and due to jekyll templating a lot of links include targets like `{% link path/to/file.md %}`) which might be marked as failures and would hvae to be ignored. Well, we'll look into it! Thanks for the tip

PR #4763, opened 2024-03-03
Typo fix in "Use Jupyter notebooks in Galaxy" tutorial

Borkan1:
Changed mentions of the RNA-seq file suffix from ".count" to ".counts" to match up with the actual linked file.


hexylena (approved):

hexylena:
Thanks @Borkan1 !

PR #4779, opened 2024-03-07
Describe how to view multiple multiQC outputs

paulzierep:
We need a quick fix, since we run a training tomorrow.
The issue is that multiQC does only show one file / collection; if the file names are the same.
Now we run multiQC twice.

### DIFF
--- a/topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md
+++ b/topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md
@@ -238,9 +238,18 @@ In this tutorial we use similar tools as described in the tutorial ["Quality con
     >                    - {% icon param-repeat %} *"Insert FastQC output"*
     >                        - *"Type of FastQC output?"*: `Raw data`
     >                        - {% icon param-files %} *"FastQC output"*: 4 `Raw data` outputs of **FastQC** {% icon tool %}
+    > 4. {% tool [MultiQC](toolshed.g2.bx.psu.edu/repos/iuc/multiqc/multiqc/1.11+galaxy0) %} with the following parameters:
+    >    - In *"Results"*:
     >        - {% icon param-repeat %} *"Insert Results"*
-    >            - *"Which tool was used generate logs?"*: `fastp`
-    >                - {% icon param-files %} *"Output of fastp"*: `JSON report` outputs of **fastp** {% icon tool %}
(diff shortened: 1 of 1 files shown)


topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md
```diff
@@ -238,9 +238,18 @@ In this tutorial we use similar tools as described in the tutorial ["Quality con
     >                    - {% icon param-repeat %} *"Insert FastQC output"*
     >                        - *"Type of FastQC output?"*: `Raw data`
     >                        - {% icon param-files %} *"FastQC output"*: 4 `Raw data` outputs of **FastQC** {% icon tool %}
+    > 4. {% tool [MultiQC](toolshed.g2.bx.psu.edu/repos/iuc/multiqc/multiqc/1.11+galaxy0) %} with the following parameters:
```
bernt-matthias, line 241: Why do you not use fastp for fastp output? 

Such a change should probably also be reflected in the workflow coming with the workflow, or?

paulzierep: You mean fastp as input to mutliQC or looking at the output of fastp ?

paulzierep: In order to show multiple fastqc reports in multiQC one needs to rename the input elements for fastQC, since this part: 
```
>>Basic Statistics	pass
#Measure	Value
Filename	Barcode11_Spike2b_fastq_gz.gz <<<<<<<<<<<<<<<<
File type	Conventional base calls
Encoding	Sanger / Illumina 1.9
```
is leading to the not rendering of the duplicated file names.

And that would mean to use ` Extract element identifiers`, `regex`, `Relable` for each additional multiQC input, that looks super complicated and difficutl for the students to follow. And also takes me to long to fix for tomorrow.
But I think @EngyNasr will fix it for IWC and then we can update here as well.

paulzierep:
Since we anyway to late for the training today, I will collect some more changes here and fix it the right way after all.
Next issue: Using Kalamari DB for host filtering - Kalamari does not contain all host reads, so this makes no sense, but could still be used, but needs to be explained correctly.

paulzierep:
Next one:
 `Parse parameter value` does not work outside a workflow (Gene based pathogenic identification - step)

PR #4790, opened 2024-03-11
Document collapsing boxes in reports/pages in an FAQ

hexylena:
- [x] collapse
- [x] show_column_headers=true
- [x] compact=true
- [x] title='The title'
- [x] footer='The footer'

also found out that if you want to use three backticks in a block, you have to use 4 in the parent block. And if you want 4, you guessed it, use 5.

markdown parsing is a literal unending nightmare. We hardcode looking for 3 backticks in places like the admin tutorial extraction, so, that's fun. Maybe need to switch to a real markdown engine eventually to fix that.

q: does it get filed as a Report faq or a Pages faq? More people probably use reports than pages but they're the same underlying thing.


hexylena:
https://github.com/galaxyproject/galaxy/pull/16681 more are documented here in images

shiltemann (approved):

PR #4796, opened 2024-03-11
vgp assembly: replaced est. genome size calculation with upload via paste data

abueg:
hello! 👋🏼 

simplifies the genomescope parameter part of the tutorial by having users just copy paste an existing number into a file, instead of parsing a file to get it


abueg:
@bgruening ah but when it's uploaded via copy&paste it's still available as input? like it becomes an item in the history and i can still use it later on:
image

for using it in workflows, it will need to get passed through the `parse parameter value from input` box which i think is used already when we get the parameter programmatically like this (because the evaluation result ends up `tabular` file if it's the result of `cut`, etc.)

bgruening (approved):

bgruening:
Thanks!

PR #4807, opened 2024-03-13
Add EDAM topic for apollo tutorial

bebatut:


topics/genome-annotation/tutorials/apollo/tutorial.md
```diff
@@ -48,6 +48,11 @@ requirements:
       - annotation-with-prokka
 
 subtopic: prokaryote
+edam_ontology:
+- topic_0622 # Genomics
```
hexylena, line 52: i guess this would be better at the topic level maybe? it should apply to more or less everything in genome-annotation

bebatut: But if we get rid of the topics at some point??

hexylena: Do you suspect that it is likely we will at some point? I'm not sure.  If we did, we would have to find another way to ensure each material has a unique 'key'/folder name.

and if we did, at that time we would copy the `edam_ontology` terms from the parent topic to each material, before we re-organised I guess?

hexylena (approved):

PR #4833, opened 2024-03-14
Alternative DESeq2 hands-on for ref-based RNA-seq tutorial

pavanvidem:
Choose your own tutorial for DESeq2 run. The current DESeq2 hands-on based on collection inputs with tags is not for everyone. It also needs 3 more steps to prepare the data for DESeq2. In most cases, users have only one factor with ~3 replicates to deal with. A basic datasets-based DESeq2 run should be sufficient.


pavanvidem:
When I tested locally, both the divs were displayed by default. Am I doing something wrong?

bebatut:
It should be fixed now

bebatut (approved):

PR #4843, opened 2024-03-18
vgp assembly: clarifying text to one of the hic images

abueg:
hello! 👋🏼 

added some clarifying text to one of the images in the HiC part of the tutorial

without the text, the images side-by-side could look like they are before-and-after, but they are two completely different examples


hexylena (approved):

abueg:
thank you!!

PR #4847, opened 2024-03-19
update advanced cut tool

Marie59:
Small update to fill in correctly the Advanced cut tool according to its new version.

### DIFF
--- a/topics/ecology/tutorials/obisindicators/tutorial.md
+++ b/topics/ecology/tutorials/obisindicators/tutorial.md
@@ -137,11 +137,13 @@ In the downloaded folder you should have your data either csv format (Occurence.
 
 > <hands-on-title>Clean your data</hands-on-title>
 > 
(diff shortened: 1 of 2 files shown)


bgruening (commented):
Do you need to update the worklow as well?

Marie59:
> Do you need to update the worklow as well?

I am not sure I don't think it changes anything for the workflow in the end ...

hexylena:
@Marie59 no, but bjoern is right, we like to the keep the version of the tool used in the workflow matching the tutorial itself. Could we trouble you to update the workflow as well to ensure it uses this same newer version of cut as well? :)

Marie59:
It should be up to date now ;)

hexylena:
Ah super!!

hexylena (approved):
Oh and even with some nice improvements! great to see :)

Marie59:
as always I don't have the test :)

PR #4851, opened 2024-03-20
FAIR medchem tutorials updates

wee-snufkin:
Hey @kkamieniecka!

Here is the PR with some updates:
- updated authorship and funding
- cross-link to the FAIR section


hexylena:
I believe the linting is failing because there is a `fair-data` subtopic in one, but you're symlinking it into a place where there isn't a `fair-data` subtopic which it's flagging as an error. I'm not sure how best to handle that, we want to flag it when people are putting tutorials into non-existent subtopics because 99% of the time it's a typo of the subtopic identifier which we want to catch but here it's a strange scenario where you'd probably rather it just show up in `other`. I don't know.

wee-snufkin:
Ugh, I wanted to fix it but I broke it even more 🤦

hexylena:
removing the subtopic was a good short term fix, but it looks like the symlink is wrong?

```
{"message":"This is a BAD symlink","location":{"path":"./topics/fair/tutorials/fair-med-chem-data","range":{"start":{"line":1,"column":1},"end":{"line":1,"column":1}}},"severity":"ERROR","code":{"value":"GTN:013","url":"https://github.com/galaxyproject/training-material/wiki/Error-Codes#gtn013"}}

No such file or directory @ rb_sysopen - /home/runner/work/training-material/training-material/topics/fair/tutorials/fair-med-chem-data
 (Errno::ENOENT)
```

wee-snufkin:
Oh no, now it's failing because each image is pointed to `topics/computational-chemistry/tutorials/med-chem-data`/<image_path> 

Any other option how to solve this without the need to create the images folder inside `topics/computational-chemistry/tutorials/med-chem-data` so that the generated paths are valid? 

Totally didn't expect this PR to be so convoluted, was supposed to be one simple line

hexylena (approved):

wee-snufkin:
Wohooo! Cheers @hexylena!

PR #4852, opened 2024-03-20
Simplify MSA generation steps in tutorials by using latest MAFFT version

wm75:
The latest wrapper version of MAFFT allows generating MSAs of sequences from multiple input datasets.

This eliminates the need to concatenate those datasets as an extra step.


nekrut (approved):

hexylena:
Too bad this is already merged, I wanted to make an additional request: I'm assuming you've tested this in a galaxy somewhere, could you export that workflow for this tutorial while you're there?

wm75:
No worries, WFs for these are coming (pox tiled amplicon actually has a iwc version already), I just wasn't happy with a few steps yet (mafft being one of them) and updating WFs is much more work than updating tutorials.

hexylena:
ah that's fantastic news, thanks @wm75.

PR #4866, opened 2024-03-26
Add fair RNA tutorial, add news about adding the pathway, and fix the ELIXIR dash funder username

khaled196:
Add fair RNA tutorial, add news about adding the pathway, and fix the ELIXIR dash funder username


FUNDERS.yaml
```diff
@@ -162,7 +162,7 @@ biont:
     funding_statement: |
       Co-funded by the European Union
 
-ELIXIR-UK-DaSH:
+elixir-uk:
```
hexylena, line 165: I know you've changed a lot of these, but, is there a reason for the change? Since this is a specific grant from elixir uk, it could make sense  to keep it as the more specific `elixir-uk-dash` and set `github: false` (which is what I'm assuming was the goal to fix the github link)

FUNDERS.yaml
```diff
@@ -162,7 +162,7 @@ biont:
...
     name: DASH UK
     joined: 2023-10
     avatar: "https://github.com/galaxyproject/training-material/blob/main/shared/images/dash.png?raw=true"
```
hexylena, line 168:
```suggestion
    avatar: "/training-material/shared/images/dash.png"
```

news/images/2024-03-26-emblem-fellowship.png
hexylena, line 1: I think this is the same as `shared/images/dash.png`, right? could that be referenced instead?

FUNDERS.yaml
khaled196: Thank you, Helena, Yes. That was the goal. I will return it to how it was

hexylena (approved):
This looks so fantastic!

Issue #4594, opened 2023-12-14
Feature Request: Example history in header


nomadscientist:
Can we use this symbol: 
galaxy-history

and add into the tutorial headers, when present:
![Screenshot 2023-12-14 at 23 37 49](https://github.com/galaxyproject/training-material/assets/44605769/2c91415e-9b00-49b2-b93c-71e0ac06408d)

Under 'supporting materials', an "Input history" and an "Answer Key History" or "Example history"? And allow for >1 if needed, a bit like how if you click workflows it takes you to multiple if present?

Why?

1 - Sometimes importing data from paste-fetch takes a while, so this is a quick way of getting the data across during courses
2 - Especially during courses, but also anytime someone has an error, it's super helpful to see a history where it worked.

PR #4881, opened 2024-04-04
GTN Header - Adding a Answer Key Histories drop-down menu

nomadscientist:
This is the first attempt to get the conversation started

I'm wondering if it would be ideal to have a landing page, like with the workflows, where we stipulate how to import a history, compare, and drag datasets between histories so they can help themselves if they get stuck.

Fixes https://github.com/galaxyproject/training-material/issues/4594


topics/single-cell/tutorials/scrna-case_alevin/tutorial.md
```diff
@@ -14,6 +14,17 @@ redirect_from:
...
+histories:
+<div class="dropdown">
+  <button class="btn btn-secondary dropdown-toggle" type="button" data-toggle="dropdown" aria-expanded="false">
+    {% icon galaxy-history %} Answer Key Histories
+  </button>
+  <div class="dropdown-menu">
+    <a class="dropdown-item" href="#">[galaxy.eu](https://humancellatlas.usegalaxy.eu/u/j.jakiela/h/generating-a-single-cell-matrix-using-alevin-3) </a>
+    <a class="dropdown-item" href="#"> [Older Alevin version](https://humancellatlas.usegalaxy.eu/u/wendi.bacon.training/h/cs1pre-processing-with-alevin---answer-key) </a>
+  </div>
+</div>
```
hexylena, line 26:
```suggestion
histories:
  current:
    label: "UseGalaxy.eu"
    history: https://humancellatlas.usegalaxy.eu/u/j.jakiela/h/generating-a-single-cell-matrix-using-alevin-3
  older-alevin:
    label: "Older Alevin version"
    history: https://humancellatlas.usegalaxy.eu/u/wendi.bacon.training/h/cs1pre-processing-with-alevin---answer-key
```

topics/single-cell/tutorials/scrna-case_alevin/histories_index.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: workflow-list
+redirect_from:
+  - /topics/transcriptomics/tutorials/droplet-quantification-preprocessing/workflows/index
```
hexylena, line 4:
```suggestion
layout: history-list
```

(todo: write this layout.)

hexylena (commented):
Thanks for getting the conversation started! 

I've proposed a more structured data format, and it's an interesting idea about the histories index. I'm not sure how best to handle that. 

You would envision people going to this separate page to see the available histories?

And I guess I wonder what we want to do in the future, with the workflow testing framework, if we should plan to attach these to specific workflows (somehow).

topics/single-cell/tutorials/scrna-case_alevin/tutorial.md
nomadscientist: oh score! Actually, that's a good point - I think the histories index could be a separate level

nomadscientist:
I put an 'input histories' there, but actually, that's probably not necessary (although I could absolutely see a future wherein the 'Get data' sections automatically pull from the input histories just like they pull from Zenodo, but it's not strictly needed atm).

I think mentally I was thinking it would be valuable to quickly see which tutorials have input histories and which don't, so that we could make the input histories and keep everything up to date, but that probably goes with the whole workflow / history / automation plan and isn't necessary NOW, so happy for it to be binned.

hexylena:
I think this is a good interim plan. We can make something like this work

nomadscientist:
Ahk, bad me, I've put this in two places
https://github.com/galaxyproject/training-material/issues/4594
Well when we make this work then we can resolve that issue!

hexylena:
> Ahk, bad me, I've put this in two places
> https://github.com/galaxyproject/training-material/issues/4594

no this is fine, perfect even. you filed an issue to see if the implementation was reasonable/desired, we agreed, and now you've opened a PR to implement this (which we will get around to eventually i promise). I've tagged this PR as "Fixes #4594" which will ensure that when this is merged the associated issue is auto-closed.

that's a great workflow! Especially for feature requests where you want feedback on if it's possible or so.

hexylena:
Looks good I think, added an optional 'date' field that can be used (or not, or automatically extracted from the history/etc).

Screenshot 2024-05-13 at 15-59-55 Single Cell _ Hands-on Generating a single cell matrix using Alevin

hexylena (approved):

hexylena (dismissed):

hexylena (approved):

nomadscientist:
Woooo!

PR #4887, opened 2024-04-10
Remove spaces in front of zenodo file names in VGP tutorial

AnnaSyme:
Remove spaces in front of zenodo file names, because if not, the files may not upload correctly in Galaxy. 
I think this formatting still looks ok in the preview but could someone else also check, thanks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,10 @@ jobs:
         run: |
           ! fgrep -R 'lt;blockquote' _site
 
+      - name: Ensure no poorly indented code blocks
+        run: |
+          bundle exec ruby bin/check-indent.rb
+
       - name: Ensure no unexpected jekyll in output
         run: |
           ! fgrep -R 'site.pages' _site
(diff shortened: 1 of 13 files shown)


hexylena (changes requested):
This is unfortunately the wrong fix. The indentation is correct and was correct. But because of https://github.com/galaxyproject/training-material/commit/70662e5d739568f686782687774d8ef9b7458a2f where it was changed from a semantic list `1. text` to this bolded semantically incorrect version, this broke. 

we should revert it back to a proper semantic list.

hexylena:
I can do that later and add it to this PR, apologies @AnnaSyme

AnnaSyme:
ah I see. thanks @hexylena!

hexylena:
I've added a script to check for it and fixed the rest of the errors (there were a handful more, so, cheers to vgp for triggering this.)

The VGP tutorials are very bad about using these non-semantic `**Step #** ...` which should be removed at some point since they're a bad experience for screenreader users.

Additionally the workflow import process could be improved significantly. See for example how we do the workflow import in this news item where you can simply click a button.  https://training.galaxyproject.org/training-material/news/2024/03/28/by-covid-pathways.html (I'm not sure who from VGP is working on tutorials currently so just noting it out loud to no one in particular.)

hexylena (approved):
ok, fixed differently. Thanks for raising it @AnnaSyme, the GTN now has more linting to prevent this in the future :)

Issue #4349, opened 2023-09-18
Make section headings easier to link to


hexylena:
https://amberwilson.co.uk/blog/are-your-anchor-links-accessible/#what-are-anchor-links-exactly%3F

PR #4889, opened 2024-04-10
Make sections easier to link to by adding deep anchor tags

shiltemann:
Attempt to fix https://github.com/galaxyproject/training-material/issues/4349

Adds links to sections and all boxes on hover (can be customized).

anchors2

@hexylena could you check accessibility on screen readers? looks like there are some config settings we can use to optimize this.


hexylena:
yes can do. I've had https://amberwilson.co.uk/blog/are-your-anchor-links-accessible/#what-are-anchor-links-exactly%3F bookmarked for years now but hadn't gotten around to it yet, glad you could! I'll let ya know

hexylena (approved):
This looks good! 

I've rewritten it a tad but, I think I'm happy with it! Thanks for adding this :)

PR #4901, opened 2024-04-16
Add ELIXIR and a few other affiliations

bgruening:
Here we are adding ELIXIR affiliations to a lot of ELIXIR members and also a few other affies that I could spot.

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -68,6 +68,7 @@ abretaud:
     affiliations:
       - gallantries
       - eurosciencegateway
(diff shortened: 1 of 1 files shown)


CONTRIBUTORS.yaml
```diff
@@ -916,6 +953,10 @@ jj-umn:
     name: James Johnson
     email: jj@umn.edu
     joined: 2017-09
+    in_memoriam: |
```
hexylena, line 956: thank you for doing this

hexylena (approved):

PR #4904, opened 2024-04-18
[VGP assembly] clarifying text in scaffolding tooltip

abueg:
previous text was confusing, we do not scaffold the alternate assembly in the pipeline

### DIFF
--- a/topics/assembly/tutorials/vgp_genome_assembly/tutorial.md
+++ b/topics/assembly/tutorials/vgp_genome_assembly/tutorial.md
@@ -1291,7 +1291,7 @@ At this point, we have a set of contigs, which may or may not be fully phased, d
 
 > <comment-title>What assembly am I scaffolding??</comment-title>
 >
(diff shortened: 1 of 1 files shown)


abueg:
ah oops, i didn't build a local mirror because i was just changing the one line 😅 

sorry, i'm not super familiar with how to make the lists semantic thru the MD->HTML, should the ordinal list be encoded in a different way? i see the numbers on the front-end site, but there aren't numbers in the HTML, just the `<ol>` tag...

i can make the accessibility changes sometime next week, but if this edit to fix an inaccuracy in the text could be pushed before then, i would appreciate that!

hexylena (approved):
Happy to merge this small improvement now :) 

That is a new linter we added recently, to fix it one would just replace

```
**Step 1**: Some text
```

with

```
1. Some Text
```

i.e. a proper markdown list

hexylena:
Thanks for the improvement @abueg !

_If you have time for it_, to fix the formatting later, that would be fantastic, otherwise we will get around to it eventually on our end.

PR #4908, opened 2024-04-22
dev.wfh upload

hexylena:
todo:

definitely:

- [x] finalise our ID format since it is *forever*.
- [x] bot account for uploading (https://workflowhub.eu/people/731, `PROD_WFH_TOKEN`)
- [x] improve WF description
- [x] improve tags (uniform capitalisation, remove `name:`, etc.)
- [x] discussion channels (worked around https://github.com/seek4science/seek/issues/1928)
- [x] permissions (gtn + some of our leads need edit permissions)

maybe:

- [x] can we generate a useful image? we have the mermaid if we could render that (or with `dot` or similar) https://dev.workflowhub.eu/workflows/830
- [ ] related organisations?
- [ ] ~~edam topics from tutorial?~~ no goes through API.

future work:

- [ ] enable DOIs
- [ ] benchmarks
- [x] per-WF pages.

https://dev.workflowhub.eu/workflows?filter%5Bcontributor%5D=103&page=1&view=table


hexylena:
I fear this might be ready to ship.

The deployment plan is:

1. Merge this
2. Run the workflow (generates metadata/workflowhub.yml)
3. Check that everything looks correct, workflows point nicely to dev.workflowhub.eu, everythings rendered correctly on their end.
4. A new PR will update to point to the production workflowhub.eu, remove metadata/workflowhub.yml (so we don't have any incorrect IDs for sure), and then it'll be live live.

hexylena:
It may require more changes but that's what dev.wfh is for.

shiltemann (approved):

PR #4913, opened 2024-04-24
Elixir ack

bgruening:
Does that look useful? If so I can do the rest in my next meetings.

### DIFF
--- a/topics/epigenetics/tutorials/ewas-suite/tutorial.md
+++ b/topics/epigenetics/tutorials/ewas-suite/tutorial.md
@@ -13,10 +13,15 @@ objectives:
 key_points:
 - "Infinium Human Methylation BeadChip is an array based technology to generate DNA methylation profiling at individual CpG loci in the human genome based on Illumina’s bead technology."
 - "Time and cost efficiency followed by high sample output, and overall quantitative accuracy and reproducibility made Infinium Human Methylation BeadChip one of the most widely used arrays on the market."
-contributors:
-  - kkamieniecka
-  - khaled196
-  - poterlowicz-lab
+contributions:
+  authorship:
+    - kkamieniecka
+    - khaled196
+    - poterlowicz-lab
+  editing:
+  testing:
+  funding:
+    - elixir
 ---
 
 This tutorial is based on Hugo W, Shi H, Sun L, Piva M et al.: Non-genomic and Immune Evolution of Melanoma Acquiring MAPKi Resistance {% cite Hugo2015 %}.
--- a/topics/epigenetics/tutorials/ewas-suite/tutorial.md
+++ b/topics/epigenetics/tutorials/ewas-suite/tutorial.md
@@ -21,7 +21,7 @@ contributions:
   editing:
   testing:
   funding:
-    - elixir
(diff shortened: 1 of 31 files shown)


topics/epigenetics/tutorials/ewas-suite/tutorial.md
```diff
@@ -13,10 +13,15 @@ objectives:
...
-  - khaled196
-  - poterlowicz-lab
+contributions:
+  authorship:
+    - kkamieniecka
+    - khaled196
+    - poterlowicz-lab
+  editing:
+  testing:
+  funding:
```
hexylena, line 23: I'm not sure about this one, @khaled196 @kkamieniecka was this also part of DASH or a previous grant?

topics/genome-annotation/tutorials/apollo-euk/tutorial.md
```diff
@@ -33,6 +33,8 @@ contributions:
   funding:
     - gallantries
     - eurosciencegateway
+    - elixir-europe
+    - ifb
```
hexylena, line 37: not sure about these here?  that work was more or less strictly gallantries funded, we weren't paid by e.g. the elixir galaxy IS or anything?

topics/genome-annotation/tutorials/apollo/tutorial.md
```diff
@@ -36,6 +36,8 @@ contributions:
   funding:
     - gallantries
     - eurosciencegateway
+    - elixir-europe
+    - ifb
```
hexylena, line 40: same

hexylena (commented):
some of these look like improvements (especially for the ones you personally know/are from your group), but some of these I'm less sure about.

I think there's a distinction to be made between `elixir-europe` (an organisation that people are members of) and specific funding grants from them like the elixir galaxy IS and similar. 

if we need to show "tutorials produced by members of this org" on an org's page, in a separate category, that could make sense. But tagging it elixir-europe seems off, I think for a number of these it would make more sense for me if they were tagged something from the FUNDERS.yaml file.

bgruening:
Ok, then we can close this one I guess and wait for the funders.yml support.

hexylena:
> funders.yml support.

the funders.yml file is there currently, do you want to add entries for the IS? for any other specific elixir grants? (do we know which projects there were over the last few years that were elixir funded?)

it's also a bug on our end, that the funding field allows organisations, I think we should restrict that and fix the old cases. it's not always a clear distinction, especially with organisations like elixir. orgs.yaml is mostly employers, and funders.yml is mostly grants with a grant ID in a public database, but there are also cases where folks are not wanting to go into this level of detail

bgruening:
Funding is complicated and I doubt we will cover all aspects correctly, or if we do it will be so complex that most people will not use it.

ELIXIR is paying for a lot of events, cares for and funds community building, and conferences. I think this is helping GTN and Galaxy a lot, even if you can nail this down to a specific grant number. How do you acknowledge the ELIXIR-funded trip to AU - which yielded to a few new GTN materials? I don't have an answer to this and don't want to make the system here much more complicated.

PR #4924, opened 2024-05-01
Upgrading workflows in large genome assembly tutorial

AnnaSyme:
- Upgraded all tools in all workflows to latest versions and tested to ensure it still works (on Galaxy Australia). 
- Added the new versions of all 9 workflows in this tutorial. 
- (Some of these are subworkflows within the combined workflow.)
- To do this, I first re-named all the existing workflows with the suffix `_old`.
- Then I uploaded the upgraded workflows that are named in the same way as the original set. 
- This is so I didn't need to edit the references to all the workflows within the tutorial.md text. 
- I can't see any toolshed IDs in the tutorial.md text that need to be updated, as this tutorial refers to workflows rather than tools.


hexylena:
Looks like you're missing the license field on the workflows, could we trouble you to add those?

AnnaSyme:
> Looks like you're missing the license field on the workflows, could we trouble you to add those?

Yep, should I use  "license": "GPL-3.0-or-later" ?

hexylena:
Yes!

hexylena (dismissed):

AnnaSyme:
> Yes! (that's the syntax, but of course you're free to choose any spdx license)

thanks @hexylena have added these

hexylena:
Could you remove all of the 'old' workflows? They'll be retained in the git history but they won't pass linting and I guess we don't need them anymore?

AnnaSyme:
> Could you remove all of the 'old' workflows? They'll be retained in the git history but they won't pass linting and I guess we don't need them anymore?

yep will do this now

hexylena (approved):
looks like a nice improvement

PR #4983, opened 2024-06-03
import onedata blogpost

hexylena:
@bgruening any additional orgs/funders/authors/infrastructure folks to credit here? (we use `infrastructure:` for crediting server admins who set stuff up for us/etc. in this case also probably whoever wrote the code.)


bgruening:
for `orgs` maybe [EGI](https://www.egi.eu/)

hexylena:
Done!

shiltemann (approved):

PR #4984, opened 2024-06-03
maintainer home page

hexylena:
for @nomadscientist who wants all of the information she needs as a topic editor to be in one easy-to-find place.

works now:

- only for single cell (waiting feedback from @nomadscientist before scaling up)
- basic checklist of topic maintenance tasks (summary, enough editorial board members, various metadata you might want to annotate)
- statistics for all pages in that topic
- stats for all learning pathways that include materials from that topic.

will use it for someday

- workflow testing results across a topic.
- your idea here.

Screenshot 2024-06-03 at 12-55-36 Single Cell

### DIFF
--- a/_layouts/topic-maintainer.html
+++ b/_layouts/topic-maintainer.html
@@ -1,0 +1,99 @@
+---
+layout: base
+---
+
+{% assign topic = site.data[page.topic_name] | default: page.topic %}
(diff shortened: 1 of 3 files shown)


shiltemann (approved):

nomadscientist:
My only request here is that "Goal Conversions" I find quite vague to understand. Can that be relabelled? Then I might use it more

PR #4985, opened 2024-06-03
ensure contribs/orgs/funders are linted

hexylena:
since the change got missed in #4980 and has a key that's causing other PRs to fail.


bgruening (approved):

bgruening:
Thanks @hexylena!

PR #4989, opened 2024-06-03
Update 2024-06-10-mtb-ngs.html

dbrites:
Merged organizers and instructors

Added a new video to day 1

added bold to the different activities, "webinar", "Q&A", etc.

### DIFF
--- a/events/2024-06-10-mtb-ngs.html
+++ b/events/2024-06-10-mtb-ngs.html
@@ -51,6 +51,8 @@ infrastructure:
     channel: event-mtb-ngs
 
 program:
+  - section:
+    description: |
       There were unusually high rates of TB cases in your country this year. To characterize the underlying
       bacterial strains driving the epidemic, isolates have been sent for whole-genome sequencing.  Doctors and public
       health authorities request information in order to take decisions. In this course it will be demonstrated how
--- a/events/2024-06-10-mtb-ngs.html
+++ b/events/2024-06-10-mtb-ngs.html
@@ -14,7 +14,7 @@ date_start: 2024-06-10
 date_end: 2024-06-14 # optional, if event is more than one day
 
 contributions:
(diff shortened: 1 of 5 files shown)


events/2024-06-10-mtb-ngs.html
```diff
@@ -14,15 +14,15 @@
 date_end: 2024-06-14 # optional, if event is more than one day
 
 contributions:
-  organisers: # GTN contributors or funders, must be defined in CONTRIBUTORS.yaml
-  - dbrites
-  instructors:
+  organisers &  instructors: # GTN contributors or funders, must be defined in CONTRIBUTORS.yaml
```
shiltemann, line 17: @dbrites this will not be valid. You can either put everybody as organizers, or everybody as instructors. It is also possible to put people both under organisers and under instructors (e.g. you?). So whatever you prefer.

events/2024-06-10-mtb-ngs.html
```diff
@@ -14,15 +14,15 @@
...
   - dbrites
   - cstritt
   - andreacabibbe
   - arashghodousi
   - federicodimarco
   - pvanheus
+  - Liliana Rutaihwa
+  - Linzy Elton
```
shiltemann, line 25: These two contributors will have to be added to the CONTRIBUTORS.yaml file. I will do that for you, just let me know if they have github accounts, and what their affilliations are (not mandatory, but we can add that if you want)

events/2024-06-10-mtb-ngs.html
```diff
@@ -51,8 +51,6 @@
     channel: event-mtb-ngs
 
 program:
-  - section: "Day 1: Overview of NGS technologies & TB specific NGS solutions"
-    description: |
```
shiltemann, line 55: so you would like this text to appear above the "Day 1" header right? (we will have to do that a bit differently, but we can make that work)

shiltemann (approved):

PR #5011, opened 2024-06-06
Pathogen detection from (direct Nanopore) sequencing data using Galaxy - Foodborne Edition - Training update to latest workflows update

EngyNasr:
Adding Tags to main workflow outputs makes it easier to highlight some results for the training takers. 

Thanks,
Engy


bebatut (approved):
Thanks @EngyNasr

PR #5024, opened 2024-06-11
fix a little typo in ref-bas rna-seq tutorial.md

nilchia:


lldelisle:
Thank you very much for spotting it.

lldelisle (approved):

PR #5059, opened 2024-06-19
Pathogen detection from (direct Nanopore) sequencing data using Galaxy - Foodborne Edition

EngyNasr:
Correcting some training explanations in the preprocessing workflow, and adding a missing step in the long version of the training

Thanks @paulzierep for spotting them :) 

Thanks,
Engy

### DIFF
--- a/topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md
+++ b/topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md
@@ -198,6 +198,14 @@ In this tutorial we use similar tools as described in the tutorial ["Quality con
     >    > This step, as it does not require the results of FastQC to run, can be launched even if FastQC is not ready
     >    {: .comment}
     >
+    > 3. {% tool [MultiQC](toolshed.g2.bx.psu.edu/repos/iuc/multiqc/multiqc/1.11+galaxy0) %} with the following parameters:
+    >    - In *"Results"*:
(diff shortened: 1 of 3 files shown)


paulzierep:
Thanks Engy. I will go trough it now and add more when I find them:
Next one: Can you explain what samples profile is/means also in the Training: https://github.com/galaxyproject/training-material/blob/56f79079677e95a64567889f4c06802504afb611/topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md?plain=1#L163

topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md
```diff
@@ -198,6 +198,14 @@ In this tutorial we use similar tools as described in the tutorial ["Quality con
     >    > This step, as it does not require the results of FastQC to run, can be launched even if FastQC is not ready
```
paulzierep, line 198:
```suggestion
    >    > The `NanoPlot` step, as it does not require the results of FastQC to run, can be launched even if FastQC is not ready
```

paulzierep:
Here: https://github.com/galaxyproject/training-material/blob/56f79079677e95a64567889f4c06802504afb611/topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/tutorial.md?plain=1#L299 
We could add that if the host is not know, kraken2 with calamari could be used to detect it.

paulzierep (approved):

PR #5066, opened 2024-06-23
Miscellaneous stuff

hexylena:
- branding resources page
- feed fix for empty events
- add an opml
- missing redirect for an ancient topic
- fix cors proxy


hexylena:
jeez, 6 minutes just for node 19 to get setup, we really need to get rid of our javascript code.

PR #5076, opened 2024-06-25
Further RTL linting

hexylena:
because we keep adding left/right instead of start/end

TODO: someone should check these look correct. (manually set `dir="rtl"` in the html files you look at)


hexylena:
this has gotten out of date, sorry arabic and other RTL speakers.

PR #5078, opened 2024-06-25
Fixed columns selection

matuskalas:
1. The c[...] expression comes from the workflow, but cannot be used when launching Advanced cut individually.

2. The column numbers were outdated (and different in the 2 scenarios). As these seem to be changing n the source data model, better not mention them.


hexylena (approved):

hexylena:
thanks @matuskalas !

PR #5093, opened 2024-06-28
Update analysis_troubleshooting.md

nomadscientist:
Change matrix to Slack for single-cell users


hexylena:
could you also add this to the footer.html ?

PR #5094, opened 2024-06-28
Change Matrix to Slack for user-facing channels

nomadscientist:
Have changed the links to point towards user-facing single-cell channels


hexylena:
great! can you also put it in the footer in this pr please? next to the matrix link, should be _includes/footer.html or so

nomadscientist:
I'm really sorry, I don't understand - in that same FAQ, or in the index.md page, or in the tutorial page? I was looking for anything that said '_includes/footer.html' so that I could copy that format but couldn't find it.

shiltemann (approved):

nomadscientist:
Thanks for sorting me out. Now I know!
Wow I didn't know that about Slack and the 400-limit!

PR #5113, opened 2024-07-02
Biont event added as external event

teresa-m:
BioNT event added as external event


teresa-m:
Is the indentation the issue? I used one of the [previous events](https://raw.githubusercontent.com/galaxyproject/training-material/main/events/2024-06-18-FAIR-data-management.md) as a template. Locally I did not get an error, and from the error message here I am a bit puzzled. It seems to be an Error with the FAQs, which I should not have changed or are involved in building this event, or are they? @hexylena @shiltemann do you have an idea what could be wrong and what I could check?

shiltemann:
@teresa-m the error seems unrelated to your changes, I will check and merge today, thanks for adding it!

shiltemann (approved):

shiltemann:
P.S. We are working on a Google form to submit external events like this, so you don't have to worry about formatting, coming soon :)

teresa-m:
Cool. Thanks a lot!

PR #5155, opened 2024-07-12
SnapATAC2 tutorial: update data-library.yaml

pavanvidem:
The latest version of this  Zenodo record has a compressed fragments file which is 4 times smaller than the old one.


mtekman (approved):
I checked the links, they seem good

PR #5162, opened 2024-07-15
Add a tutorial section for renaming sub-WF outputs

RZ9082:
I added a tutorial section about renaming sub-WF output using WF variables.
@paulzierep @mvdbeek Whenever you have some spare time, could you please review this? Thanks a lot!

### DIFF
--- a/topics/galaxy-interface/tutorials/workflow-parameters/tutorial.md
+++ b/topics/galaxy-interface/tutorials/workflow-parameters/tutorial.md
@@ -19,6 +19,7 @@ key_points:
 contributors:
 - mvdbeek
 - hexylena
+- RZ9082
 level: Intermediate
 subtopic: workflows
 
@@ -222,6 +223,7 @@ In this example we will construct a workflow where we calculate the sum of all v
 
 With this you're ready to run the workflow!
 
+> <a name="running-the-workflow-2"></a>
 > <hands-on-title>Running the workflow</hands-on-title>
 >
 > 1. Upload a dataset using "Paste/Fetch data" with the contents, and set the filetype manually to "tabular" during upload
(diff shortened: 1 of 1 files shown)


topics/galaxy-interface/tutorials/workflow-parameters/tutorial.md
```diff
@@ -222,6 +223,7 @@ In this example we will construct a workflow where we calculate the sum of all v
 
 With this you're ready to run the workflow!
 
+> <a name="running-the-workflow-2"></a>
```
hexylena, line 226:
```suggestion
```

every hands on box has an automatic ID, please use that instead.

https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/workflow-parameters/tutorial.html#hands-on-running-the-workflow

topics/galaxy-interface/tutorials/workflow-parameters/tutorial.md
```diff
@@ -294,7 +293,7 @@ In the following example we will construct a straightforward workflow, and utili
 
 > <hands-on-title>Running the main workflow</hands-on-title>
 >
-> 1. Upload/reuse the tabular file from the following [section](#running-the-workflow-2)
+> 1. Upload/reuse the tabular file from the following [section](#hands-on-running-the-workflow-1)
```
hexylena, line 296:
```suggestion
> 1. Upload/reuse the tabular file from the following [section](#hands-on-running-the-workflow)
```

your `-1` suggests to me you're running the training materials locally, in `--incremental` mode. (or `make serve-quick`) there is a known issue whereby if you're running it locally, it will append incrementing numbers to the end of IDs. Those numbers should only appear for 'duplicate' header IDs, but due to how the incremental mode works, every time that page is rebuilt, the number is increased. The online version of the material uses `hands-on-running-the-workflow` without the `-1`, and given that you haven't added another section with that ID, that should be the correct final value. Sorry about this odd bug!

RZ9082: Yes i noticed that difference between local and online.
Actually i am referring to this section here: https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/workflow-parameters/tutorial.html#hands-on-running-the-workflow-1
It is the second 'Running the workflow' header.
Thank for reviewing :)

bgruening (approved):
Thanks @RZ9082! LGTM, will wait for Marius a few days and then we can get it in I think.

mvdbeek:
Sorry, I had missed this. Yeah, looks good, but this is not related to subworkflows. You should use this always, and in fact the best paractice panel will warn you if you use a variable that doesn't match a formal input.

shiltemann:
@mvdbeek so do you propose some changes or can this go in as is or?

PR #5184, opened 2024-07-24
Fix rendering issue in Filter Plot and Explore (Python) tutorial

hexhowells:
Some of the code blocks were not being rendered correctly in the tutorial and generated Jupyter notebook due to some missing python tags and trailing whitespaces after closing one of the code blocks.

Also made some minor updates to some of the code blocks:
- The initial call to ```get()``` should get the first file since that will likely be the position of the required file if following the tutorial
- Fixed the name of the file saved at the end using ```adata.write()``` not matching the name of the file being saved in the next code block using ```put()``` (MarkersCluster.h5ad)
- Changed the filename of the exported Jupyter notebook file to make it slightly more clear that it needs to be the name of the current Jupyter notebook

### DIFF
--- a/topics/single-cell/tutorials/scrna-case-jupyter_basic-pipeline/tutorial.md
+++ b/topics/single-cell/tutorials/scrna-case-jupyter_basic-pipeline/tutorial.md
@@ -88,7 +88,7 @@ import pandas as pd
 You can import files from your Galaxy history directly using the following code. This will depend on what number in your history the final annotated object is. If your object is dataset #4 in your history, then you import it with the following:
 
 ```python
-mito_counted_anndata = get(4)
+mito_counted_anndata = get(1)
 ```
 
 You now need to read it in as a h5ad object.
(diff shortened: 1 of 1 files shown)


topics/single-cell/tutorials/scrna-case-jupyter_basic-pipeline/tutorial.md
```diff
@@ -88,7 +88,7 @@ import pandas as pd
 You can import files from your Galaxy history directly using the following code. This will depend on what number in your history the final annotated object is. If your object is dataset #4 in your history, then you import it with the following:
```
hexylena, line 88:
```suggestion
You can import files from your Galaxy history directly using the following code. This will depend on what number in your history the final annotated object is. If your object is dataset #1 in your history, then you import it with the following:
```

bgruening (approved):

bgruening:
Thanks!

PR #5190, opened 2024-07-31
Add beacon2 tutorial

khaled196:


shiltemann:
Thanks @khaled196!

This tutorial feels very 'developer-y' and I am a bit afraid it will scare researchers away from Beacons if we put it in the science topic. Is there any way we could have a tutorial just about querying an already existing Beacon here in the variant analysis topic, and then move this full tutorial to admin/dev topic? What do you think?

khaled196:
HI @shiltemann 

Both of the ideas are great, and both are doable, but I am concerned about the querying part, I have a beacon for CNV from the 1000hg which does not have the full beacon protocol collection (genomicVariations, runs, biosample, etc). Can the tutorial still be made for just the CNV? 

We are working on creating a beacon, but it may take some time.

shiltemann:
Hi @khaled196! My feeling is that it would still be a good showcase of how beacons work even if it's only CNV for now. So it would be great to have that, and hopefully we can extend it to more variant types in the future once there is a good 'full' public beacon available we can use.

khaled196:
@shiltemann 

I added another tutorial as a prequel. Please have a look at it

topics/variant-analysis/tutorials/beacon_cnv_query/tutorial.md
```diff
@@ -0,0 +1,134 @@
...
+key_points:
+- Understanding Beacon Databases involves grasping their purpose and significance in genomic research and data sharing.
+- Mastering the Beacon2 CNV Tool includes developing skills to effectively query and analyze specific CNVs within the Beacon2 database.
+- Efficient Querying for CNVs focuses on accurate data retrieval and precise query formulation using the Beacon2 tool.
+- Hands-On Application provides practical experience in querying Beacons, ensuring data integrity, and troubleshooting queries.
+- Recognizing the Importance of CNVs highlights the role of CNVs in genetic variation and their relevance in research using Beacon databases.
+contributors:
+- khaled196
+- poterlowicz-lab
+---
```
shiltemann, line 24: you can also define the other tutorial as a follow-up training here, then it will be recommended at the bottom of the tutorial, with syntax like:

```
follow_up_training:
- type: internal
  topic_name: statistics
  tutorials:
      - age-prediction-with-ml
```

(see also https://training.galaxyproject.org/training-material/topics/contributing/tutorials/schemas/tutorial.html#follow_up_training)

topics/variant-analysis/tutorials/beaconise_1000hg/tutorial.md
```diff
@@ -0,0 +1,495 @@
...
+- Analyze and transform variants and metadata into structures compatible with Beacon requirements
+- Execute a step-by-step process to import data into Beacons
+- Develop the ability to query Beacons for variants
+requirements:
+- type: external
+  title: The Unix Shell
+  link: "http://swcarpentry.github.io/shell-novice/"
+- type: external
+  title: Version Control with Git
+  link: "https://swcarpentry.github.io/git-novice/"
```
shiltemann, line 26: would you like to add the cnv beacon as a prerequisite her?

shiltemann (commented):
@khaled196 thanks a lot!

khaled196:
HI @shiltemann 

I made additional changes. Please take a look and give me your feedback.

shiltemann (approved):

shiltemann:
Thanks a lot @khaled196!

PR #5192, opened 2024-08-01
[Bacterial Genome Annotation] Update tool versions

bebatut:
In this PR, the tools used in the Bacterial Genome Annotation tutorial are updated


paulzierep (approved):
Thanks !

PR #5205, opened 2024-08-06
fix typing error

khaled196:
Minor correction to the tutorial


shiltemann:
good catch, thanks @khaled196 :+1:

PR #5232, opened 2024-08-15
Update short introduction tutorial.md with interface changes for 24.1

ahmedhamidawan:
All screenshots are updated to what we have in 24.1. The masthead is actually from dev. It will be in the next release (24.2) and pushing users to navigate around using the activity bar instead felt right.


ahmedhamidawan:
This is my first contribution PR for GTN, so I was a little confused on whether (and where) I should add myself as a contributor. Please let me know if I got that part right?

shiltemann (approved):

PR #5273, opened 2024-09-01
Create tool version FAQ to help debugging

nomadscientist:
This will help users debug a weird error message to check the tool version when doing the scRNA-seq tutorials using the EBI Scanpy tool suite

### DIFF
--- a/topics/single-cell/faqs/sc_version.md
+++ b/topics/single-cell/faqs/sc_version.md
@@ -1,0 +1,63 @@
+---
+redirect_from:
+- /topics/transcriptomics/faqs/amplification
+
+title: Why is my tool erroring as 'Above error raised while reading key '/layers' of type <class 'h5py._hl.group.Group'> from /.'
+area: Analysis
+box_type: tip
+layout: faq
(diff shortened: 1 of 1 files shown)


topics/single-cell/faqs/sc_version.md
```diff
@@ -0,0 +1,63 @@
+---
+redirect_from:
+- /topics/transcriptomics/faqs/amplification
```
shiltemann, line 3: is this intended or a copy/paste from the amplification FAQ?

(Also, it is better to create symlinks if you want an FAQ to show up in 2 places, this should just be used if you have moved or renamed the physical file from this location)

nomadscientist: Oh good to know about symlinks! Does that work for tutorials as well?

shiltemann (approved):

topics/single-cell/faqs/sc_version.md
shiltemann: yep, works for any FAQs :+1:

PR #5288, opened 2024-09-06
Add tips to regular expression FAQ

Delphine-L:
Add a couple of useful regex : 
- Conversion to lower and upper cases
- get whole match with `&`
- a couple useful examples for matching fasta files (I am not attached to it I can remove it if you think that is too specific)

### DIFF
--- a/faqs/galaxy/analysis_regular_expressions.md
+++ b/faqs/galaxy/analysis_regular_expressions.md
@@ -43,11 +43,13 @@ Below are just a few examples of basic expressions:
 | `chr\d{1,2}`         | `chr` followed by 1 or 2 digits |
 | `.*abc$`             | anything with `abc` at the end of the line |
 | `^$`                 | empty line |
+| `^>.*`			   | Line starting with `>` (e.g. Fasta header) |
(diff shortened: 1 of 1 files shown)


faqs/galaxy/analysis_regular_expressions.md
```diff
@@ -43,11 +43,13 @@ Below are just a few examples of basic expressions:
 | `chr\d{1,2}`         | `chr` followed by 1 or 2 digits |
 | `.*abc$`             | anything with `abc` at the end of the line |
 | `^$`                 | empty line |
+| `^>.*`			   | Line starting with `>` (e.g. Fasta header) |
```
bgruening, line 46: @Delphine-L what do you think about adding `regular expression` to this list in Galaxy https://github.com/galaxyproject/galaxy/pull/18722/files#diff-1832837c9e0fd387cc997af7d63fa5e92dc1e50a311e36dd09f8536469747d6c

Explain what that term is and add a link to this FAQ?

ping @jmchilton as well for input

bgruening (approved):

PR #5300, opened 2024-09-11
Updating case study tutorials

MarisaJL:
- Some minor updates to tool names, links 
- Also changed the bit on the order in which Anndata are concatenated in the combining datasets tutorial - the first one you click is the first one added, so they should click N702 to N707 not 7 to 2

New changes:
-  The Filter, Plot, Explore tutorial has also been checked now - only change is that we need to include the older input history as the new one doesn't work for the current tool versions/workflow - I'm not sure if we should have both available for now or just have the older one until we update the tools?
- Also added a note for people using the newer tools on the trajectories tutorial as they need to change n-pcs (again, we can improve this when we update the tool versions)

### DIFF
--- a/topics/single-cell/tutorials/scrna-case_alevin-combine-datasets/tutorial.md
+++ b/topics/single-cell/tutorials/scrna-case_alevin-combine-datasets/tutorial.md
@@ -168,9 +168,13 @@ You can access the data for this tutorial in multiple ways:
 > 1. {% tool [Manipulate AnnData](toolshed.g2.bx.psu.edu/repos/iuc/anndata_manipulate/anndata_manipulate/0.10.3+galaxy0) %} with the following parameters:
 >    - {% icon param-file %} *"Annotated data matrix"*: `N701-400k`
 >    - *"Function to manipulate the object"*: `Concatenate along the observations axis`
->    - {% icon param-file %} *"Annotated data matrix to add"*: `Select all the other matrix files from bottom to top, N702 to N707` (if you imported files from Zenodo, your history may not be in this order, but you can still select them in order from N702 to N707 using the file names)
->
->    > <warning-title>N701 to N701!</warning-title>
+>    - {% icon param-file %} *"Annotated data matrix to add"*: `Select all the other matrix files from bottom to top, N702 to N707`
+> 
+>    <comment-title></comment-title>
+>     >If you imported files from Zenodo instead of using the input history, yours might not be in the same order as ours. Since the files will be concatenated in the order that you click, it will be helpful if you click them in the same order, from N702 to N707. This will ensure your samples are given the same batch numbers as we got in this tutorial, which will help when we're adding in metadata later!
+    {: .comment}
+> 
+>    > <warning-title>Don't add N701!</warning-title>
 >    > You are adding files to N701, so do not add N701 to itself!
 >    {: .warning}
 >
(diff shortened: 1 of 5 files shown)


topics/single-cell/tutorials/scrna-case_alevin-combine-datasets/tutorial.md
```diff
@@ -168,9 +168,9 @@ You can access the data for this tutorial in multiple ways:
 > 1. {% tool [Manipulate AnnData](toolshed.g2.bx.psu.edu/repos/iuc/anndata_manipulate/anndata_manipulate/0.10.3+galaxy0) %} with the following parameters:
 >    - {% icon param-file %} *"Annotated data matrix"*: `N701-400k`
 >    - *"Function to manipulate the object"*: `Concatenate along the observations axis`
->    - {% icon param-file %} *"Annotated data matrix to add"*: `Select all the other matrix files from bottom to top, N707 to N702`
+>    - {% icon param-file %} *"Annotated data matrix to add"*: `Select all the other matrix files from bottom to top, N702 to N707` (if you imported files from Zenodo, your history may not be in this order, but you can still select them in order from N702 to N707 using the file names)
 >
->    > <warning-title>N707 to N702!</warning-title>
+>    > <warning-title>N701 to N701!</warning-title>
```
bgruening, line 173: those are now the same ...?

topics/single-cell/tutorials/scrna-case_alevin-combine-datasets/tutorial.md
```diff
@@ -168,9 +168,9 @@ You can access the data for this tutorial in multiple ways:
 > 1. {% tool [Manipulate AnnData](toolshed.g2.bx.psu.edu/repos/iuc/anndata_manipulate/anndata_manipulate/0.10.3+galaxy0) %} with the following parameters:
 >    - {% icon param-file %} *"Annotated data matrix"*: `N701-400k`
 >    - *"Function to manipulate the object"*: `Concatenate along the observations axis`
->    - {% icon param-file %} *"Annotated data matrix to add"*: `Select all the other matrix files from bottom to top, N707 to N702`
+>    - {% icon param-file %} *"Annotated data matrix to add"*: `Select all the other matrix files from bottom to top, N702 to N707` (if you imported files from Zenodo, your history may not be in this order, but you can still select them in order from N702 to N707 using the file names)
```
bgruening, line 171: I don't understand the sentence in the brackets? What means selecting in order? They are all selected together, no?

MarisaJL: I've tried to make this clearer in the new version - the point is that they're concatenated in the order you click and as we'll use the batch numbers to add sample metadata, if they're not in the same order as the tutorial it will cause problems!

topics/single-cell/tutorials/scrna-case_alevin-combine-datasets/tutorial.md
MarisaJL: I've changed this now as I think both versions weren't really making the main point to not add N701!

bgruening (approved):

MarisaJL:
@nomadscientist the updates are done now, but I'm not sure what to do with the input histories for F, P, E - should we include 2 or just the old one that works with current tool versions?

nomadscientist:
In general, we should make the tutorials fit their own tool versions (so that's why we put 'tutorial mode' at the top of all of them), otherwise we end up with buckets of text going through every possible combination of other tool versions

nomadscientist (approved):

PR #5338, opened 2024-09-19
match updated Scanpy fa2-modified, add an option to download input data

wee-snufkin:
A new version of Scanpy was released so no more need to change the location of fa2-modified
Also, added a back-up option to download input data using a link


nomadscientist (approved):
Excellent work sorting this Julia, so frustrating that it updated LITERALLY during the course!

PR #5359, opened 2024-09-25
Move to e-learning learningResourceType

hexylena:
all of our materials are e-learning, rather than simply 'material'. They're all available online, constantly, and have both interactive components and self-check points where the learner can be sure they are understanding the material and test their knowledge.

Thus we move from 'hands-on tutorial' to 'e-learning' so we can show up in the appropriate section of TeSS and learners looking for real hands on materials will be able to find us.

### DIFF
--- a/_plugins/jekyll-jsonld.rb
+++ b/_plugins/jekyll-jsonld.rb
@@ -868,12 +868,39 @@ module Jekyll
                         name: "Associated Workflows"
                       })
       end
-      if actual_material.key?('zenodo_link')
+
+      # Notebooks
+      if actual_material.key?('notebook') && actual_material['notebook']['language'] != 'r'
+        # Python, Bash, SQL (all via jupyter)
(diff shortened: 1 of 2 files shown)


bebatut:
Is it possible to have both types? Hands-on tutorial might be useful for instructors to find material. wdyt?

hexylena:
as far as I can tell it's not, (but I could be wrong.)

That said, every material will appear in both the 'materials' and 'e-learning' tab and hopefully as the term "e-learning" becomes more well known, instructors will understand it's meaning as a self-study appropriate "hands on material" as that's part of the definition.

We could also consider writing 'hands on' more explicitly in the description field. 

the [big box at the top with](https://tess.elixir-europe.org/materials/hands-on-for-de-bruijn-graph-assembly-tutorial) 'questions this will address' could include "This is a Hands-on Tutorial from the GTN which is usable either for individual self-study, or use as a teaching material in the classroom"? 

would that maybe achieve the same?

bebatut:
Sounds good!

hexylena:
the updated tess card: 
image

PR #5374, opened 2024-09-27
tool/pipeline wrapping FAQ

hexylena:
Adding some quick FAQs on pipeline/tool wrapping to help colleagues.

This is based on _real world internal experiences_ trying to ship things that are needed on _very tight deadlines_. As such, I won't be accepting criticism that this advice isn't "best practice", yeah, obviously! It's a question of return on investment for certain valid internal use cases. And I think not everyone who uses galaxy knows that there are options if they need them. Suggestions for improving the snippets to make them more easily copyable and to have them approach better practices would of course be welcome!


hexylena:
I would have liked to have finished this for my colleagues, but, lack of time sadly cancels it.

PR #5393, opened 2024-10-02
Ensure pages without short IDs are part of search results

hexylena:
per request from @sujaikumar, now tutorials without short IDs will show up in search results. Additionally workflows are now included as well.


bgruening (approved):
Cool!

PR #5423, opened 2024-10-08
update funding/funders annotation

shiltemann:
Our current split between FUNDERS and ORGANISATIONS was a bit unclear, so I have changed it to GRANTS and ORGANISATIONS.

**What is the difference?**
- Grants should have a funding id and be listed in a grants database we can link to 
    - e.g. CORDIS for EU grants, highergov for US, UKRI for UK ( we can easily add support for others as needed)
- Organisations can be institutes such as universities, but also consortia like ELIXIR, or smaller projects

Other than that we are pretty flexible regarding this distinction 
- Both grants and organisations can be listed as affiliations for contributors
- Both can be listed in as funding contributors in tutorial/slides metadata

**How it is displayed on the GTN?**
- organisations or grants listed as funding contributors in tutorials metadata are listed at end of tutorial in the "Funding" section
- each grant or organisation gets a Hall of Fame page listing
  - all materials for which they have been listed as contributor or funder
  - all individuals who have added the grant or organisation as their affiliation


hexylena (approved):
I can't find anything obvious missing from this, have checked the plugins and corrected every remaining instance I can find. Thanks so much for doing this, I really hope it provides more clarity for contributors!!

PR #5429, opened 2024-10-09
updated deseq2 paramter selection

teresa-m:
update deseq2 parameter setting


pavanvidem:
In the latest version of the DEseq2, the `beta_priors` param was exposed and the default is set to `False`. To get the same results as in the tutorial, this should be set to `False`. In the next tutorial update, we should change it back to its default value and change the tutorial results.

pavanvidem (approved):

PR #5433, opened 2024-10-10
Add metaproteomics per request

hexylena:
cc @PratikDJagtap @bebatut


topics/microbiome/metadata.yaml
```diff
@@ -44,3 +44,6 @@ subtopics:
   - id: metatranscriptomics
     title: "Metatranscriptomics"
     description: "Taxonomic and functional characterisation of mixed samples using transcriptome data."
+  - id: clinical-metaproteomics
```
bebatut, line 47: Can we remove the `clinical`?

bebatut (commented):
Thanks @hexylena

topics/microbiome/metadata.yaml
hexylena: No, we cannot, unfortunately. The subtopic ID must match exactly in both topics, since the tutorials are symlinked in.

shiltemann: users will only see the title field, "Metaproteomics", so maybe ok to leave it?

bgruening: @bebatut lets remove the `clinical` in a follow-up if its too confusing.

PR #5435, opened 2024-10-10
Add in bioinfo bootcamp details retrospectively

nomadscientist:
Would you advise adding anything else here? (esp. any information / metadata that helps you in particular)

Do these events get automatically pulled into the TESS/ELIXIR database, or do I need to add it separately?

### DIFF
--- a/events/2024-09-16-bioinfo_bootcamp-2024.md
+++ b/events/2024-09-16-bioinfo_bootcamp-2024.md
@@ -10,6 +10,8 @@ description: |
 
     We will provide full information on registering for Galaxy and other necessary platforms in the last week in August and will have a drop-in session to troubleshoot any problems on these registrations and any questions you have about the Bootcamp on Monday, 2nd, September from 19:30-20:30.  
 
+    We strongly thank the Galaxy-Europe and Galaxy Training Network for their computational and training resources that make this course possible for the first time.
+
 date_start: 2024-09-16
 date_end: 2024-09-20
 contributions:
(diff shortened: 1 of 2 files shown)


nomadscientist:
I don't actually have a link out for it, as it was an internal course, is that a problem?

shiltemann:
not a problem, but in that case we should change the layout to `event`, not `event-external`, then you just get a GTN event page with any description and info you want :+1:

hexylena:
event-external do not get submitted to TeSS, it's assumed that's handled by the course authors elsewhere. (and reduces the metadata required for those events.)

if it's an `event` layout, then it will be submitted to TeSS, you won't have to do that at all :)

events/2024-09-16-bioinfo_bootcamp-2024.md
```diff
@@ -0,0 +1,48 @@
+---
+layout: event
+title: 2024 Bioinformatics Bootcamp - The Open University
+description: |
```
hexylena, line 4: maybe some of this goes after the `---` at the end, and just keep a shorter description like "The School of Life, Health, and Chemical Sciences (LHCS) at The Open University (OU) is running a free, week-long Bioinformatics Bootcamp"

events/2024-09-16-bioinfo_bootcamp-2024.md
```diff
@@ -0,0 +1,48 @@
...
+location:
+  name:  The Open University - Online
+  city: Milton Keynes
+  country: United Kingdom
+
+tags:
+ - single-cell
+ - introductory
+
+mode: "online"
```
hexylena, line 40:
```suggestion
mode: "online"
async: false
```

shiltemann (approved):

shiltemann (approved):

PR #5456, opened 2024-10-16
Update Helixer tutorial.md - gffread inputs

AnnaSyme:
In the tutorial, for the tool gffread, for `select fasta outputs`, I think the wrong item is specified. 

I have changed it from `fasta file with spliced exons` to `protein fasta file`, because the next step in the tutorial (using Busco) needs this protein fasta file.


bgruening (approved):

bgruening:
Thanks @AnnaSyme

PR #5506, opened 2024-11-07
Update CoDex tutorial and make an FAQ

nomadscientist:
Now that the CoDex holds two resources, we need to specify this, so I've made an overall FAQ for adding a community.

Also, this tutorial belongs in the Community topic, not Dev.


nomadscientist:
@bebatut do you understand why the Testing is failing?

shiltemann (approved):

PR #5512, opened 2024-11-07
Add a third way to run DESeq2 in Ref-based tuto

lldelisle:
Currently there are 2 ways proposed:
- Select each count file for each category (I find it error prone)
- Use tags (this is not easy for new users as it requires to use regex)
I propose a third way where a collection is created and then identifiers are selected using patterns.

### DIFF
--- a/topics/transcriptomics/tutorials/ref-based/tutorial.md
+++ b/topics/transcriptomics/tutorials/ref-based/tutorial.md
@@ -1601,7 +1601,7 @@ DESeq2 requires to provide for each factor, counts of samples in each category. 
 >    - *"How should the elements to remove be determined"*: `Remove if identifiers are ABSENT from file`
 >        - *"Filter out identifiers absent from"*: `Search in textfiles on data XXX` (output of  **Search in textfiles** {% icon tool %})
 >
-> 5. Rename both collections `untreat` (the filtered collection) and `treat` (the discarded collection).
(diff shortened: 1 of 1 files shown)


lldelisle:
Good catch thank you.

topics/transcriptomics/tutorials/ref-based/tutorial.md
```diff
@@ -1575,6 +1575,80 @@ We can now run **DESeq2**:
...
+>    - *"Select lines from"*: `Extract element identifiers on data XXX` (output of  **Extract element identifiers** {% icon tool %})
+>    - *"that"*: `Match`
+>    - *"Regular Expression"*: `untreat`
+>
+> 4. {% tool [Filter collecion](__FILTER_FROM_FILE__) %} with the following parameters:
+>    - *"Input collection"*: `all counts`
+>    - *"How should the elements to remove be determined"*: `Remove if identifiers are ABSENT from file`
+>        - *"Filter out identifiers absent from"*: `Search in textfiles on data XXX` (output of  **Search in textfiles** {% icon tool %})
+>
+> 5. Rename both collections `untreat` (the filtered collection) and `treat` (the discarded collection).
```
lldelisle, line 1604:
```suggestion
> 5. Rename both collections `untreated` (the filtered collection) and `treated` (the discarded collection).
```

hrhotz (approved):

PR #5529, opened 2024-11-12
update training for CYOA if space in option

lldelisle:

PR #5531, opened 2024-11-12
lint cyoa branches

hexylena:
for @lldelisle, xref https://github.com/galaxyproject/training-material/pull/5529  featuring a number of new warnings:

```
$ ruby bin/lint.rb --limit GTN:041
./topics/climate/tutorials/ocean-variables/tutorial.md:1:1:1:1:GTN041 You have non-unique options in your Choose Your Own Adventure. Please ensure that each option is unique in its text. Unfortunately we do not currently support re-using the same option text across differently disambiguated CYOA branches, so, please inform us if this is a requirement for you.We identified the following duplicate options in your CYOA: Options xview, xview became the key: xview; Options png, png became the key: png
$ ruby bin/lint.rb --limit GTN:042
./topics/contributing/tutorials/create-new-tutorial-content/tutorial.md:1:1:1:1:GTN042 We recommend specifying a default for every branch
./topics/ecology/tutorials/marine_omics_bgc/tutorial.md:1:1:1:1:GTN042 We recommend specifying a default for every branch
$ ruby bin/lint.rb --limit GTN:043
./topics/genome-annotation/tutorials/apollo-euk/tutorial.md:1:1:1:1:GTN043 We did not see a corresponding option# for the default: _Create your own organism"_, but this could have been written before we automatically slugified the options. If you like, please consider making your default option match the option text exactly.
$ ruby bin/lint.rb --limit GTN:044
./topics/climate/tutorials/ocean-variables/tutorial.md:1:1:1:1:GTN044 We did not see a corresponding option# for the default: «tools», please ensure the text matches one of the branches.
$ ruby bin/lint.rb --limit GTN:045
./topics/assembly/tutorials/mrsa-nanopore/tutorial.md:1:1:1:1:GTN045 We did not see a branch for Without Illumina MiSeq data (Without-Illumina-MiSeq-data) in the file. Please consider ensuring that all options are used.
```

I think there are valid use cases for GTN:045, but, maybe people should be adding an explicit 'no' branch anyway, even if it is just an empty div?

the linting is not as nice as it could be, it reports the error as being on the first line of the file. But due to needing to use multi-line matches, we don't have a good way to know where the real start is...

### DIFF
--- a/_plugins/gtn.rb
+++ b/_plugins/gtn.rb
@@ -111,7 +111,7 @@ module Jekyll
     #  slugify_unsafe("Hello, World!") # => "Hello-World"
     def slugify_unsafe(text)
       # Gets rid of *most* things without making it completely unusable?
-      text.gsub(%r{["'\\/;:,.!@#$%^&*()]}, '').gsub(/\s/, '-').gsub(/-+/, '-')
+      unsafe_slugify(text)
     end
 
     ##
--- a/_plugins/util.rb
+++ b/_plugins/util.rb
@@ -52,3 +52,7 @@ def markdownify(site, text)
     Jekyll::Converters::Markdown
   ).convert(text.to_s)
 end
+
+def unsafe_slugify(text)
+  text.gsub(%r{["'\\/;:,.!@#$%^&*()]}, '').gsub(/\s/, '-').gsub(/-+/, '-')
+end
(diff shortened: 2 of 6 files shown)


lldelisle:
This is amazing! Thank you so much.

lldelisle:
For GTN045 I would be in favor of an empty div like when you wrap a conditional in a tool.

PR #5589, opened 2024-12-02
Update example history share link slug

jennaj:
Recreated the history in tnabtaf's account. Not sure how it was lost but was able to replace it. The shared history link has a new slug, but still in his account, and everything seems to be the same (minus a deprecated tool version). 

Hi @tnabtaf -- hope this was Ok! 

Reported at https://help.galaxyproject.org/t/guide-extracting-workflows-from-histories/14035/


bgruening (approved):
Ok, that is strange. Do you want to flag this to Nate?

PR #5592, opened 2024-12-04
Add shrutikhare-git as contributor

shrutikhare-git:


CONTRIBUTORS.yaml
```diff
@@ -2247,6 +2247,12 @@ Shorla:
     email: Sogunkelu@gmail.com
     joined: 2021-10
 
+shrutikhare-git:
+    name: Shruti Khare
+    joined: 2024-12
+    email: skshrutikhare@gmail.com
+    orcid: 0000-0001-5622-2663
+
```
nomadscientist, line 2255:
```suggestion
    location:        # add location so we can add you to the world map
      country: SG
```

nomadscientist: Is that where you're joining from? Or are you based somewhere else?

nomadscientist: If you select 'commit suggestion', then we should be able to merge this! @shrutikhare-git

shiltemann (approved):

PR #5606, opened 2024-12-06
[Fix failing deploy] Remove duplicate metadata

shiltemann:
guess we don't lint for that


hexylena (approved):
we don't indeed :( *most* yaml parsers ignore it, nodejs does which is inconvenient. it's not trivial in ruby to do this, we need to inspect the AST read by psych itself, https://stackoverflow.com/a/55705853

PR #5627, opened 2024-12-10
Add the DFG grant to the omero training matherial

rmassei:
Hello, 
small PR to add the DFG grant number to the OMERO training material


GRANTS.yaml
```diff
@@ -233,5 +233,14 @@ skills4eosc:
     url: https://www.skills4eosc.eu/
     github: false
 
+dfg:
+    name:  Deutsche Forschungsgemeinschaft
+    short_name: "DFG"
+    github: false
+    joined: 2024-12
+    funding_id: "501864659"
```
bgruening, line 241: but that is the funding ID of NFDI4Bioimaging, isn't it?

I guess the key should be NFDI4Bioimage, with a DFG logo? The grant is NFDI but the funding agency is DFG.

@shiltemann any advise?

shiltemann: I think what would make the most sense, is to use NFDI4Bioinfmaging as the key, name and avatar here, since that matches the grant id. And then add dfg as the `funding_database` here.

I will have to  make some changes to the framework to support that, but then we can automatically construct the url with the funding id to the DFG page for the grant: https://gepris-extern.dfg.de/gepris/projekt/501864659?language=en.  And then I will also automatically add the DFG logo to any pages referencing grants from DFG (so both nfdi4bioimaging and DFG logos should then automatically appear on the tutorial page). I will try to do all this today.

shiltemann (approved):

shiltemann:
thanks @rmassei!

PR #5629, opened 2024-12-10
Archive answer key histories

nomadscientist:
1. Combined two FAQ and changed the address of any references in the materials to the corrected one
2. Added an 'archive history' FAQ
3. Referenced that in the 'Input and answer key histories' FAQ for making GTN material


faqs/gtn/using-answer-key-histories.md
```diff
@@ -1,20 +0,0 @@
----
```
bgruening, line 1: Isn't that FAQ still useful to have?

nomadscientist: Yes - except this one was better (https://training.galaxyproject.org/training-material/faqs/gtn/gtn_example_histories.html) so I set up a redirect

nomadscientist: Beating software smog! Woo! Unifying two FAQ into 1! Woo!

shiltemann (approved):

PR #5631, opened 2024-12-10
Update answer key histories to be archived

nomadscientist:
Where answer key and input histories were linked in the header, I have imported, shared, updated the link, and then archived them to protect from accidental edits.


topics/single-cell/tutorials/EBI-retrieval/tutorial.md
```diff
@@ -28,13 +28,23 @@ answer_histories:
   - label: "EBI SCXA retrieval tool (UseGalaxy.eu)"
     history: https://singlecell.usegalaxy.eu/u/j.jakiela/h/ebi-scxa---anndata-scanpy-or-seurat-object-1
     date: 2024-01-10
+  - label: "EBI SCXA retrieval tool (UseGalaxy.eu)-ARCHIVED"
```
bgruening, line 31: Should we remove the link to the older history? Isn't that confusing to have two links?

nomadscientist: I was keeping it as a back-up. Also a bunch of these histories I just copied and then archived just in case, so they're actually identical but I didn't want to over-write someone else's work.

bgruening (approved):

PR #5637, opened 2024-12-12
Music update workflows/tools

hexhowells:
Updated workflows for the [MuSiC Deconvolution](https://training.galaxyproject.org/training-material/topics/single-cell/tutorials/bulk-music/tutorial.html) tutorial since the previous workflow didn't fully match the tutorial and the tools used were outdated. This splits the tutorial into two workflows since there are two separate datasets used / tasks being performed. Workflows also utilise frames to indicate which sections of the tutorial the tools are from.

Todo:
- [x] Update the tool versions in the tutorial


mtekman-RBS:
amazing work, thanks for this

mtekman:
amazing work, thanks for this

hexhowells:
I think the main workflow name needs to be updated in ```shortlinks.yaml``` but I'm too familiar with this https://github.com/galaxyproject/training-material/blob/6ed6824690fe018940719999c77d3a5cdbd4a478/metadata/shortlinks.yaml#L1894

shiltemann:
Thanks @hexhowells! We will come up with a better way of handling renamed workflows, but renaming back is definitely the easiest fix here :+1:

shiltemann:
@hexhowells ok I figured out a way to rename workflows and keeping the url persistency.

create a file in our new `redirects` folder with content like:

```
---
redirect_from:
- /topics/mytopic/tutorials/mytutorial/workflows/main_workflow
redirect_to:
- /topics/mytoppic/tutorials/mytutorial/workflows/new_workflow_name
---
```

here is an example: like this one: https://github.com/galaxyproject/training-material/blob/main/redirects/proteomics_neoantigen-2_rename.md?plain=1  (this example doesn,t rename the workflow file itself, but the entire tutorial folder)

So if you want to rename your workflow you can do it now and add such a file, or at least for next time we can do this :)

PR #5648, opened 2024-12-17
Fix event program for small screens

shiltemann:
@teresa-m this should fix your issue for events on small screens:

image

I also updated the event to have a section per day instead of one big table, please check if it is ok for you:

image


teresa-m (approved):
Thanks, the split table looks way nicer!

PR #5652, opened 2024-12-18
fix reports nginx config to avoid localhost issues

martenson:
fixes https://github.com/galaxyproject/galaxy/issues/15966 by applying patch from @jdwhatsup from
https://github.com/jdwhatsup/training-material/commit/c88b7561e36ae8c4c467e0b96d6e41216eed4da5

I tested this on a fairly standard instance with reports under `/reports` and it loads jobs fine


hexylena (approved):
that was all it took? oooooof. thanks for finding this.

hexylena (approved):

PR #5658, opened 2024-12-19
Updated the examples of `cX.count` operations to make them easier to understand.

lybCNU:
Updated the examples of `cX.count` operations to make them easier to understand.


bgruening:
@lybCNU the title and the content does not fit together. The changes also do not look correct. We really wanted to search for "December". Can you elaborate a bit?

shiltemann:
@bgruening I'm guessing the point is that we are currently using the count() function as a contains(), and @lybCNU wanted to show an example where we actually set a higher count.  I'm just not sure that the birthday column is the best for that.

@lybCNU I would suggest leaving the current question as it is, and simply adding a second question/exercise to the box. Then people will have 2 examples to fully understand the `c5.count()` function. 

And maybe instead of the December/September example, how about finding out how many athletes have a 3-part, i.e. by counting spaces in that column with  `c2.count(' ') == 2`? (and bonus question could be finding the athletes with the most-part names by increasing this number, e.g. you can go up to 5 and find several athletes with 6-part names, e.g. "Patricia Galvin de la Tour d'Auvergne")

What do you think?

lybCNU:
@bgruening @shiltemann

Apologies for the delayed response, as it’s been a busy end-of-semester period. Thank you both for your valuable feedback and suggestions!

During this semester, I noticed that quite a few students found the use of `cX.count()` confusing, likely because the original example didn’t fully demonstrate the function’s purpose. This led me to consider replacing it with a more suitable example, which is why I initially proposed the “December/September” scenario to illustrate setting a higher count. However, I realize that this example might not be the best fit.

I really appreciate @shiltemann’s suggestion! Adding a second question or exercise alongside the current one sounds like a great way to help clarify `cX.count()` for students. The example of counting spaces in names to identify athletes with three-part names, as well as a bonus challenge to find those with the longest names, seems both practical and engaging for students.

I’ll update the exercises accordingly. Thanks again for your thoughtful suggestions and support!

bgruening:
Sounds great to me! :) Nice!

lybCNU:
Hi, @bgruening @shiltemann

I have made the updates to the exercises as discussed. Could you please review them when you have a chance?

Thank you!

shiltemann:
Thanks a lot @lybCNU! Looking at the question again in context, I agree that it is perhaps clearer to remove the December example (since we used find() already before which would be the obvious method to solve that question). Instead I have added a couple more counting questions/exercises based on the names. And just some minor formatting changes

shiltemann (dismissed):

shiltemann (approved):

PR #5662, opened 2024-12-19
News layout same as LP layout

shiltemann:
use the same layout as learning pathways for the news index

image

We can tweak further later but already an improvement I think


nomadscientist:
Oh that is much better

nomadscientist:
One day a bar to tick which labels to show perhaps!

nomadscientist (approved):

PR #5699, opened 2025-01-17
Fix Automated video creation GitHub action

shiltemann:
after much trial and error this works now


bgruening (commented):
Outch, thanks a lot @shiltemann for fixing this!

PR #5713, opened 2025-01-25
Natural philosophy: on the origin of science

gallardoalba:
New training about the origin of science.

### DIFF
--- a/topics/introduction/tutorials/origins_science/tutorial.md
+++ b/topics/introduction/tutorials/origins_science/tutorial.md
@@ -15,9 +15,9 @@ contributions:
 
 # Overview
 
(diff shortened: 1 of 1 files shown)


gallardoalba:
I don't know if this is the correct section @shiltemann @hexylena. Also I don't know if the training fits the Galaxy community at all. It is a original text I wrote about how science emerged in ancient Greece. If you consider it doesn't fit, I will close it. The content is complete, just need to fix the format and include some images, so if you are interested to review, would be very great.

shiltemann:
Hi @gallardoalba, thanks for you contribution! This isn't really a tutorial as much as an essay, and probably a bit out of scope for the GTN, but how would you feel about adding it as a blog (news) post?

PR #5774, opened 2025-02-13
Add metadata to understanding barcodes

nomadscientist:
Add prerequisites & follow-up training


nomadscientist:
Wow nice work solving that error!

shiltemann:
@nomadscientist I hijacked your PR to fix the build error that has been popping up everywhere. This should fix it, so if you see the "Document tree depth exceeded" error anywhere, just hit the "update branch" button to pull in this change.

(error happened because a zipped html file was added as part of workflow testing data and the link checker tried to parse it. the test should now ignore any files in test-data folder)

shiltemann (approved):

PR #5781, opened 2025-02-14
Importing From Public atlas - Tool Updates

hexhowells:
Updates the tool versions in both the tutorial and attached workflow


bgruening (approved):
Thanks!

PR #5782, opened 2025-02-15
Fix galaxy installation tutorial: add raw tags to ensure template not evaluated

jdavcs:
Surround value with raw tags; otherwise template is evaluated, resulting in this:
```
galaxy_user: {name: "", shell: /bin/bash}
```
Should be:
```
{name: "{{ galaxy_user_name }}", shell: /bin/bash}
```
(username cannot be empty)


shiltemann (approved):
nice catch, thanks a lot @jdavcs!

PR #5800, opened 2025-02-26
Swich from FastQC to falco in Quality control tutorial

VerenaMoo:
To implement Falco instead of FastQC I rewrote the tutorial to use Falco. Since the images for specific examples were generated with FastQC and I could not find any corresponding datasets, I could not replace them with plots from Falco. Therefore, I left them as they are and the rest of the tutorial shows both plots from FastQC and Falco for the dataset used and then describes any differences.


wm75:
Thanks @VerenaMoo for your work!

@shiltemann I'm not 100% sure what's the best way of doing this. Many people coming to this tutorial may have seen fastqc plots before or will encounter them later somewhere and that's why we thought the side-by-side comparison of fastqc and falco plots might actually be a good thing on its own.

I'm going to review the current version tomorrow I hope and maybe I will change my opinion then. At the same time, @VerenaMoo only has limited time to spend on this tutorial and I'm not sure whether a Choose your own adventure rewrite is realistic with the GTA coming up, but I'll definitely think about it some more.

shiltemann:
@wm75 I understand completely. If there is no time for CYOA, I would add least instruct the users to run both tools rather than only Falco, so that they can compare the outputs in their Galaxy histories as well. Since we do also get a lot of learners who have not used FastQC before.

We can always turn it into a CYOA at a later stage then

But yes, have a look and see what you think is best, I will go with your judgment on this :+1:

dadrasarmin:
Hi everyone,

I was wondering what is the current decision with this PR. @teresa-m asked me and Daniela to check the introduction track for GTA2025 and this is one of them. I wanted to update the tutorial and actually I was thinking about replacing FastQC with Falco as well. I am glad to see it has been done already.

If this is the new version, I think we need a new YouTube recording and appreciate it, if we can merge the PR a couple of weeks before GTA2025 so someone hopefully could record the video.

No matter whether we want to use FastQC or Falco or both, I would be glad to add `answer_histories` to metadata of tutorials that I am checking and update the workflows with to the latest version of each tool.

wm75:
I think this is too big a change for GTA2025, the user-experience hasn't been tested and @shiltemann has raised some valid points. I would prefer to finish this after the training academy. Thanks for your help though @dadrasarmin!

PR #5804, opened 2025-02-27
Update request-galaxy-tools-on-a-specific-server.md

nomadscientist:
AU has a specific form now

Would be good to know if the process is different in EU or ORG now


shiltemann (approved):
neat! thanks @nomadscientist

PR #5822, opened 2025-03-07
Update request-galaxy-tools-on-a-specific-server.md

teresa-m:
Not sure if there was a reason why the EU and US URLs were not a link? I think it is better if people can just click. 

@nomadscientist: Thanks a lot for the faq.


nomadscientist (approved):

nomadscientist:
ace!!

Issue #5762, opened 2025-02-13
Double place a tutorial


nomadscientist:
There's this single cell proteomics tutorial that's in proteomics (https://training.galaxyproject.org/training-material/topics/proteomics/tutorials/bioconductor-scp/tutorial.html) but should also be in single-cell (it's single-cell proteomics).

Is there a way it can exist in both? Some sort of fancy redirect so it gets in the tutorial listing but then on clicking goes to the place where the tutorial page exists?

shiltemann:
I think the easiest way is to just create a symlink from one topic to the other

shiltemann:
symlinks won't work when subtopics or priority is used, need a more sophisticated approach

shiltemann:
this is now supported, with cross_topic_tutorials in the topic metadata file

PR #5834, opened 2025-03-11
Multiple locations for single-cell proteomics

nomadscientist:
Re: https://github.com/galaxyproject/training-material/issues/5762

The single-cell proteomics tutorial should be in proteomics and single-cell. Here's an attempt at doing that!


nomadscientist:
OK, so I think the problem might be how / where to place the tutorial. If I put the metadata in the index file on subtopic (multiomics) and priority, that's not getting picked up.
However, I can't put it in the main file either, or that will mess up its own priority system.
Help please 🙏🏾

shiltemann:
I think we just create a symlink from the tutorial in one topic to the other

nomadscientist:
Did you message me an idea at one point on needing to make this into a feature? Should I convert this into an issue instead, then?

shiltemann:
@nomadscientist yes its going to require a bit more work to make work, given things like priority and subtopics that may be different in the different topics, so converting to an issue makes sense

shiltemann:
guess we already have an issue for it: https://github.com/galaxyproject/training-material/issues/5762 

going to close this until we have a good solution, then re-implement this

PR #5842, opened 2025-03-12
Onboarding learning pathway

nomadscientist:
Here's a collaboration between @PhilReedData  and myself on onboarding new contributors!

I left in a few placeholders for the future, FYI, but I think it works reasonably well as it stands.

- [ ] There are FAQ in the learning pathway, idk if this will render

I also deleted my one on building training material, because I think this one is just better.


nomadscientist:
OK, we still need help with the FAQ, but there are two other relevant PRs that need merging and then we can put results into this one

nomadscientist:
OK, I've updated this based on the relevant PRs that are now merged!

- [ ] Key thing is - I don't think we can add an FAQ into a learning pathway, which is a shame. It would be nice to have it in as basically a link and then if they click it it takes them to the FAQ page (rather than a drop-down snippet).
- [ ] The FAQ wrecks the time estimation (I suspect) so having it ignore any materials that don't have a time would be good, or being able to turn the time estimation off? (time_estimation: false)

shiltemann:
@nomadscientist FAQs are now supported in pathways, and the time estimate will just ignore them

image

shiltemann (approved):

PR #5869, opened 2025-03-19
Update Jbrowse parameter

abretaud:
Updated a few tutorials that were using an obsolete JBrowse option


shiltemann:
thanks @abretaud!

PR #5872, opened 2025-03-19
add training comparison annotation

rlibouba:
Hello, 
I've written a tutorial to compare annotation between two annotation tools that differ in their approach and techniques: Helixer and Braker3. 

Thank you! Have a nice day!
Romane


rlibouba:
Hello @bgruening  , I have integrated jbrowse in the tutorial

PR #5873, opened 2025-03-19
Change from fastqc to falco in ENA_Biodiv_submission

VerenaMoo:
I changed this tutorial to use Falco instead of FastQC.


wm75 (approved):
Thank you!

PR #5879, opened 2025-03-20
Case Study 1 - Tutorial update

nomadscientist:
Massive re-write

Add mito labelling step
Update tools
Add CYOA features
Update workflow history
Drastically simplify (run Alevin only once, etc.)


nomadscientist:
I have moved this to  https://github.com/galaxyproject/training-material/pull/5906 because there were too many clashes

PR #5909, opened 2025-03-27
update CODEOWNERS file

shiltemann:
I realized our CODEOWNERS file was very outdated (missing/renamed topics etc)

The team based approach was not working for us, so I've written down all editorial board members individually. 

So all editorial board members should get notified about any PRs concerning their topic.

Everybody is free to add/remove themselves to parts of this file as they want

I also made a section on the editorial-board page that autogenerates the codeowners from metadata, so that it is easy to update in the future.


bgruening:
I see this:

grafik

Should we remove them?

shiltemann:
I saw as well, they all exist, I think we just never added them as collaborators. So I did that now.

bgruening (approved):

bgruening:
Thanks!

PR #5936, opened 2025-04-07
Update GRANTS.yaml

Sch-Da:
Started adding de.kcd by request of @bgruening. 

TODO: @bgruening can you please fill out the remaining info or let me know, where to find it, thank you!

### DIFF
--- a/topics/microbiome/tutorials/metatranscriptomics-short/tutorial.md
+++ b/topics/microbiome/tutorials/metatranscriptomics-short/tutorial.md
@@ -23,16 +23,17 @@ key_points:
 - A lot of metatranscriptomics sequences are identified as rRNA sequences
 - With shotgun data, we can extract information about the studied community structure and also the functions realised by the community
 - Metatranscriptomics data analyses are complex and must be careful done, specially when they are done without combination to metagenomics data analyses
-contributors:
-- pratikdjagtap
-- subinamehta
-- jraysajulga
-- bebatut
-- emmaleith
-- pravs3683
-- shiltemann
-- paulzierep
-- EngyNasr
+contributions:
+  authorship:
+  - pratikdjagtap
+  - subinamehta
+  - jraysajulga
+  - bebatut
+  - emmaleith
+  - pravs3683
+  - shiltemann
+  - paulzierep
+  - EngyNasr
 redirect_from:
 - /topics/metagenomics/tutorials/metatranscriptomics-short/tutorial
 edam_ontology:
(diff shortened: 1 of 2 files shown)


GRANTS.yaml
```diff
@@ -67,6 +67,17 @@ CINECA-Project:
     url: https://www.cineca-project.eu
 
 
+dekcd:
+    name: de.kcd
+    github: https://github.com/deKCD
```
shiltemann, line 72: you can remove this line, by default it will assume that the key at the top of the entry (in your case `dekcd`) is also the github name (capitalization doesnt matter, so should work by just deleting this line)

GRANTS.yaml
```diff
@@ -67,6 +67,17 @@ CINECA-Project:
...
+    joined: ?
+    avatar: ?
```
shiltemann, line 74: similarly, avatar will by default be the same as the github image, so you can leave this line out unless you want to use a different image than the github profile picture

GRANTS.yaml
```diff
@@ -67,6 +67,17 @@ CINECA-Project:
     url: https://www.cineca-project.eu
 
 
+dekcd:
+    name: de.kcd
+    github: https://github.com/deKCD
+    joined: ?
```
shiltemann, line 73: This is used to indicate when they were added to the GTN, so can just set it to this month:

```suggestion
    joined: 2025-04
```

shiltemann (commented):
Thanks @Sch-Da! 

Just to makes sure: did you receive a specific grant from or for this project that you want to acknowledge? 

If you just want to list de.KCD as contributing/supporting organisation without a specific grant, it should go in ORGANISATIONS.yaml file

Sch-Da:
Thanks for the explanation @shiltemann ! Again, I am amazed about the efficiency enabled by your embedded GTN structures!
@bgruening wanted to add a specific dekcd grant that we already got in the past but forgot to add. Then, I want to add them to my affiliations. But hopefully he can otherwise correct this

GRANTS.yaml
shiltemann, line 72:
```suggestion
```

GRANTS.yaml
```diff
@@ -67,6 +67,17 @@ CINECA-Project:
     url: https://www.cineca-project.eu
 
 
+dekcd:
+    name: de.kcd
+    github: https://github.com/deKCD
+    joined: 2025-04
+    avatar: ?
```
shiltemann, line 74:
```suggestion
```

PR #5941, opened 2025-04-08
OpenRefine Tutorial Update pt 2

Sch-Da:
I updated part two of the OpenRefine tutorial, which is a great and insightful tutorial. This will be very helpful in introducing people to the program's functionalities in Galaxy. Thanks a lot to the authors!

I made some minor changes in consistency and typos.
I added small steps, particularly when starting and ending OpenRefine, to explain the transition from the platform to the interactive tool and back.
There was an error in the API section, as the link given in the "API use" section should be
"http://api.gbif.org/v1/species/match?verbose=true&name="+escape(value,'url')
or it does not work.
 
I think the API section at the very end, after the main file has already been exported back to Galaxy, is a bit confusing. However, the order seems necessary for the original course. To make the transition clearer, I added the part explaining that this part again takes place in OpenRefine.

TODO:
There is an error in some headlines I cannot explain. for example in line 212 of the raw file
In the HTML, it shows for example
"Hands On: hands_on Faceting and duplicates" instead of "Hands On: Faceting and duplicates "
This appears several times from the "Faceting" part onwards.

@bgruening is my closing of OpenRefine, as I explained here, best practice? I was a bit irritated by the red job afterwards.

### DIFF
--- a/topics/ecology/tutorials/openrefine_gbif/tutorial.md
+++ b/topics/ecology/tutorials/openrefine_gbif/tutorial.md
@@ -96,7 +96,7 @@ This tutorial is based on the GBIF {% cite GBIFtraining2021 %} "Biodiversity dat
 >    - click the button "run tool"
 > You will then be redirected. A notification "Interactive Tools" shows up on the left-hand side, indicating you have an interactive tool running.
 > Click on the notification, click on the empty box in front of the text saying "Openrefine".
-> Then, click on the text "Openrefine" to open the tool in a new tab.
+> Then, click on the text "Openrefine" to open the tool in a new tab. It might take a moment until your cursor changes and the text becomes a hyperlink.
 >
 > 2. Open the OpenRefine instance
 >    - go to "Open Project" space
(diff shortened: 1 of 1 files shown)


topics/ecology/tutorials/openrefine_gbif/tutorial.md
```diff
@@ -251,7 +252,7 @@ kind of text, numbers and dates.
 
 > <hands-on-title> {% icon hands_on %} Basic clustering </hands-on-title>
```
shiltemann, line 253: This will fix the issue with the box headers that you mentioned, somebody tried to add an icon there but that does not work. So remove this everywhere you see it to fix that. Thanks for noticing!

```suggestion
> <hands-on-title> Basic clustering </hands-on-title>
```

yvanlebras (approved):
Thank you Daniela for ameliorating so much the tutorial and cleaning so many errors!!!!

Sch-Da:
Thank you @yvanlebras - I learned a lot through your tutorial and really enjoyed getting to know OpenRefine like that!!

Issue #6001, opened 2025-04-25
Installation failure of Ansible playbook due to deprecated lib2to3 module


rhiniodontypus:
### ModuleNotFound error during playbook run
After updating to the latest Galaxy release (v24.2, `galaxy_commit_id: release_24.2`) in the `./galaxy/group_vars/galaxyservers.yml` config file, the [Galaxy Admin tutorial playbook](https://training.galaxyproject.org/training-material/topics/admin/tutorials/ansible-galaxy/tutorial.html) (Revision: 102, Last modified: Apr 8, 2025) deployment fails on Ubuntu 24.04 LTS with the following error:
```
TASK [galaxyproject.galaxy : Get current Galaxy DB version] *****************************************************
fatal: [localhost]: FAILED! => changed=false
  cmd:
  - /srv/galaxy/venv/bin/python
  - /srv/galaxy/server/scripts/manage_db.py
  - -c
  - /srv/galaxy/config/galaxy.yml
  - db_version
  delta: '0:00:00.765326'
  end: '2025-04-25 18:07:44.262573'
  failed_when_result: true
  msg: non-zero return code
  rc: 1
  start: '2025-04-25 18:07:43.497247'
  stderr: |-
    Traceback (most recent call last):
      File "/srv/galaxy/server/scripts/manage_db.py", line 12, in <module>
        from galaxy.model.migrations.scripts import LegacyManageDb
      File "/srv/galaxy/server/lib/galaxy/model/__init__.py", line 146, in <module>
        from galaxy.files.templates import (
      File "/srv/galaxy/server/lib/galaxy/files/__init__.py", line 16, in <module>
        from galaxy.files.sources import (
      File "/srv/galaxy/server/lib/galaxy/files/sources/__init__.py", line 37, in <module>
        from galaxy.util.template import fill_template
      File "/srv/galaxy/server/lib/galaxy/util/template.py", line 28, in <module>
        from lib2to3.refactor import RefactoringTool
    ModuleNotFoundError: No module named 'lib2to3'
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
ModuleNotFoundError: No module named 'lib2to3'
```
---
### Workaround
After manually installing
```
sudo apt install python3-lib2to3
```

the playbook is running the deployment without further issues.

---
### Root Cause

[...]

PR #6005, opened 2025-04-27
Update tutorial: Fix of installation failure of Ansible playbook due to deprecated lib2to3 module

rhiniodontypus:
This PR implements the changes proposed in issue #6001 to fix the installation failure of the current [Ansible playbook template](https://training.galaxyproject.org/training-material/topics/admin/tutorials/ansible-galaxy/tutorial.html) due to the depricated Python lib2to3 module (Python >= 3.11) on Ubuntu 24.04 LTS. The main changes are:

- Added a Python and platform version check to the pre-tasks section of the galaxy.yml playbook template to install the lib2to3 package if necessary
- Updated the role versions in requirements.yml, and adjusted the Galaxy release and Miniconda versions in group_vars/galaxyservers.yml

Please review and let me know if further changes are needed.

### DIFF
--- a/topics/admin/tutorials/ansible-galaxy/tutorial.md
+++ b/topics/admin/tutorials/ansible-galaxy/tutorial.md
@@ -312,15 +312,15 @@ We have codified all of the dependencies you will need into a YAML file that `an
 >    @@ -0,0 +1,13 @@
 >    +# Galaxy, Postgres, Nginx
 >    +- src: galaxyproject.galaxy
->    +  version: 0.10.14
+>    +  version: 0.11.2
 >    +- src: galaxyproject.nginx
->    +  version: 0.7.1
+>    +  version: 1.0.0
 >    +- src: galaxyproject.postgresql
->    +  version: 1.1.2
+>    +  version: 1.1.8
 >    +- src: galaxyproject.postgresql_objects
->    +  version: 1.2.0
+>    +  version: 1.2.1
 >    +- src: galaxyproject.miniconda
->    +  version: 0.3.1
+>    +  version: 0.3.2
 >    +- src: usegalaxy_eu.certbot
 >    +  version: 0.1.11
 >    {% endraw %}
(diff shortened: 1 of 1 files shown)


hexylena:
Thanks for the contribution @rhiniodontypus! And welcome to the GTN :) (Please consider adding yourself to the CONTRIBUTORS.yaml, and then also to the tutorial metadata as tester perhaps?) 

Onto the content: the changes to the versions are significant, and it would require a lot of testing (there's some 15+ GAT modules downstream of this that would also need to be re-validated). We want to keep the training materials (as much as possible) in a known working state, so we generally don't update those versions unless they're running a admin training course (when other material updates happen, and testing is easier.)

If you could reduce this PR to just the pre-task that's required, then it would be much easier to merge in a short timeframe.

Otherwise, I'm not sure if there's any of the admins with time to spin up a VM, and run through each stage of the playbook + at minimum the automated testing that's available from the command line. (I wrote the 'automated' tests but I don't think any of the admin maintainers has ever run it themselves :( and I have very limited involvement currently.)

rhiniodontypus:
Many thanks for your feedback! I totally understand the dimension of the changes.
The problem with the old settings (Galaxy release v23.0) is that it throws the error

error_galaxy_23 0_pkg_resources__deployment

This doesn't seem to be easily fixable because it occurs although the ```python3-pkg-resources``` package was already installed manually and therefore available at the time of deployment. The same is true for Galaxy release v.23.1.

Galaxy release v.23.2 generates a ```ModuleNotFoundError: No module named yaml``` error although ```python3-yaml``` was available at the time of deployment.

Galaxy releases v.24.0, 24.1, and 24.2 all are in conflict with the default miniconda v.23.9 setting:

error_galaxy_24 0_miniconda__deployment

Deploying the v.24 releases with the latest miniconda v25.1, however, work (if ```python3-lib2to3``` is installed via apt).

So at the moment it is not possbile to deploy the current playbook template settings on Ubuntu 24.04 LTS without customizing the Galaxy release version and the miniconda version.
For now, I would try to continue the training with the updated versions and see if I run into any issues as the tutorial progresses. Otherwise, I will ask for a VM running Ubuntu 22.04.

Would you like me to close the PR until further notice?

PR #6012, opened 2025-04-28
update Calling variants in non-diploid systems workflow

khaled196:
Update Calling variants in non-diploid systems workflow for Galaxy Training Academy.


khaled196:
Hi @shiltemann 

Can you please help me with the error, as it does not show why it fails?

shiltemann:
@khaled196 I will have a look :+1: 

In the meantime, the test-data folder is 426 MB large, can you look into reducing that (a lot)? maybe use smaller dataset, and use asserts as tests where the files are big etc?

shiltemann:
thanks! that is already a lot better, but still 86MB, can we subsample the files to just 100 reads orso? The test files do not need to match the tutorial, but can be anything just to confirm the workflow is functional. Thanks!

shiltemann (approved):

shiltemann:
@khaled196 thanks a lot!

PR #6018, opened 2025-04-29
updates based on Lisannas comments

teresa-m:
Updates to the Jupyter Notebook tutorial for the GTA based on Lisannas Comments:

1. When opening this specific Jupyter tutorial, the learner may not know how to start Galaxy at all. The tutorial starts with “create a new history” but could also maybe link to use galaxy.eu and the other Galaxy instances.
2. The tool icon / button 🔧 JupyterLab doesn’t work, I don’t know if they usually do - I had this expectation. 
3. The fact that the box with instructions for installation was below the one to import modules was counterintuitive to me.
        - Also, I have a “not found” error while trying to install ‘math'
        - Importing ‘seaborn’ also results in an error
4. This instruction was quite confusing to me: “Find the file_numbers of all of the gapminder datasets (gapminder_all.csv, ...) with your galaxy history” - does it mean “in” my Galaxy history? Maybe better to add a screenshot to show where exactly?
        In general, I don’t understand how the function get() works. I would suggest more text explaining it.
        
My updates:
1. I added the history tutorial as a requirement. I hope on the first day they will anyway do the introduction to Galaxy tutorials. But if they can just do the history tutorial before starting with this one.

2. It should be clickable if you open the tutorial in an Usegalaxy.* instance. From the GTN homepage, it will be not. I will try to remember to show this feature in an introduction video.

3. I switched the hands-on box order for installation and import

-> Can you check the errors you got with the install and import? I could not reproduce them.

4. I added a bit of information and a screenshot.


shiltemann (approved):

teresa-m:
I will update this PR as I saw that the version of the jupyter lab changed and then the Interface is again different. 
Please wait for merging.

teresa-m:
@shiltemann there is still a data-library.yaml pointing to the original input data of this tutoral. However, we are now using the Caprenites input (on Zenodo but not the data lib). Should I do something about this file?

shiltemann:
If you update the Zenodo link, the data library on EU should update automatically

teresa-m:
> If you update the Zenodo link, the data library on EU should update automatically

Yes, I did update the link in the tutorial. So I will delete the data-library.yaml and we see if it gets automatically generated.

shiltemann:
thanks @teresa-m!

PR #6019, opened 2025-04-30
Switch to production WorkflowHub

shiltemann:
The GTN workflows are currently automatically pushed to the development instance of WorkflowHub, this changes that to target production WFH


shiltemann:
ping @bgruening

supernord:
Thanks @shiltemann

Are you happy with the metadata that has been submitted to the WorkflowHub dev instance? If so, I think it’s ok to merge.

Please let us know if you notice any issues with the final workflow entries on the production instance of WorkflowHub.

@bgruening @fbacall

shiltemann:
@supernord thanks for having a look, we might improve the metadata in the future, but good enough to start with from our side. So if it is also good from your side then let's merge this :+1:

PR #6033, opened 2025-05-06
Update training assembly

scorreard:
Did a list of changes following live training :
- Folder name changed (from flye-assembly to hifi-assembly)
- Time estimation : Reduced the estimated time to 3hr
- Quast : Reordered the quast with reference genome and the quest without
- Quast with reference genome : Removed the reference genome as an input (as it is used as the reference)
- BUSCO : updated version
- BUSCO : Suggested to use the cached lineage
- BUSCO : Changed the lineage (and updated the results)


shiltemann:
thanks a lot @scorreard! Can I ask why you changed the folder name?

scorreard:
Hi @shiltemann! The folder name was flye-assembly, as flye was the assemble used in the previous training. I updated the training so that users have the choice between 2 assemblers (flye and hifiasm), so it's not just a flye-assembly tuto, hence the change.
HiFi is the read type, not the assembler name, so it fits both assemblers

shiltemann:
thanks for explaining, makes sense, one question:

Did you intend for this tutorial to replace the original one, or do you think we should keep both separately? From your explanation it sounds like it should just replace the previous one right, since now you just get an additional choice of tool?

If that is the case, could you please:

1. Delete the `flye-assembly` folder
2. In the metadata of your tutorial, add
   ```
   redirect_from:
     - topics/assembly/tutorials/flye-assembly/tutorial
   ``` 
   This will make sure that people who had bookmarked the original URL will still end up at the right tutorial, even though we renamed the folder.

   (and we will need to do something similar for the `workflows/index.md` and `faqs/index.md` files)

scorreard:
Yes, the idea is to replace the original one as they now have the choice (the original one being one of the choice).

I deleted the folder and added the redirect in the tutorial metadata.

Should I do something in the workflow and faqs even if they are empty?

Thanks!!

shiltemann:
Thanks!

For the other redirects:

`workflows/index.md file`

```
---
layout: workflow-list
redirect_from:
- topics/assembly/tutorials/flye-assembly/workflows/index.md
---
```

(and similar for `faqs/index.md` file)

For the `main_workflow.ga` workflow file itself, its a little bit different, but you can add a file similar to [this one](https://raw.githubusercontent.com/galaxyproject/training-material/refs/heads/main/redirects/text-mining-chinese.md) inside the `redirects/` folder in the root of the GTN

scorreard:
Great, this should do it?

shiltemann:
looks good to me (the tests will let us know if not), thanks!

shiltemann:
This looks good from the technical side, perhaps @abretaud would like to review the content?

redirects/hifi-assembly-renamed.md
```diff
@@ -0,0 +1,6 @@
+---
+redirect_from:
+- /topics/assembly/tutorials/flye-assembly/workflows/main_workflow.ga
+redirect_to:
+- /topics/assembly/tutorials/hifi-assembly/workflows/main_workflow.ga
+---
```
shiltemann, line 6:
```suggestion
redirect_from:
- /topics/assembly/tutorials/flye-assembly/workflows/main_workflow
redirect_to:
- /topics/assembly/tutorials/hifi-assembly/workflows/main_workflow
---
```

shiltemann:
Thanks for having a look @abretaud, all green checkmarks now, so merging, thanks a lot @scorreard!

PR #6042, opened 2025-05-07
[Event] Add DHLab workshop as external event

shiltemann:
@Sch-Da this adds your upcoming workshop also to the GTN events page :)


Sch-Da (approved):
Thank you very much for setting the event up here, Saskia!

PR #6043, opened 2025-05-07
Proteogenomics 1 - fix regex for merging fasta files

Delphine-L:
The previous regex for identifying duplicates when merging fasta files was not picking up on the full accessions.


shiltemann:
thanks @Delphine-L!

Issue #6037, opened 2025-05-06
Help for Climate tutorials


teresa-m:
It would be great if anyone could help with the following issues of the two Climat track tutorials.

For the climate 101 (https://training.galaxyproject.org/training-material/topics/climate/tutorials/climate-101/tutorial.html#climate-variables), I had issue getting data from Copernicus but I prepared a dataset in Zenodo (https://doi.org/10.5281/zenodo.15349982)   but the visualisation tool (https://github.com/NordicESMhub/galaxy-tools/tree/master/tools/psy-maps) fails. I think it would require old version of Python and psy-maps, etc. I tried to fix it but did not manage.

For Pangeo ecosystem 101 for everyone, it fails at https://training.galaxyproject.org/training-material/topics/climate/tutorials/pangeo/tutorial.html#select--subset-from-coordinates. Some Xarray tools changed and I did not manage to find what was wrong.

Please contact @annefou if you can help out!

teresa-m:
As you prefer. We can also keep this open, and you just update on the tasks that are still open. See what feels better for you.

PR #6054, opened 2025-05-08
fix Getting your hands on climate

annefou:
use data from Zenodo and Panoply to visualize Essential Climate Variables

(partly address #6037 e.g. Climate 101)


shiltemann (approved):

shiltemann:
Thanks a lot @annefou!

PR #6066, opened 2025-05-11
add FAIR2Adapt to GTA as funding contributor

shiltemann:
For the climate track


teresa-m:
@shiltemann thanks a lot!

PR #6093, opened 2025-05-14
updated FAQ on collection

teresa-m:
Update of the FAQ.
@paulzierep it is according to the question on GAT Slack.


teresa-m:
I have a question related to the data library. I thought I should find the data under `GTN_material -> Microbiome`. But I don't see a `Microbiome` folder.

shiltemann (approved):

PR #6112, opened 2025-05-20
update contributors for Teresa and Lisanna

bgruening:
update contributors for Teresa and Lisanna


teresa-m (approved):
@bgruening looks great, thanks

PR #6126, opened 2025-05-27
new phylodiversity workflow tutorial

Morphy427:
A tutorial for the phylodiversity workflow


yvanlebras:
Did I miss something? ;) Seems to me you closed both PR but one would be ok normally isn't it?

PR #6142, opened 2025-06-02
Fix "access_type is too permissive" on WorkflowHub submission

fbacall:
Fixes this issue: https://github.com/galaxyproject/training-material/actions/runs/15389089921/job/43294210078#step:8:1030

which was due to WorkflowHub not handling partial updates of the policy very well. The submission script tried to add a "manage" permission for members of the GTN Team on WorkflowHub, but the WorkflowHub API also requires the public access type to be declared in the request.

(FYI I fixed the policies of the GTN workflows server-side, so future submissions should succeed regardless of this PR)

### DIFF
--- a/bin/wfh-upload.py
+++ b/bin/wfh-upload.py
@@ -115,6 +115,9 @@ def doUpload(crate_path):
     #
     push = False
     if updated_policy:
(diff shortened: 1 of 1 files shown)


shiltemann (approved):

shiltemann:
thanks a lot @fbacall!

PR #6157, opened 2025-06-03
Add EDAM to hifi assembly tutorial

scorreard:
During the Elixir All Hands, EDAM were suggested for this tuto


scorreard:
This is an outdated file, my branch was not up to date, IGNORE

PR #6166, opened 2025-06-03
Added a FAQ for exporting citations of a history.

dadrasarmin:
I added a FAQ to the feature section to explain how a user can get the list of all publications they should cite based on the history of their analysis.


shiltemann (approved):
nice, thanks @dadrasarmin!

PR #6179, opened 2025-06-10
Removed the biodiversity tag from the tuto

scorreard:
During the All Hands, a user mentioned that about this tutorial "Very old tutorial, should be updated or removed"

Removing the biodiversity tag will avoid having it in the biodiversity codex, or should we just remove it completely?


bebatut:
The linting fails:

```
./metadata/genome-annotation.yaml has errors
  [/edam_ontology/0] 'operation_0362': not matched to pattern /^topic_[0-9]+$/.
```

Could you fix that in this PR?

PR #6199, opened 2025-06-19
Start a map-over tutorial

bernt-matthias:
Here I started a tutorial o how Galaxy processes dataset(s). I think it's important that users understand. It also should  describe  how to map over for tools that process multiple datasets at once. 

As far as I can see this was missing so far. 

In the near future I won't be able to actively work on this. But I would be super happy about any help that I could merge here :) 

We could also think to integrate this here: https://training.galaxyproject.org/training-material/topics/galaxy-interface/tutorials/collections/tutorial.html .. but to me it seemed not to fit well.

### DIFF
--- a/topics/galaxy-interface/images/mapover/build_list.svg
+++ b/topics/galaxy-interface/images/mapover/build_list.svg
@@ -1,0 +1,1344 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   version="1.1"
+   viewBox="0 0 300.73563 184.43085"
+   fill="none"
+   stroke="none"
+   stroke-linecap="square"
+   stroke-miterlimit="10"
+   id="svg1545"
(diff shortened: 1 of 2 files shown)


bernt-matthias:
> The initiative is great, but maybe we can first review missing pieces in Galaxy itself ? 
> This is such a central concept that maintaining this information externally to Galaxy seems like giving up in a way.

Good points. Seems that I forgot about these docs :) So this should be more a tutorial. We could feed in some tool examples and point the users to the existing docs in Galaxy. 

> For instance your explanation for how to build higher order collections is not the easiest way to do this anymore (starting with 25.0 you have a wizard for all types of collection), again highlighting how that documentation should be shipped with Galaxy.

For existing collections one will still need the rule builder, or? I could not find a way in the wizard for nested list to auto create the identifiers for the inner lists. I guess a tutorial for the rule builder might still be good.

PR #6207, opened 2025-06-24
Remove dead link from DH tutorial.bib

Sch-Da:
The link to the newspaper article mentioned in my tutorial is no longer accessible; therefore, it has been deleted.  unnecessary personal tags from another publication were also removed

Thanks for your help in updating :)


bgruening:
We need a DOI, URL or ISBN in every entry.

Sch-Da:
Found a replacement, so it's up to date again and has a URL :)

bgruening:
I'm not sure why it complains now, but "@thesis{Schneider2024," also is missing a URL, DOI or ISBN.

@shiltemann can we ignore those warnings then?

PR #6217, opened 2025-06-30
Editing numbers in neoantigen tutorials

katherine-d21:


bebatut:
@subinamehta: There are issues that were introduced by the partial renaming (naming and not the files themselves) in this PR (check the linting error). Could you look into it? Thanks a lot.
For the time being, I will revert it as it impacts follow-up PRs.

shiltemann:
Thanks @bebatut! 

@subinamehta I will take care of the open PR to do the renaming, please be careful with merging when the tests fail, just ping me in the future if you need a hand with the linting errors!

PR #6222, opened 2025-07-01
Update funding in SnapATAC2 tutorial.md

Sch-Da:
Adding dekcd as funding for this, as @bgruening proposed


shiltemann (approved):
thanks @Sch-Da :+1:

PR #6229, opened 2025-07-02
updating foodborne tutorial workflows

EngyNasr:
Two of the tutorial workflows used "filtered failed datasets" tool which is now updated to have an extra optional input, so I updated these workflows with the new filtered failed datasets tool to avoid tutorial's doers to deal with the importing error


bebatut:
Does that change the tutorial itself?
Btw, there seems to be conflict for the workflow files

EngyNasr:
> Does that change the tutorial itself? Btw, there seems to be conflict for the workflow files

this does not change it at all. I will resolve conflicts

Issue #5951, opened 2025-04-13
lint checks | Planemo test workflow sometimes add _tests.yml instead of -tests.yml


dadrasarmin:
I notcied that sometimes the workflow tests file that is produced by Planemo causes lint error on GTN. It took sometime until I noticed the problem is that sometimes the files that are produced by Planemo for workflow test are ending in `_tests.yaml` instead of `-tests.yaml`. 

The reason is that Planemo checks the workflow filename (*.ga) and depending on the input file name it may use `-tests.yml` or `_tests-yml`. This is the chunk of code that does this in Planemo:

https://github.com/galaxyproject/planemo/blob/11011d85a52c20d72815de8a4df5fe702b3a603d/planemo/galaxy/workflows.py#L330-L340

In my view, this is a very difficult error to debug, specially for new users who want to contribute to maintaining and updating the tutorials and workflows. Therefore, I suggest to change 

https://github.com/galaxyproject/training-material/blob/353367373d11362547eb4e69ec2e5cb7bb119570/bin/lint.rb#L1518-L1525

to:

```
            if !test_file.match(/tests.yml/)
              results += [
                ReviewDogEmitter.file_error(path: path,
                                            message: 'Please use the extension tests.yml ' \
                                                     'for this test file.',
                                            code: 'GTN:032')
              ]
            end
```

Probably, we have to update a couple of other files/codes as well. I noticed that `-tests.yml` has been mentioned in both Planemo and GTN documentation while it can also be `_test.yml`.

I can make a PR, if you agree with me.

bgruening:
You could match `-tests.yml` or `_tests.yml`? Your suggestion seems good to me.

PR #6231, opened 2025-07-02
Update lint.rb to cover both -tests.yml and _tests.yml for workflow tests lint check

dadrasarmin:
I noticed a strange behaviour when I was updating some tools using [this tutorial](https://training.galaxyproject.org/training-material/topics/contributing/tutorials/updating_tutorial/tutorial.html).

I reported an issue [here](https://github.com/galaxyproject/training-material/issues/5951) about workflow test lint checks on GTN. I believe it is better to accept all combinations of `-tests.yml`, `-tests.yaml`, `_tests.yml`, and `_tests.yaml`. I changed the code to be similar to this code of GTN (here with `match` instead of `grep`):

https://github.com/galaxyproject/training-material/blob/353367373d11362547eb4e69ec2e5cb7bb119570/bin/list-workflows-with-tests.rb#L11

It seems to me that it should work. But I do not know how can I check it locally.


bin/lint.rb
```diff
@@ -1515,10 +1515,10 @@ def self.fix_file(path)
         else
           # Load tests and run some quick checks:
           possible_tests.each do |test_file|
-            if !test_file.match(/-tests.yml/)
+            if !test_file.match(/[_-]tests?.ya?ml/)
```
shiltemann, line 1518: this regex is now identical to the regex selecting `possible_tests` on line 1500, so this warning will never trigger.

PR #6269, opened 2025-07-25
Update training for users how to attach external storage to Galaxy in SRC

mirelaminkova:
Update the picture examples of how users to work inside SRC.  How to rename Galaxy, and how to attach external volumes.


mira-miracoli (approved):
I can not verify weather it works, because my University of Freiburg account does not have access to surf, but from the GTN perspective it looks fine I think :)

PR #6281, opened 2025-07-30
Text updates based on feedback for YOLO training tutorial 

sunyi000:
some text updates based on feedback @maulakhan

@kostrykin could you review it, thanks a lot.


topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -23,9 +23,9 @@ contributors:
 ---
 
 
-Image annotations and model training are essential in bioimage analysis tasks.  In biology and related fields, researchers often deal with large volumes of microscopy images that require accurate annotation to train machine learning models.  Automate this process can save time and improve reproducibility, but high-quality training data remains critical. Human-in-the-loop workflows have emerged as a solution to bridge the gap between manual annotation and automated model training, enabling iterative improvements through user interaction.
+In bioimages, challenging morphologies are often quite hard to segment using traditional computer vision/image analysis methods. Therefore, using semi-supervised machine learning or deep learning for such tasks is getting more popular.
```
kostrykin: The usage of *"or"* would suggest that these are disjoint concepts, but the latter is rather a subset, right?
```suggestion
In bioimages, challenging morphologies are often quite hard to segment using traditional computer vision/image analysis methods. Therefore, using semi-supervised machine learning methods like deep learning for such tasks is getting more popular. 
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -23,9 +23,9 @@ contributors:
...
+In bioimages, challenging morphologies are often quite hard to segment using traditional computer vision/image analysis methods. Therefore, using semi-supervised machine learning or deep learning for such tasks is getting more popular. 
 
-This tutorial introduces a reusable human-in-the-loop image segmentation and training workflow implemented in Galaxy, designed to streamline the creation of custom segmentation models using YOLO (You Only Look Once). The workflow integrates both interactive and headless tools. We demonstrate the workflow here with tail analysis images from growing embryo samples. The workflow can also be adapted to other datasets with minimal adjustments.
+This tutorial shows an easy-to-use and efficient workflow of training YOLO models in Galaxy. The workflow integrates both interactive and headless tools. We demonstrate the workflow here with bright field images of tails. The workflow can also be adapted to other datasets with minimal adjustments.
```
kostrykin: I think _"images of tails"_ is quite unspecific and doesn't tell much, the more important information is that the images are from growing embryos, isn't it?
```suggestion
This tutorial shows an easy-to-use and efficient workflow of training YOLO models in Galaxy. The workflow integrates both interactive and headless tools. We demonstrate the workflow here with bright field microscopy images from growing embryo samples. The workflow can also be adapted to other datasets with minimal adjustments.
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -145,10 +145,11 @@ In this tutorial, we annotate all 22 example images using the tool's `Auto label
 >
 > ![al open dir](../../images/yolo-train/al_label.png){: width="75%"}
 >
->    - If the inferred polygon is not accurate, click `-Point`, then click on a background (non-object) area to refine the results. The model will re-run the inference.
+>    **Note:** If the selected auto labeling model detects background or object not part of the object of interest, we can refine it by adding a negative prompt to the image where it detects background. Vice versa, if the model does not completely detect the boundaries of the objects, one can add a positive prompt to the areas inside object that were missed by the model.
```
kostrykin: > we can refine it by adding a negative prompt […] one can add a positive prompt

Is _"prompt"_ the proper term for image labeling? I only witnessed it in the context of LLMs…

```suggestion
>    **Note:** If the selected auto labeling model falsely detects background or objects that are not part of the object of interest, we can refine it by adding an annotation to indicate image background. Vice versa, if the model does not completely detect the boundaries of the objects, one can add annotations to the areas inside the object that were missed by the model. 
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -145,10 +145,11 @@ In this tutorial, we annotate all 22 example images using the tool's `Auto label
...
+>    **Note:** If the selected auto labeling model detects background or object not part of the object of interest, we can refine it by adding a negative prompt to the image where it detects background. Vice versa, if the model does not completely detect the boundaries of the objects, one can add a positive prompt to the areas inside object that were missed by the model. 
+>    - In the previous step, the results included some background, as shown in the picture. To refine them, click `-Point`, then click on the background area. The model will re-run the inference.
 >
 > ![al open dir](../../images/yolo-train/al_label_1.png){: width="75%"}
->    - After the inference is complete, click `Finish Object`. In the `Enter object label` box, type `tail` then click `OK` to save the annotation.
+>    - After the inference is complete, click `Finish Object`. In the `Enter object label` box, type `tail` then click `OK`. This will automatically create a label file in json format that contains the x,y coordinates of the polygon.
```
kostrykin: Lets be consistent with capitalization: (cf. lines 163, 180, etc.)

```suggestion
>    - After the inference is complete, click `Finish Object`. In the `Enter object label` box, type `tail` then click `OK`. This will automatically create a label file in JSON format that contains the coordinates of the polygon.
```

I think the x- and y-coordinates aren't referred to any later, so it's not important to mention them specifically?

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -159,7 +160,7 @@ In this tutorial, we annotate all 22 example images using the tool's `Auto label
 
 
 # Prepare training data
-In this step, we will convert the annotation file generated by AnyLabeling (in JSON format) into YOLO compatible TXT files. This conversion is necessary because the YOLO training tool expects annotations in its specific text format, where each object is described by a class ID and corresponding bounding box coordinates.
+In this step, we will convert the annotation file generated by AnyLabeling (in JSON format) into YOLO compatible txt files. This conversion is necessary because the YOLO training tool expects annotations in its specific text format, where each object is described by a class ID and corresponding bounding box coordinates. The YOLO txt file must be formatted according to the described [specifications](https://roboflow.com/formats/yolo).
```
kostrykin: I'm confused, you're using a capitalized notation of file formats (e.g., TIFF, JPG, JSON), but here you have changed from a consistent capitalization like `TXT` to `txt` that now is inconsistent with the other formats?

I'd suggest to stick to the consistent capitalization, like:

```suggestion
In this step, we will convert the annotation file generated by AnyLabeling (in JSON format) into YOLO compatible TXT files. This conversion is necessary because the YOLO training tool expects annotations in its specific text format, where each object is described by a class ID and corresponding bounding box coordinates. The YOLO TXT file must be formatted according to the described [specifications](https://roboflow.com/formats/yolo).
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -191,7 +192,7 @@ In this step, we will convert the annotation file generated by AnyLabeling (in J
 >        - *"Learning rate"*: `0.02`
 >        - *"Weight decay"*: `0.001`
 >        - *"Confidence"*: `0.5`
->        - *"IoU"*: `0.7`
+>        - *"IoU (Intersection over union)"*: `0.7`
```
kostrykin: This is confusing because the parameter is actually named `IoU` in the tool. When the reader is following the tutorial and trying to find the right parameter, they will probably take longer to spot it, because they will be looking for a much longer label. I'd suggest to stick to the previous version, and maybe add a comment on this separately.

```suggestion
>        - *"IoU"*: `0.7` (this is the *intersection over the union*, that means that… XXX)
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -210,17 +211,16 @@ In this step, we will convert the annotation file generated by AnyLabeling (in J
 > >
 > >    **Image size**: Determines the resolution used during training. Larger sizes can improve accuracy but require more memory and time.
 > >
+> >    **Image scale augmentation**: Resizes images by a random factor within the specified range. The scale hyperparameter defines the scaling factor, with the final adjustment randomly chosen between 1-scale and 1+scale. For example, with scale=0.5, the scaling is randomly selected within 0.5 to 1.5.
```
kostrykin: When reading the *"1-scale and 1+scale"* part, it takes a moment to understand what this is about, because one usually reads *"1-scale"* as *"one-scale"*, not as *"one minus the scale parameter"*. Therefore, I suggest changing as follows:
```suggestion
> >    **Image scale augmentation**: Resizes images by a random factor within the specified range. The scale hyperparameter defines the scaling factor, with the final adjustment randomly chosen within the range of [1–scale; 1+scale]. For example, with scale=0.5, the scaling is randomly selected within 0.5 to 1.5.
```
Note that this suggestion also employs using a minus sign instead of a hyphen (different lengths).

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -210,17 +211,16 @@ In this step, we will convert the annotation file generated by AnyLabeling (in J
...
 > >
-> >    **Image scale augmentation**: Applies random resizing of images during training to improve robustness. Values between 0.5 and 1.0 are commonly used.
+> >    **Image rotation augmentation**: Rotates images randomly within the specified range. The degrees hyperparameter defines the rotation angle, with the final adjustment randomly chosen between -degrees and degrees. For example, with degrees=10.0, the rotation is randomly selected within -10.0 to 10.0.
```
kostrykin: Change the hyphen to a minus sign:
```suggestion
> >    **Image rotation augmentation**: Rotates images randomly within the specified range. The degrees hyperparameter defines the rotation angle, with the final adjustment randomly chosen between –degrees and +degrees. For example, with degrees=10.0, the rotation is randomly selected within –10.0 to +10.0.
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -210,17 +211,16 @@ In this step, we will convert the annotation file generated by AnyLabeling (in J
...
 > >
-> >    **Image rotation augmentation**: Helps the model generalize to rotated objects. Keep small (e.g., 5–15°) for biological images to avoid distortion.
-> >
-> >    **Image HSV-Value**: Randomly adjusts image brightness and color during training. Can help with generalization, especially when image lighting varies.
+> >    **Image HSV-Value**: Changes the brightness of the image. The hsv_v hyperparameter defines the shift magnitude, with the final adjustment randomly chosen between -hsv_v and hsv_v. For example, with hsv_v=0.4, the intensity is randomly selected within -0.4 to 0.4.
```
kostrykin: Change the hyphen to a minus sign:

```suggestion
> >    **Image HSV-Value**: Changes the brightness of the image. The hsv_v hyperparameter defines the shift magnitude, with the final adjustment randomly chosen between –hsv_v and +hsv_v. For example, with hsv_v=0.4, the intensity is randomly selected within –0.4 to +0.4.
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -210,17 +211,16 @@ In this step, we will convert the annotation file generated by AnyLabeling (in J
...
 > >
 > >    **Learning rate**: Controls how fast the model learns. If training is unstable or not improving, try lowering this (e.g., to 0.01 or 0.005).
 > >
 > >    **Weight decay**: Helps prevent overfitting by penalizing large weights. Usually works well between 0.0001 and 0.01.
-> > 3. Increasing the confidence threshold results in fewer, more reliable segmentations, while decreasing it may produce more segments, including uncertain or noisy ones.
+> > 3. This value sets the minimum confidence threshold for detections. Objects detected with confidence below this threshold will be disregarded. Adjusting this value can help reduce false positives.
```
kostrykin: Takes a moment to understand, because *"value sets"* might be read in terms of *sets of values*, hence I suggest to use a different term here:
```suggestion
> > 3. This value governs the minimum confidence threshold for detections. Objects detected with a confidence below this threshold will be disregarded. Adjusting this value can help reduce false positive detections.
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -1,16 +1,16 @@
 ---
 layout: tutorial_hands_on
 
-title: Training Custom YOLO Models for Object Detection and Segmentation in Bioimages
+title: Training Custom YOLO Models for Segmentation in Bioimages
```
kostrykin:
```suggestion
title: Training Custom YOLO Models for Segmentation of Bioimages
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
```diff
@@ -237,4 +237,4 @@ Once training is complete, the resulting model can be used to label new images.
 
 # Conclusion
 
-This tutorial demonstrated how to train a custom YOLO model for object detection and segmentation in bioimages using Galaxy. By combining image preprocessing, human-in-the-loop annotation, format conversion, and model training, we have built a complete workflow. We can apply the trained model to new image data and further refine our workflow as needed.
+This tutorial demonstrated how to train a custom YOLO model for segmentation in bioimages using Galaxy. By combining image preprocessing, human-in-the-loop annotation, format conversion, and model training, we have built a complete workflow. We can apply the trained model to new image data and further refine our workflow as needed.
```
kostrykin:
```suggestion
This tutorial demonstrated how to train a custom YOLO model for segmentation of bioimages using Galaxy. By combining image preprocessing, human-in-the-loop annotation, format conversion, and model training, we have built a complete workflow. We can apply the trained model to new image data and further refine our workflow as needed.
```

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
sunyi000: i actually think just `IoU` should be fine. The explanation of IoU is in the tool's help ....

topics/imaging/tutorials/yolo-segmentation-training/tutorial.md
sunyi000: I agree..how about use `positive point`, `negative point` instead, they are the same as the button label in Anylabeling? @kostrykin 

"...we can refine it by adding a **negative point** to indicate image background....one can add  a **positive point** to the areas inside the object ...."

kostrykin: Yes the post is that the meaning of "positive" and "negative" im somewhat obscure here. I think that explicitly saying background vs. foreground/object is much clearer. Maybe you can also combine it like
```suggestion
**Note:** If the selected auto labeling model falsely detects image background or objects that are not part of the object of interest, we can refine it by adding an annotation to indicate image background (negative point). Vice versa, if the model does not completely detect the boundaries of the objects, one can add annotations to the areas inside the object that were missed by the model (positive points).
```
But still, speaking of points sure is better than speaking of prompts:
```suggestion
**Note:** If the selected auto labeling model falsely detects image background or objects that are not part of the object of interest, we can refine it by adding a *"negative point"* to indicate image background. Vice versa, if the model does not completely detect the boundaries of the objects, one can add a *"positive point"* to the area inside the object that was missed by the model.
```

sunyi000: thanks @kostrykin , text has been updated as suggested.

PR #6316, opened 2025-08-28
Update 2025-09-17-ws-historikertag.md

Sch-Da:
Adding additional funding to the training event, as @bgruening suggested.

Does this also update the Galaxy Hub event here https://galaxyproject.org/events/? Or can I do this elsewhere? As it is an external link, I was unsure here.
Thanks for your suggestions!


bgruening:
I'm not sure about the update to the Hub, we will see.

Sch-Da:
@teresa-m - I did not notice - sorry. I was trying to multitask, and it did not pay off.

PR #6327, opened 2025-09-02
[Bugfix] Fix pURLs being replaced

shiltemann:
Entries for E00013 and T00080 were missing, throwing off the count used to determine the next number in the sequence. Added dummy entries to prevent this going forward.

Also updated the pIDs after this change


shiltemann:
ping @bgruening FYI

bgruening:
Merci!!

PR #6321, opened 2025-09-01
Add GAT Brno event, the gtn-style

martenson:
modified from the 23 ghent event

PR #6328, opened 2025-09-02
added admin training as external event

teresa-m:
Adding the admin training in November as an external event.

@martenson is it ok to add cesnet as an Organisation? And is it fine to add elixir-europ instead of elixir-cz as funding? I can also add elixir-cz if you want.


shiltemann:
@teresa-m It looks like @martenson is already adding a proper GTN event page for this here: https://github.com/galaxyproject/training-material/pull/6321

If you just want to announce it to the GTN audience, maybe add it as a news post?

martenson:
@teresa-m I added the two in https://github.com/galaxyproject/training-material/pull/6321/commits/ad1295c109d3d9b5897449fcc85a0c712d56a8e6, we can close this I think

PR #6350, opened 2025-09-12
Update features_scratchbook.md

Sch-Da:
When using this feature, step 3, exiting the window manager by clicking, did not work. Maybe someone who wrote this initially can check, and if they agree, this information could be deleted.


shiltemann:
Thanks @Sch-Da!

PR #6366, opened 2025-09-19
Mass spectrometry: LC-MS analysis - add links to tools

nilchia:
This PR adds links to the corresponding tools.


foellmelanie:
Thank you for adding the links, please confirm that you have double checked every result/Q&A/picture that they didn't change with the tool version you have chosen for the link?
Cheers, Melanie

nilchia:
For the batch correction, the latest version does not work. I had to downgrade it.

bgruening (approved):

melpetera:
Hi, and thank you for your work.
Just a quick comment about the Batch_correction latest version not working. It works well on usegalaxy.fr, but indeed you can not currently run it successfully on usegalaxy.eu [due to an installation issue somehow](https://community.france-bioinformatique.fr/t/batch-correction-error/5625/3). 
Best regards

melpetera:
Ah, indeed! I missed that, sorry.
Best

nilchia:
Hello @melpetera,
I noticed that if participants use the data in the library, they will get an error in the univariate analysis:
```
            Error: Column names of the dataMatrix are not identical to the row names of the sampleMetadata; check your data with the 'Check Format' module in the 'Quality Control' section
Execution halted
```
and that is because the order of the samples is reversed.
(please look at the first univariate analysis [here](https://usegalaxy.eu/u/nilchia/h/lc-ms-input-error-test))

Should we update the library sample metadata file? Or should we add a `sort` step?
like the one I did in the history above after the failed job?

best,
Amirhossein

melpetera:
This is very strange. The order is indeed reversed in your history, and this is because the order in the dataset collection you get for the original mzML data is reversed (knowing that the order in the dataMatrix is based on the order in the dataset collection in fact). 
I checked what happens when importing the data on usegalaxy.eu and usegalaxy.fr, and in both cases I obtained in the dataset collection the order corresponding to the one in the GTN. But in the case of your history, I see that the order is reversed: 
image
Interestingly, when importing the data not as a dataset collection directly, but building the dataset collection afterward from the history option, then you obtain the same order as the one from your history, which seems to be the alphabetical order. 

From this, I see several possibilities (some we already discussed previously in fact):
- option 1: adding in the GTN a note about the order in the dataset collection (specifying that in case you missed to import the data as dataset collection from the start, you will need to handle the sample order between table)
- option 2: updating the Univariate tool to correctly handle the tables instead of returning this unconvenient error message
- option 3: adding a systematic step to check the tables in the GTN
- option 4: changing the original dataset order for it to be in original order

About option 2, it has been years since this is pending, so maybe not the best horse to bet on (although would be the most sustainable thing to do). 
About option 3, it is possible to do so, but then it would be more appropriate to add a dedicated step with the "Check_format" tool about checking the data. As I see it, it would be the best option from a "training" point of view; the reason it is currently not the case is mainly because doing it would indirectly induce to accept that option 2 will never be (and I still had some hope somehow so this is why I had not proceed to option 3 previously).
About option 1, no specific comment, it is feasable. It would be also to recommend to use Check_format anyway but in a different manner. 
About option 4, yeah I can do that - seems the simpliest in fact.

I have an update of this GTN on my plate regarding the CAMERA part, so I can include this question in it.

Thanks for the feedback.

nilchia:
Thanks a lot for checking this.

That's right. As you mentioned, it has the correct order when imported as a collection.

Maybe sorting the samplemetadata is the easiest solution.

Please let me know if I can help with this matter!

PR #6399, opened 2025-10-04
try migrating labeller to new version

bgruening:
@shiltemann I tried to adjust here the labeler bot to the new version 6


shiltemann:
ah thanks!

Issue #6397, opened 2025-10-01
Correction to scATAC preprocessing tutorial


pavanvidem:
[Pre-processing of 10X Single-Cell ATAC-seq Datasets
](https://gxy.io/GTN:T00335)
 
**Sinto fragments** tool should use the param **Take cell barcode into account when collapsing duplicate fragments** to avoid randomly choosing one of the duplicated fragments. With this param on, the tool should consider the barcode when collapsing.

PR #6418, opened 2025-10-15
Fix a param in scATAC-seq preprocessing tutorial and update the tools

pavanvidem:
- Fixes https://github.com/galaxyproject/training-material/issues/6397 .Thanks @lldelisle for pointing this out.
- Updated the tools in the FASTQ to Mtx workflow


MarisaJL (approved):
Looks good to me!

PR #6421, opened 2025-10-16
Update metadata.yaml to add media annotation subsection

yvanlebras:
Add a subsection for media annotation


shiltemann:
thanks @yvanlebras!

PR #6426, opened 2025-10-20
Reviewing OpenRefine Tutorial

Sch-Da:
I reviewed the tutorial mostly written by @dadrasarmin - thanks so much for your work and time invested in this! 

I made the following adaptations:
- streamlined some things - punctuation and user address when I noticed irregularities.
- had to relocate some pictures that came earlier than the text they described, found one typo that impacted the GREL expressions and fixed that.
- In some places, where tasks and logic were not clear to me, I adapted them with the help of the original tutorial and expanded them. 

I could not use the link to the Zenodo item to upload the 2 files. I, therefore, changed them into 2 separate links to the respective items. Hope this is the best practice. Please change if I am wrong.

We talked about the question of screenshots in this tutorial @shiltemann - I would leave it like this for now, if this works for you.

If there are no objections, this could be merged from my side.

### DIFF
--- a/topics/digital-humanities/tutorials/open-refine-tutorial/tutorial.md
+++ b/topics/digital-humanities/tutorials/open-refine-tutorial/tutorial.md
@@ -102,7 +102,7 @@ We suggest that you download the data from the Zenodo record as explained below.
 >
 {: .hands_on}
 
-In this first part, we will focus on working with the metadata from the Powerhouse Museum. The additional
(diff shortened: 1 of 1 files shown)


shiltemann (approved):

shiltemann:
@Sch-Da good catch regarding the Zenodo files, that is indeed how it should be done :+1: 

Thanks all!

Sch-Da:
thanks so much @shiltemann!

PR #6435, opened 2025-10-28
ref-based-rad-seq tutorial.md

jennaj:
@yvanlebras or @VerenaMoo -- please double check my work for this URL change! The original is pointing to a dead end and _I think_ my suggestion is the best choice now .. but please feel free to reject this and replace with whatever is actually better!


shiltemann:
Thanks @jennaj! This looks correct to me, so going to merge it now since the old link is broken, if yvan has a better suggestion we can update it again later :+1:

PR #6437, opened 2025-10-28
Do not run rss-to-bluesky job on schedule on forks

mittler-works:
Within the blueskyfeedbot action there is a job rss-to-bluesky which will try to create bluesky posts for gtn news posts.

This _should_ not run on forks and as there are no secrets in forks it also _can_ not run in forks (and it fails). As this action is triggered daily via cron, it is pretty annoying having a fork with daily failing runs triggering daily emails to the dev.

This commit allows scheduled runs only on the `galaxyproject/training-material` repository and not on forks. However, for forks it is still possible to trigger the action manually, e.g. for testing.

### DIFF
--- a/events/2025-11-11_ecology.md
+++ b/events/2025-11-11_ecology.md
@@ -28,7 +28,7 @@ async: false
 mode: hybrid
 
 registration:
-  link: contact Yvan
+  link: "mailto:yvan.le-bras@mnhn.fr"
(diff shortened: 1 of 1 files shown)


shiltemann (approved):
thanks @mittler-works :+1:

shiltemann (approved):

PR #6494, opened 2025-11-20
[GAT] adjust admin path FAQ for gatbrno schedule

martenson:


mvdbeek (approved):

topics/admin/faqs/git-gat-path.md
```diff
@@ -8,7 +8,7 @@ contributors: [hexylena]
 
 The yearly Galaxy Admin Training follows a specific ordering of tutorials. Use this timeline to help keep track of where you are in Galaxy Admin Training.
 
-{% assign tutorials = "admin/ansible-galaxy admin/backup-cleanup admin/customization admin/tus admin/cvmfs admin/apptainer admin/tool-management admin/reference-genomes admin/data-library dev/bioblend-api admin/connect-to-compute-cluster admin/job-destinations admin/pulsar admin/celery admin/gxadmin admin/reports admin/monitoring admin/tiaas admin/sentry admin/ftp admin/beacon" | split: " " %}
+{% assign tutorials = "admin/ansible admin/ansible-galaxy admin/connect-to-compute-cluster admin/job-destinations admin/tus admin/celery admin/pulsar admin/apptainer admin/cvmfs admin/reference-genomes admin/object-store admin/gxadmin admin/sentry admin/backup-cleanup admin/tool-management admin/interactive-tools" | split: " " %}
```
hexylena, line 11: so, this is somewhat tied into the knitting process (hence `git-gat`-path). that's why admin/ansible wasn't in there. not 1:1, but, if you're thinking about cutting things from here, you may wish to think about cutting them from git-gat as well, as the steps will no longer match up.

if you're dropping these tutorials, then you *need* to drop this snippet from those tutorials (e.g. reports) that are no longer included. Otherwise, they probably won't render correctly at best, and will be actively confusing for users at worst.

this is a "faq" insofar as it was the easiest place to stick the snippet. It isn't really an FAQ, it's a template box to include on tutorials, so, please check that as well.

martenson: @hexylena thanks mate! will follow up

PR #6502, opened 2025-11-21
[GAT] Fix slurm and tpv trainings

pauldg:
There are some legacy configuration options for slurm and TPV in the trainings.

For the connecting to a cluster training: https://github.com/galaxyproject/training-material/blob/e822f5fe70007fb085e8f1382833c79ea02d14a1/topics/admin/tutorials/connect-to-compute-cluster/tutorial.md?plain=1#L165 
`cons_res` is legacy and is replaced by `cons_tres`
For the TPV training: https://github.com/galaxyproject/training-material/blob/e822f5fe70007fb085e8f1382833c79ea02d14a1/topics/admin/tutorials/job-destinations/tutorial.md?plain=1#L732 
`params['walltime']` should be `entity.params.get('walltime')`

### DIFF
--- a/topics/admin/tutorials/job-destinations/tutorial.md
+++ b/topics/admin/tutorials/job-destinations/tutorial.md
@@ -729,7 +729,7 @@ Lastly, we need to write a rule in TPV that will read the value of the job resou
 >         max_mem: 8
 >         params:
 >    -      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores} --mem={round(mem*1024)}
->    +      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores} --mem={round(mem*1024)} --time={params['walltime']}:00:00
+>    +      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores} --mem={round(mem*1024)} --time={entity.params['walltime']}:00:00
 >    {% endraw %}
 >    ```
 >    {: data-commit="process resource params in TPV"}
--- a/topics/admin/tutorials/pulsar/tutorial.md
+++ b/topics/admin/tutorials/pulsar/tutorial.md
@@ -807,7 +807,7 @@ For this tutorial, we will configure Galaxy to run the BWA and BWA-MEM tools on 
 >    @@ -54,3 +54,18 @@ destinations:
 >         max_mem: 8
 >         params:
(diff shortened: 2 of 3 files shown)


topics/admin/tutorials/job-destinations/tutorial.md
```diff
@@ -729,7 +729,7 @@ Lastly, we need to write a rule in TPV that will read the value of the job resou
 >         max_mem: 8
 >         params:
 >    -      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores}
->    +      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores} --time={params['walltime']}:00:00
+>    +      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores} --time={entity.params.get('walltime')}:00:00
```
mvdbeek: This seems odd, if you anticipate walltime isn't set you would need to set a default, otherwise indexing via `[]` is what I'd use ... there's an obvious error if the value isn't set.

... also if `{cores}` works, i would keep it consistent and use `{params['walltime']}` ?

pauldg: Good point, I was testing the training on one of the EU training vms and was getting an error saying params was undefined. I'll post the logs in a couple of hours.

mvdbeek: I'm rerunning the tutorials and will make sure it works (or not 😆 )

mvdbeek: OK, I think I get it now, it does need to be entity, because params is evaled as an f-string and is not a top level attribute of the EntityModel. I've used `[` though since I don't think we'd ever want `None:00:00` as walltime.

mvdbeek (approved):

topics/admin/tutorials/job-destinations/tutorial.md
```diff
@@ -361,7 +361,7 @@ We want our tool to run with more than one core. To do this, we need to instruct
 >    +    runner: slurm
 >    +    max_accepted_cores: 16
 >    +    params:
->    +      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores}
+>    +      native_specification: --nodes=1 --ntasks=1 --cpus-per-task={cores} --mem={round(mem*1024)}
```
lldelisle, line 364: :heart_eyes:

PR #6503, opened 2025-11-22
[GAT] Remove diff fixing in admin training

mvdbeek:
This currently results in diffs missing all of the first column. This was initially added in
https://github.com/galaxyproject/training-material/commit/3dbb1c2355ab0d69cd6f02c15475604c989bf1df. I can only specualte why this fix isn't working now, perhaps the class list has changed since then. Just removing the fixup results in corrects diffs, e.g. in
training-material/topics/admin/tutorials/ansible-galaxy/tutorial.html#hands-on-installing-roles, but also all other places I've spot checked.

### DIFF
--- a/topics/admin/tutorials/ansible-galaxy/tutorial.md
+++ b/topics/admin/tutorials/ansible-galaxy/tutorial.md
@@ -1093,7 +1093,7 @@ The configuration is quite simple thanks to the many sensible defaults that are 
 >    +        pools:
 >    +          - job-handlers
 >    +          - workflow-schedulers
->     
+>
 >     galaxy_extra_dirs:
 >       - /data
 >    {% endraw %}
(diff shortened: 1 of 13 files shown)


martenson (approved):

mvdbeek:
> Did you expect real patches?

yes, certainly if you look at https://training.galaxyproject.org/training-material/topics/admin/tutorials/cvmfs/tutorial.html#tip-how-to-read-a-diff which is included after every diff. This explains how a real diff looks like, so that's very confusing if we're not rendering a real diff.

> Additionally this PR likely breaks some of the knitting, as the whitespace was removed.

i've reknittted many times since then, and the diffs as rendered are otherwise including extra whitespaces in the context lines. How are you linting this ?

PR #6515, opened 2025-11-24
Admin training: add some lines on how to apply a diff

bernt-matthias:
Maybe we should add an additional FAQ, but this way it will be shown everywhere where needed.

### DIFF
--- a/topics/admin/faqs/diffs.md
+++ b/topics/admin/faqs/diffs.md
@@ -6,6 +6,21 @@ layout: faq
 contributors: [hexylena]
 ---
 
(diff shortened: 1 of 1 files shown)


martenson:
@hexylena we didn't necessarily choose this, but before https://github.com/galaxyproject/training-material/pull/6503 the copying was broken (gave you all the diff content, just without `+` and `-`) and we "fixed" it this way, to have full diffs. I am not sure which way is superior.

>fingers feel rather twisted after typing patch -p1 << 'EOF' quite a bit :)

@bernt-matthias use reverse search ;)

PR #6523, opened 2025-11-26
Add subtopics to imaging

beatrizserrano:
Added subtopics, annotated tutorials and included a couple of references to the imaging topic

PR #6525, opened 2025-11-27
Remove one more duplicate subtopic key

mvdbeek:
xref https://github.com/galaxyproject/training-material/pull/6523


shiltemann:
I am afraid this will still break because the fair-data subtopic is not defined for the imaging topic. I undid the symlinking of the tutorials here: https://github.com/galaxyproject/training-material/pull/6526

PR #6568, opened 2025-12-15
fix indentation

hexylena:
some snippets even fail to render due to incorrect indentation.


bgruening:
@khaled196 can you maybe look at this?

Thanks @hexylena

Issue #6377, opened 2025-09-23
Update tutorials to use collections instead of separate paired datasets


bernt-matthias:
https://github.com/galaxyproject/tools-iuc/pull/7281 removes the option to input paired data as separate datasets, but requires a pair. 

A few tutorials could need an update: 

- topics/sequence-analysis/tutorials/mapping/tutorial.md
- topics/genome-annotation/tutorials/amr-gene-detection/tutorial.md
- topics/epigenetics/tutorials/atac-seq/tutorial.md
- topics/epigenetics/tutorials/formation_of_super-structures_on_xi/tutorial.md

wm75:
@shiltemann @teresa-m @bernt-matthias the above is the list of tutorials affected by the bowtie2 change.

Since last year, however, a number of additional tools have been changed in similar ways:
- minimap2
- kraken2
- fastp
- spades (and all its variants)

I went through the tutorials using any of these tools and, if I haven't missed anything, this is the list of affected ones:

- minimap2: didn't find anything (typical tutorials use SE reads with the tool)

- kraken2:
  - https://training.galaxyproject.org/training-material/topics/ecology/tutorials/bacterial-isolate-species-contamination-checking/tutorial.html#hands-on-assign-taxonomic-labels-with-kraken2
  - https://training.galaxyproject.org/training-material/topics/microbiome/tutorials/taxonomic-profiling/tutorial.html#hands-on-assign-taxonomic-labels-with-kraken2 (just the name of the option)
  - https://training.galaxyproject.org/training-material/topics/sequence-analysis/tutorials/quality-contamination-control/tutorial.html#hands-on-assign-taxonomic-labels-with-kraken2

- fastp:
  - https://training.galaxyproject.org/training-material/topics/assembly/tutorials/mrsa-illumina/tutorial.html#hands-on-quality-improvement
  - https://training.galaxyproject.org/training-material/topics/assembly/tutorials/mrsa-nanopore/tutorial.html#hands-on-quality-improvement-of-the-illumina-reads
  - https://training.galaxyproject.org/training-material/topics/sequence-analysis/tutorials/quality-contamination-control/tutorial.html#hands-on-quality-improvement
[...]

PR #6575, opened 2025-12-18
Update spades to use PE collection in De Bruijn Graph Assembly tutorial

tflowers15:
Based on the GTN issue "Update tutorials to use collections instead of separate paired datasets #6377".

This update to the [De Bruijn Graph Assembly](https://training.galaxyproject.org/training-material/topics/assembly/tutorials/debruijn-graph-assembly/tutorial.html#de-bruijn-graph-assembly) tutorial updates the version of spades used to 4.2.0+galaxy0 and replaces the separate paired end dataset inputs with a paired end collection (required PE input for this version of spades). Also added the tool tip for creating paired end collections to the "Hands-on: Getting the data" and updated the versions of velveth and velvetg (versions 1.2.10.3 and 1.2.10.2, respectively).


tflowers15:
I have created three new workflows to account for not all tools in the tutorial being available (or having restrictions on usage) on all usegalaxy.* servers.
- `Debruijn-Graph-SPAdes-Assembly.ga`: only runs SPAdes steps
- `Debruijn-Graph-Velvet-Optimiser-and-SPAdes-Assembly.ga`: Runs SPAdes and Velvet Optimiser steps
- `Debruijn-Graph-Velvet-Full-and-SPAdes-Assembly`: Runs all steps (SPAdes, Velvet Optimiser and velvetg/h)

I originally deleted the old workflow `debruijn-graph.ga` (runs SPAdes and Velvet Optimiser steps), however this cause the Linting check to fail as there is a link/reference expecting to find this workflow file. To fix the Linting check, I added a new file with this name (and an associated tests.yml), which is the same as `Debruijn-Graph-Velvet-Optimiser-and-SPAdes-Assembly.ga` (corresponds to the update for the old workflow).

Is it possible to remove the link/reference to the `debruijn-graph.ga` workflow? I think it would be better to use the updated, more descriptive name as there will now be multiple versions of the workflow.

topics/assembly/tutorials/debruijn-graph-assembly/tutorial.md
```diff
@@ -208,21 +222,21 @@ The next thing to be aware of is that there can be multiple valid interpretation
...
 >
 > Because the repeated element is longer than the sequencing reads, the assembler was not able to reproduce the original genome as a single contig. Rather, three contigs are produced: one for the repeated sequence (even though it occurs twice) and one for each sequence between the repeated elements.
 >
 > Given only the contigs, the relationship between these sequences is not clear. However, the assembly graph contains additional information which is made apparent in Bandage:
 >
-> ![Simple example 3](https://camo.githubusercontent.com/406648509cf478ac0b2ab9f2447aec4e7575b7dd/687474703a2f2f72727769636b2e6769746875622e696f2f42616e646167652f696d616765732f77696b692f73696d706c655f6578616d706c655f332e706e67)
+> ![Simple example 3](https://camo.githubusercontent.com/1fc8116ea0532646250cdf702b7e2d68ab0126ad198f51fea5288d039a6875fd/687474703a2f2f72727769636b2e6769746875622e696f2f42616e646167652f696d616765732f77696b692f73696d706c655f6578616d706c655f332e706e67)
```
shiltemann: @tflowers15 I just moved these images into GTN while I was here. Just wondering if it is expected that image 1 and 3 in this box are the same?

tflowers15: That's a good idea. Sorry I didn't move the images. That must have been my mistake with the links, no image 1 and image 3 should be different. I have replaced `bandage1.png` in the images directory.

One question, is there a distinction between including images directly via the path vs using a link and is there a reason why one method should be used over the other?
  - Path: `bandage spades`
  - Link: `Simple example 1`

Should all images be include using the same method (e.g. change all to using the link method or change the four bandage*.png image to using the path)?

shiltemann: The `{% link %}` method is preferred because then broken links are detected at build time. We do not have a hard rule for this as we also have a separate link checker, but whenever I have to update a link I make sure to use the link directive.

Thanks for replacing the duplicate image!

tflowers15: Ok, thanks for the explanation. I'll start using the link method when I make GTN contributions moving forward.

PR #6623, opened 2026-01-29
Fix: Explicitly use conda-provided gem binary during install to avoid macOS system Ruby conflicts

dadrasarmin:
On Intel-based MacBooks, the system Ruby (v2.6) often takes precedence in the PATH even after environment activation, causing make install to fail because rubygems-update requires Ruby >= 3.2.0. This change explicitly points to the gem binary within the Conda prefix to ensure the correct version is used.

Tested and works on a computer with the following info:

> ProductName:            macOS
> ProductVersion:         26.2
> BuildVersion:           25C56
> Model Name: MacBook Pro
> Model Identifier: MacBookPro16,2
> Processor Name: Quad-Core Intel Core i7

To build the website locally, I also had to provide `make serve-quick HOST=127.0.0.1` instead of the default `0.0.0.0`. Added this information to `rendering_gtn` file.

### DIFF
--- a/topics/contributing/tutorials/rendering_gtn/tutorial.md
+++ b/topics/contributing/tutorials/rendering_gtn/tutorial.md
@@ -627,7 +627,7 @@ Once Jekyll and its modules are installed in our conda environment, we can check
 
 > <hands-on-title>Checking the website generation locally</hands-on-title>
 >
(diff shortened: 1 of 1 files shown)


bgruening (approved):
Looks good to me!

PR #6625, opened 2026-01-29
Galaxy hub grants update for grants.yaml file

dadrasarmin:
There are some grants on Galaxy-Hub that are missing from the GRANTS file of the GTN repository. As suggested by Bjoern, I added them to this file and added `gtn-halloffame: "no"` as I was told. I added the names in alphabetically order.

I do not think that the entry `ViCE` has `funding_id` or `funding_database`. What should I do about this?

### DIFF
--- a/bin/schema-grants.yaml
+++ b/bin/schema-grants.yaml
@@ -17,6 +17,16 @@ mapping:
                     The full name for the grant
                 _examples:
                     - Gallantries: Bridging Training Communities in Life Science, Environment and Health
+            gtn-halloffame:
+                type: str
+                enum:
+                    - "no"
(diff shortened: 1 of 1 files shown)


GRANTS.yaml
```diff
@@ -343,3 +367,10 @@ skills4eosc:
     funder_name: Horizon Europe
     url: https://www.skills4eosc.eu/
     github: false
+
+ViCE:
```
bgruening: I have ask Dirk, if he still has some more information. Drop it for now.

bgruening (commented):
Thanks @dadrasarmin

PR #6626, opened 2026-01-29
Added new organisations from galaxy-hub.

dadrasarmin:
There are some organisations on Galaxy-Hub that are missing from the ORGANISATIONS file of the GTN repository. As suggested by Bjoern, I added them to this file and added `gtn-halloffame: "no"`. I added the names in alphabetically order.

### DIFF
--- a/ORGANISATIONS.yaml
+++ b/ORGANISATIONS.yaml
@@ -146,24 +146,6 @@ eubi:
     url: https://www.eurobioimaging.eu/
     gtn-halloffame: "no"
 
-galaxy:
-    name: Galaxy Project
-    github: false
-    url: https://galaxyproject.org/
-    gtn-halloffame: "no"
-
-galaxy-australia:
-    name: Galaxy Australia
-    github: false
-    url: https://www.biocommons.org.au/galaxy-australia
-    gtn-halloffame: "no"
-
-galaxy-europe:
-    name: Galaxy Europe
-    github: false
-    url: https://galaxyproject.org/eu/
-    gtn-halloffame: "no"
-
 galaxyworks:
     name: GalaxyWorks
     url: https://galaxyworks.io/


ORGANISATIONS.yaml
```diff
@@ -47,6 +47,36 @@ avans-atgm:
...
     ror: "015d5s513"
 
+aws:
+    name: Amazon Web Services
+    github: false
+    url: https://aws.amazon.com/
+    gtn-halloffame: "no"
+
+bioconda:
+    name: BIOCONDA
```
bgruening:
```suggestion
    name: Bioconda
```

ORGANISATIONS.yaml
```diff
@@ -47,6 +47,36 @@ avans-atgm:
...
+    gtn-halloffame: "no"
+
+biolytix:
+    name: Biolytix AG
+    url: https://www.biolytix.ch/en/
+    github: false
+    gtn-halloffame: "no"
+
+bmbf:
+    name: Bundesministerium für Bildung und Forschung
```
bgruening, line 75: that should be funding

ORGANISATIONS.yaml
```diff
@@ -57,12 +87,137 @@ carpentries:
...
+    gtn-halloffame: "no"
+
+eubi:
+    name: Euro-BioImaging
+    github: false
+    url: https://www.eurobioimaging.eu/
+    gtn-halloffame: "no"
+
+galaxy:
+    name: Galaxy Project
```
bgruening: not sure that makes sense

ORGANISATIONS.yaml
```diff
@@ -57,12 +87,137 @@ carpentries:
...
+    url: https://www.biocommons.org.au/galaxy-australia
+    gtn-halloffame: "no"
+
+galaxy-europe:
+    name: Galaxy Europe
+    github: false
+    url: https://galaxyproject.org/eu/
+    gtn-halloffame: "no"
+
+GalaxyWorks:
```
bgruening: what is your feeling, should we keep it all lower-letter?

ORGANISATIONS.yaml
```diff
@@ -57,12 +87,137 @@ carpentries:
...
+    github: false
+    gtn-halloffame: "no"
+
+JXTXFoundation:
+    name: JXTX Foundation
+    url: https://jxtxfoundation.org/
+    github: false
+    gtn-halloffame: "no"
+
+KWS:
```
bgruening:
```suggestion
kws:
```

ORGANISATIONS.yaml
```diff
@@ -57,12 +87,137 @@ carpentries:
...
+    name: KWS
+    url: https://www.kws.com/corp/en/
+    github: false
+    gtn-halloffame: "no"
+
+Limagrain:
```
bgruening:
```suggestion
limagrain:
```

ORGANISATIONS.yaml
```diff
@@ -505,3 +666,15 @@ vib:
     avatar: "/training-material/shared/images/logo-vib.png"
     github: false
     ror: "03xrhmk39"
+Worfklow4Metabelomics:
```
bgruening:
```suggestion

Worfklow4Metabelomics:
```

ORGANISATIONS.yaml
dadrasarmin: What is the following information for BMBF?
```
funding_id: "XXXX"
funding_database: XXXX
```

ORGANISATIONS.yaml
dadrasarmin: I agree. But it is listed as supporter of at least one news or event item:
https://preview.galaxyproject.org/hall-of-fame/galaxy/
What happens if it is listed under contribution and it is not present in the `organisations.yaml`?

ORGANISATIONS.yaml
dadrasarmin: I converted all entries that I added to lower case. There are entries from past that start or contain uppercase. I am not sure if GTN people would be happy if I touch those.

ORGANISATIONS.yaml
dadrasarmin: Solved it by removing those here: https://github.com/galaxyproject/galaxy-hub/pull/3601

ORGANISATIONS.yaml
bgruening: every BMBF grant has its own funding_id ... where is this mentioned? Maybe we need to drop it completely and replace it with the proper funding ID. e.g. deNBI, de.KCD are all BMBF funded projects

ORGANISATIONS.yaml
bgruening: Thanks

PR #6706, opened 2026-03-06
Update mismatch statistics in mapping tutorial

Swathi266:
Tutorial-
https://training.galaxyproject.org/training-material/topics/sequence-analysis/tutorials/mapping/tutorial.html


shiltemann:
@Swathi266 thanks for the update. Can you tell why the numbers are now different? Did you use newer version of the tool than are indicated in the tutorials? (if so it would be a good idea to also update these versions)

shiltemann (approved):

shiltemann:
@Swathi266 thanks a lot for your contribution! (sorry it took so long to process) ..I also updated the tool versions and workflow to match the versions you used to get the numbers :+1:

Swathi266:
Thank you @shiltemann

PR #6746, opened 2026-03-25
elimination of quast html fetching from metagenome assembly tutorial

gdefazio:
HTML fetching from zenodo doesn't work because of filtering and security reasons. Then, is better to eliminate this part.

TODO: Describe your changes here..


shiltemann:
thanks @gdefazio! I have added a link to an example history continaing the QUAST outputs. Only on Galaxy EU for now because I had problems importing to Main and AU, but it should be visible without logging in as well

shiltemann (approved):

PR #6780, opened 2026-04-13
Add contributions categories to feeds

arash77:
Adds `contributions:<role>:<id>` categories to topic feeds, matrix itemized feeds, and event feeds, mirroring the existing block already present in `generate_matrix_feed`.

### DIFF
--- a/_plugins/feeds.rb
+++ b/_plugins/feeds.rb
@@ -182,6 +182,14 @@ def generate_topic_feeds(site, topic, bucket)
               end
             end
           end
+
+          if page.data.key?('contributions')
+            page.data['contributions'].each do |role, ids|
+              Array(ids).each do |id|
+                xml.category(term: "contributions:#{role}:#{id}")
+              end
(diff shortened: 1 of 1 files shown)


shiltemann:
thanks @arash77! Some tutorials still have the older `contributors:` metadata key instead of `contributions`, do you need any changes in those cases as well (e.g. to add the contributor ids)?

arash77:
Good catch! I've pushed a follow-up commit (https://github.com/galaxyproject/training-material/pull/6780/commits/7ad485039eb3af97e1f9039f37ed12ccd771fdde) that handles the legacy `contributors:` key as well — those IDs are now emitted as `contributions:authorship:<id>` categories in all three feed generators, consistent with how `Gtn::Contributors.get_authors` already treats them.

shiltemann (approved):

shiltemann:
@arash77 perfect, thanks! :)

PR #6788, opened 2026-04-14
Delete markdown CSS style on EU co-funding image

unode:
Removed inline style from EU co-funding image that is not understood by the template at https://galaxyproject.org/hall-of-fame/biont/

@bgruening we patched the URL yesterday but the CSS annotation doesn't seem to render either - https://galaxyproject.org/hall-of-fame/biont/


bgruening (approved):
Jupp, makes sense.

PR #6791, opened 2026-04-15
New code owners for single-cell topic

pavanvidem:
How does it work here? Is it allowed to change codeowners?
I replaced @mtekman and @hexhowells, who have not been active lately, with @nilchia and @dianichj.


shiltemann:
@pavanvidem absolutely, anybody who would like to receive notifications for any parts of the GTN is free to add themselves here. Normally I add all the maintainers by default but looks like I missed these, thanks!

shiltemann (approved):

PR #6796, opened 2026-04-21
Add avatar URL for de.KCD organization

Sch-Da:
Adding url for de.kcd avatar to show on the hub.


shiltemann:
@Sch-Da Thanks! After @bgruening's PR on the hub (https://github.com/galaxyproject/galaxy-hub/pull/3941) this should no longer be needed, as it will automatically use the GitHub avatar when nothing is specified in this file (like the GTN does).

Sch-Da:
That makes it much neater. I did not get that earlier, sorry for the redundancy!

PR #6801, opened 2026-04-24
Update tutorial.md add level

yvanlebras:
level specification


shiltemann:
@yvanlebras thanks! I only see the level specification, not the subtopic modification, did you miss committing that?

yvanlebras:
Hi Saskia, you're right, THANK YOU ! The subtopic modification is for another tuto I am looking at. So here it is ok! HAve a good end of week !

shiltemann (approved):

PR #6806, opened 2026-04-29
Updating DH Tutorial "Comparing Chinese Newspaper articles" for GTA

Sch-Da:
Hi,
I updated the WF and tutorial for GTA, included a new answer key history and incorporated best practices. It should be nearly good to go.

Two things I noticed:
- This tutorial currently does not have any Workflow tests. If those are required, I would need help, please
- I do not know how the updated WF: https://usegalaxy.eu/published/workflow?id=c917db7d7ba7ea6f can appear here: https://training.galaxyproject.org/training-material/topics/digital-humanities/tutorials/text_mining_chinese/workflows/
Help for this would be appreciated!
Thanks in advance


shiltemann:
@Sch-Da I can help with the workflow tests, but we dont need to wait for them before merging this.

All workflows that you add to the GTN will end up on that workflow page you linked to, so if that public workflow you link to is the same as the one you updated here then all is good :)

Sch-Da:
@shiltemann - that would be great, thank you!

good news, yes, this the update is on the same WF used before - so we should be good :)

shiltemann (approved):

shiltemann:
thanks @Sch-Da!

Sch-Da:
Thanks so much @shiltemann!

PR #6807, opened 2026-04-29
Review community tutorial dedicated community lab

martincarrere:
- Fix file tree display with proper markdown tree
- Fix brand name casing in link (GitHub, not Github)
- Replace bullet lists with tables for files to update
- Rephrase "script was generated" -> "script is provided"
- Fix "done" -> "made" for PR/MR references
- Fix "Add the file in the sections" -> "Add the file to the sections"

PR #6809, opened 2026-04-29
Add Martin Carrere as contributor and create brgm organization

martincarrere:
- Add new contributor Martin carrere :
  -  Merged [https://github.com/galaxyproject/training-material/pull/6807]()
- Create BRGM organization


ORGANISATIONS.yaml
```diff
@@ -83,6 +83,15 @@ bmbf:
     url: https://www.bmbf.de/DE/Home/home_node.html
     gtn-halloffame: "no"
 
+brgm:
+    short_name: BRGM
+    name: Bureau de Recherches Géologiques et Minières
+    github: false
+    url: https://www.brgm.fr/en/
+    avatar: "/training-material/shared/images/logos/brgm.svg"
+    gtn-halloffame: "no"
```
shiltemann: This is only needed when the organisation is not associated with any GTN contributors, which basically only happens because this file is re-used by the Galaxy Hub for its contributors, so we can remove it in this case. Each organisation also get a hall of fame page listing all members and their contributions (nice for e.g. reporting).

```suggestion
```

shiltemann (approved):

shiltemann:
thanks @martincarrere!

PR #6827, opened 2026-05-05
removed duplicated entry in CONTRIBUTORS.yaml

SantaMcCloud:
I was made aware that i appear twice in `CONTRIBUTORS.yaml` so i removed the latest entrey in this case and i also add my ORCID .


paulzierep:
is the id used in any training ? then you need to update this as well I guess ?

SantaMcCloud:
> is the id used in any training ? then you need to update this as well I guess ?

Not in a tutorial but in a different event. I chnage it.

paulzierep (approved):

PR #6829, opened 2026-05-06
Datasets import via link, add gif 

TuturBaba:
Add step numbers and GIFs to `datasets_import_via_link` snippet

Steps are now numbered (1→5) and a GIF is added at the end to visually guide the user.
Two GIFs: one for the standard case, one for the collection case.


shiltemann:
ooh, I like your GIfs with the numbered hand pointers. Could you add yourself to the contributors of the FAQ?

faqs/galaxy/datasets_import_via_link.md
```diff
@@ -34,7 +34,7 @@ examples:
 ---
 
 * Copy the link location
-* Click {% icon galaxy-upload %} **Upload** at the top of the activity panel
+* 1- Click {% icon galaxy-upload %} **Upload** at the top of the activity panel
```
shiltemann: please remove the numbers. This FAQ can look different depending on parameters passed to it, and the numbering may not makes sense in all cases, e.g. when the collection parameter is set there are more steps, so "close window" would not be step 5 anymore etc

TuturBaba: Yes, I can remove the numbers. Also, regarding "Close window",  it's inside the {% else %} block, so it doesn't appear when collection is enabled. The numbering was consistent in both cases. But I understand it can look strange when optional parameters like reset_form or format are present, since those steps are not numbered.

shiltemann: yes, and there can be combinations of parameters set so it seems easier to just not have any numbers so we dont have to worry about conistency in all possible cases.

Good point about "Close window", that should always be the last step I guess right? Should we take it out of the else case?

TuturBaba: No, because there's an if for collection, and when you build a collection you leave the panel automatically.

shiltemann: ah gotcha, thanks :+1:

shiltemann (approved):

shiltemann:
thanks @TuturBaba :+1:

PR #6859, opened 2026-05-12
Updating data-library.yaml files in Dh tutorials

Sch-Da:
@PlushZ noted that 2 of the tutorials in the DH track do not feature the proper topic information in the data libraries.yaml file and are therefore not working with the existing data libraries bot.
If someone with more experience with this could please check that this is correct now, I would really appreciate it, thank you!

### DIFF
--- a/bin/update-data-library
+++ b/bin/update-data-library
@@ -58,7 +58,7 @@ def update_data_library(path, topic, tutorial, zenodo_record)
     official_extension = f['type']
 
     link = f['links']['self'].sub(%r{/content$}, '')
-    unofficial_extension = link.split('.')[-2..].join('.')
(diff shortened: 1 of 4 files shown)


shiltemann:
@Sch-Da thanks for drawing attention to this, I found a bug in the script that automatically creates the data-libraries files (normally you shouldnt have to worry about this file).

I fixed the bug and regenerated the data-library.yaml files, let's see if that works for you

shiltemann (approved):

Sch-Da:
Wonderful @shiltemann - thank you! I was already surprised why I wasn't aware of those tasks but this explains it. Thanks both of you for your help!

PR #6865, opened 2026-05-12
[Infra] Include events in frontmatter linting

shiltemann:
I just realized events weren't being linted.


hexylena:
There's some legacy code in validate-other, that should probably be removed. it looks like it was only checking one single event, somehow, previously. (visible in the lint frontmatter ci log, after the learning pathways.)

hexylena (approved):

PR #6873, opened 2026-05-13
reduce size of image to prevent page loading problems

### LINKED ISSUE #6855: Some images are not compressed
Hello,

I checked some microbiome tutorials, and in one tutorial an image was not displayed. After checking it, I saw that this is the only image that was not compressed by the bot. I’m not sure if this is also a problem in other topics, but it might be possible.

Here is the link to the image that was not compressed:
https://github.com/galaxyproject/training-material/blob/77fc7038822d682eb913de774add892d8825b140/topics/microbiome/tutorials/pathogen-detection-from-nanopore-foodborne-data/images/heatmap2.png

and this is how it looks like in the GTN page:

Image

I also checked whether the image might be corrupted, but when I downloaded it, there was no problem with it.

shiltemann:
Reducing the size of this image because it was failing to load due to the very high dimensions of it. The bot wasnt able to automatically reduce it, so doing it manually.

fixes https://github.com/galaxyproject/training-material/issues/6855


bgruening (approved):

bgruening:
Danke

PR #6879, opened 2026-05-15
Update flexy classification

nilchia:
changed variance threshold from 0.8 to 80

removed tabpfn from workflow. (this will be reverted soon when tabpfn issue is fixed)


anuprulez (approved):

anuprulez:
Thanks a lot @nilchia

PR #6894, opened 2026-05-17
Update outdated parts of genomics intro and add a sort step

wm75:
Last minute change before GTA:

- Updates screenshots to reflect changes in UCSC's table browser and in Galaxy's edit dataset view
- Updates to GENCODE V49 instead of V41
- Removes outdated summary of what the gencode website's saying
- Fixes remaining mention of cat1 under its former name
- Adds a final sort step to obtain a coordinate-sorted bed dataset

Sorry, was going to do this earlier this week, but found the time only now.


wm75:
The hint about going back in the browser history after returning from UCSC to Galaxy is addressing https://github.com/galaxyproject/galaxy/issues/22647. This part can be removed once that issue is fixed.

bgruening (approved):

PR #6901, opened 2026-05-18
Fix mags-building

paulzierep:
Fixed the training
I wonder why the linter did not catch it, maybe since the tag does not exist and the tag was not closed it was a double issue. Will try to add after GTA.

PR #6902, opened 2026-05-18
Fix kMetaShot assignment detail box

bebatut:
Was still broken after #6901, because of spelling error (`detail` instead of `details`)
Screenshot 2026-05-18 at 09 27 26


paulzierep:
mhh why does the linter not check this ?

shiltemann (approved):

shiltemann:
@paulzierep the broken box linting is imperfect, in theory it should detect this, but it doesnt always. I will have to look at it more closely again at some point. Maybe you can create an issue linking to this failure mode.

PR #6924, opened 2026-05-29
Add Eamonn Bell to CONTRIBUTORS.yaml

Sch-Da:
@eamonnbell-dur I want to add you as a contributor for your upcoming news item.
Could you check that this handle is the one you want to use? Otherwise, please correct and adjust.
We will put the Gtn Hall of Fame on once you contribute there as well :)

### DIFF
--- a/CONTRIBUTORS.yaml
+++ b/CONTRIBUTORS.yaml
@@ -907,7 +907,6 @@ diegodepanis:
 eamonnbell-dur:
     gtn-halloffame: "no"
     name: Eamonn Bell
-    github: true
     joined: 2026-05
 
 emilylescak:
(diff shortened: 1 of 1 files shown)


CONTRIBUTORS.yaml
```diff
@@ -904,6 +904,12 @@ diegodepanis:
     github: false
     joined: 2026-01
 
+eamonnbell-dur:
+    gtn-halloffame: "no"
+    name: Eamonn Bell
+    github: true
```
shiltemann: the default is true, so you can omit this line unless you want to indicate that the user id in this file does NOT correspond to a GH username

```suggestion
```

shiltemann (approved):

Sch-Da:
Thanks for you sharp eyes again, @shiltemann - I forgot that one, yes!
I will wait for Eamonn to confirm and merge after :)

eamonnbell-dur:
This perfect, good to merge @Sch-Da

PR #6925, opened 2026-05-29
Correct spelling of 'nilchia' to 'Nilchia'

nilchia:
It fixes the link in the gta page to my account in hall of fame.


shiltemann (approved):
Thanks 👍 

(weird that our linter didnt catch this, will look into that)

PR #6959, opened 2026-06-25
Add a FAQ to guide users to IWC guidelines

lldelisle:
Hi,
I added a FAQ to guide the users to the IWC guidelines. We decided not to write a proper tutorial about it as it seems that this way we won't have issues in case we change the guidelines.


shiltemann:
@lldelisle thanks :) do you want to add the faq also to some workflow tutorials orso?

shiltemann (approved):

lldelisle:
Thanks, for the moment I would say no because I do not know the workflow tutorial part enough.

PR #6960, opened 2026-06-25
Adding tags

MarisaJL:
Adding tags to single-cell tutorials so we can use them in the codex


topics/single-cell/tutorials/EBI-retrieval/tutorial.md
```diff
@@ -71,6 +71,7 @@ recordings:
 tags:
   - data import
   - data management
+  - sc_format
```
pavanvidem: maybe sc_formats?

pavanvidem (commented):
please also change the tag from "single cell" to "single-cell"

pavanvidem (approved):

PR #6971, opened 2026-07-02
Update main Galaxy Project citation to latest 2026 NAR paper

B0r1sD:
After a request from one of our users to cite Galaxy, I noticed the main citation for the Galaxy Project has not yeet been updated to the one from this year (2026). 

This seems to be the only relevant place in the GTN where it should be updated, however on Galaxy level there are other places where it also still mentions to cite the 2024 edition where I will create a separate issue for so nothing is missed. 

Let me know if this was put on hold deliberately, but I could not find anything back by searching the repo.

### DIFF
--- a/topics/introduction/tutorials/galaxy-cite/tutorial.md
+++ b/topics/introduction/tutorials/galaxy-cite/tutorial.md
@@ -53,8 +53,8 @@ Citing Galaxy comes in two main actions:
 ## Citing Galaxy Project
 Please cite Galaxy Project’s primary publication: 
 
-- ***The Galaxy platform for accessible, reproducible, and collaborative data analyses: 2024 update***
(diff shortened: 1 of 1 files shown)


bgruening (approved):

B0r1sD:
I believe the GH action linting error can be ignored in this case? The linting notices a DOI and therefor mandates the presence of a `citation.bib` file, but not sure if that applies in this case (as in this case, it's not a GTN tutorial citing other material because it used material from it), as outlined in https://training.galaxyproject.org/training-material/gtn_rdoc/Gtn/Linter.html#method-c-check_dois. 
But it might be a good idea, what do maintainers think?

bgruening:
thanks @B0r1sD

PR #6980, opened 2026-07-06
Update 1: RNA-Seq Reads to Counts

tflowers15:
I have updated the [1: RNA-Seq reads to counts](https://training.galaxyproject.org/training-material/topics/transcriptomics/tutorials/rna-seq-reads-to-counts/tutorial.html) tutorial to use the latest version of `cutadapt`. The tutorial previously used `cutadapt` version `3.7+galaxy0`. I have updated the version to `5.2+galaxy2`, which reorganised the read filtering and trimming options (e.g. for setting minimum quality and length) and I have updated the tool setup to match the sections/parameter names used in version `5.2+galaxy2`.


bgruening (approved):
Great

PR #6992, opened 2026-07-12
docs: fix news event date format

### LINKED ISSUE #3226: Fix date presentation in news/events
image

sapunyangkut:
## Summary

- fix the misspelled Liquid date-format variable used for event start dates
- use the same defined `date_format` for start, end, and fallback dates

Closes #3226.

## Validation

- confirmed `date_format` is defined, `data_format` is no longer referenced, and all three date branches use `date_format`
- `git diff --check`

A full Jekyll render was not run locally because Ruby and Bundler are unavailable in the current environment.


bgruening (approved):
Thanks @sapunyangkut 

Try running GTN locally using the Docker container

Issue #6180, opened 2025-06-10
Broken tagging method with transcriptomics/tutorials/ref-based


jennaj:
**Tutorial steps:** https://training.galaxyproject.org/topics/transcriptomics/tutorials/ref-based/tutorial.html#analysis-of-the-differential-gene-expression **but specifically here** https://training.galaxyproject.org/topics/transcriptomics/tutorials/ref-based/tutorial.html#hands-on-add-tags-to-your-collection-for-each-of-these-factors

**Workflow:** https://training.galaxyproject.org/training-material/topics/transcriptomics/tutorials/ref-based/workflows/deg-analysis.html

**Issue:** The regular expression to parse the featurecounts file names into group tags in the list collection is not specific enough. The workflow breaks at DESeq2 early on and the hands-on doesn't have the correct instructions as a reference/check.

Example run with the error: https://usegalaxy.org/workflows/invocations/b51be7053d255f2e

Split `https://zenodo.org/record/6457007/files/GSM461176_untreat_single_featureCounts.counts`

Current (breaks workflow)
“Find pattern”: `(.*)_(.*)_(.*)`
“Replace with”: `\1_\2_\3\tgroup:\2\tgroup:\3`

Change to
“Find pattern”: `(.*)_(.*)_(.*)_(.*)`
“Replace with”:` \1_\2_\3_\4\tgroup:\2\tgroup:\3`

Leaving the workflow with the "error" as a teaching tool and only correcting the hands-on seems Ok but the correct instructions should be somewhere, yes? I'm guessing the zenodo file names had the extra _FeatureCounts added last year in the update but the instructions weren't adjusted to match.

Thanks! If this came up already, please close.

pavanvidem:
Can this be closed?

PR #6996, opened 2026-07-13
Fix tag parsing for optional filename suffixes

sapunyangkut:
Make the tag-based hands-on instructions accept both the three-component collection identifiers they ask learners to create and the four-component filenames reported in #6180.

The English regex added in #6181 only matches identifiers that retain the `_featureCounts.counts` suffix, while the preceding instruction uses names such as `GSM461176_untreat_single`. The anchored expression now treats any suffix as optional, preserves it when present, and consistently tags treatment and sequencing from capture groups 2 and 3. The same technical literals are synchronized across the German, Spanish, and Italian translations.

The workflow is intentionally unchanged: its tests use three-component collection element identifiers without the `_featureCounts.counts` suffix.

Refs #6180. Follow-up to #6181.

## Testing

- `git diff --check`
- `python bin/lint-diffs.py` for each changed tutorial
- GNU sed 4.9 behavior checks for both `GSM461176_untreat_single` and `GSM461176_untreat_single_featureCounts.counts`
- Regex checks for all seven collection identifiers in the workflow test data

## Automation disclosure

This focused documentation fix was prepared and checked with OpenAI Codex automation. The final expression was reviewed against the official Galaxy tool wrappers and validated with GNU sed 4.9.

### DIFF
--- a/topics/transcriptomics/tutorials/ref-based/tutorial.md
+++ b/topics/transcriptomics/tutorials/ref-based/tutorial.md
@@ -1621,8 +1621,8 @@ DESeq2 requires to provide for each factor, counts of samples in each category. 
 >      - {% icon param-file %} *"File to process"*: output of **Extract element identifiers** {% icon tool %}
 >      - In *"Replacement"*:
 >         - In *"1: Replacement"*
->            - *"Find pattern"*: `(.*)_(.*)_(.*)_(.*)`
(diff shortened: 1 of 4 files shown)


shiltemann:
@jennaj would you be able to test this change?

shiltemann:
@sapunyangkut thanks for looking at this. If your regex accepts both the 3 and 4 component file names, why do you say we shouldn't update it in the workflow?

PR #6999, opened 2026-07-15
Delete events/2026-10-12-Advanced-Galaxy-Training.md

scorreard:
Unfortunately, we have to cancel this training due to the low number of registrations.


shiltemann:
hi @scorreard, sorry to hear about the event cancellation. Instead of deleting the file, could you just add "draft: true"? ..then it wont be shown in the upcoming events anymore, but also doesn't lead to a dead link and will keep our URL persistence checker happy

shiltemann (approved):

scorreard:
> hi @scorreard, sorry to hear about the event cancellation. Instead of deleting the file, could you just add "draft: true"? ..then it wont be shown in the upcoming events anymore, but also doesn't lead to a dead link and will keep our URL persistence checker happy

Thanks for doing the change, I was on summer break :)

PR #7009, opened 2026-07-22
Add SamestrGal tutorial draft

xens25:
Adds a new GTN tutorial on detecting shared microbial strains with SameStr by using an FMT-treated rCDI dataset. PR includes tutorial.md, the workflow and tests


SaimMomin12:
> **Note:** This review was drafted by Claude (Claude Code) and reviewed by @SaimMomin12 before posting.


@xens25 Nice work on this - the pedagogical structure is genuinely good. The key-terms box, the per-tool parameter-explanation comment boxes, the "Why these parameter values?" details box, and the warning about the reduced dataset all show real care, and the question/solution boxes target the right concepts. Comments below cover `tutorial.md` only.

## Blockers

**1. All four referenced images are missing from the PR.** None of the 6 added files is an image, and all four paths 404 on the branch:

- L64 `../../images/samestr-workflow.png`
- L523 `../../images/samestr-taxon-counts.png`
- L530 `../../images/samestr-cooccurrences.png`
- L550 `../../images/samestr-strain-events.png`

This is why `build-site` and `lint` are failing. The path convention (`../../images/` → `topics/microbiome/images/`) is correct — the files just need committing. It also makes the entire results section and two of the three question boxes unanswerable, since the questions quote values that only exist in the screenshots.

**2. The PR description is still `TODO: Describe your changes here..`** — please fill it in, and mark the PR as a draft if it still is one.

## Front matter (L1–L27)

- **Missing `subtopic: metagenomics`.** Without it the tutorial won't be filed under a subtopic on the Microbiome topic page. See `topics/microbiome/tutorials/beer-data-analysis/tutorial.md` for a reference.
- **Missing `tags:`** — at minimum `metagenomics`, `microbiome`, and `microgalaxy` (the microGalaxy community filters on that tag; the `.ga` file already carries these tags, so the tutorial should match).
- **Missing `edam_ontology:`** — e.g. `topic_3174` (Metagenomics), `topic_3697` (Microbial ecology), `topic_0637` (Taxonomy).
- **Missing `requirements:`.** For `level: Intermediate`, please link prerequisites (Galaxy 101 / collections, plus an intro metagenomics tutorial) so learners aren't hitting Apply Rules cold.
- **Consider `answer_histories:`** or at least a link to a finished public history. Given the warning at L168–L171 that changing parameters can silently produce no output, a reference history is high value here.
- `time_estimation: 2H` looks optimistic — the tutorial has learners run the pipeline twice (whole workflow, then tool-by-tool), including two MetaPhlAn passes. Worth re-timing, or restructuring (see below).

## Naming

The tutorial uses **SamestrGal**, **SameStr**, and **Samestr** interchangeably, sometimes in the same sentence — e.g. the objectives at L12–L14 ("Run the SamestrGal workflow… Explain the role of each tool in the SameStr workflow"), and L62 / L66. Please settle on one convention: **SameStr** for the tool/suite, and either drop "SamestrGal" entirely or define it once as the name of the Galaxy workflow. I'd lean toward titling around the analysis rather than a coined name — e.g. "Detection of shared microbial strains with SameStr" — and renaming the directory to `samestr-strain-detection`.

## Structure

**The tutorial runs the same analysis twice.** Section 2 (L114) runs the complete workflow, then section 3 (L175) repeats every step manually. The parameter values are duplicated between the two sections, which is a maintenance trap — they will drift. Two options:

- Make the step-by-step section the main path and move the one-click workflow run to a short optional section at the end ("now that you understand each step, here's the whole thing as a workflow"), or
- keep the workflow-first order but drop the duplicated parameter table from section 2 and point at the per-tool sections instead.

Related: L177 tells the reader to "start by uploading the data and building a paired collection as described in **step 2 of Running SamestrGal as a Complete Workflow**." Collection building is a data-prep step, not a workflow step — move it into the `Get data` hands-on box at L74 so both paths share it and there's no forward reference.

## Correctness / accuracy

- **L34, L40: *Clostridium difficile* → *Clostridioides difficile*.** The currently accepted name is *Clostridioides*; the `cdifficile` bib entry is itself titled "Clostridioides difficile", so the body text contradicts its own reference. Mention the older name once in parentheses if useful.
- **L542: the solution contradicts itself.** Question 2 asks "Can you tell from this table whether any strain persisted in the patient? How would you tell?" The answer begins "**No.**" and then explains exactly how you *can* tell. The table does answer the question — the answer is "yes, and the finding is no persistence." Suggested rewrite: *"Yes. The Pre-FMT/Post-FMT row shows `shared_strain = 0`, so no clade was confirmed as the same strain between those samples — this data shows no evidence of persistence."*
- **L536: "Why is it less likely that Pre-FMT and Donor share *anything*?"** As written this is too strong — an unrelated donor and patient will routinely share common commensal *species*, which is exactly what the `shared_species` / `shared_clade` columns count. The solution only holds at strain level; suggest rewording to "share the same strain".
- **L347: the Apply Rules regex is fragile.** `^[^_]+_(.*)$` strips everything up to the *first* underscore. It works only because step 3 (L97) renamed the samples to `Pre-FMT` / `Donor` / `Post-FMT`. A learner who skips the rename (raw names are `28A_R1`, `28B_R2`, …) or brings their own data with underscores in sample names gets silently mangled identifiers and confusing downstream failures. Please add a warning box and make the rename explicitly non-optional.
- **L308: "Percent identity" = `0.9`** — the label says percent, the value is a fraction. One clarifying clause would help ("expressed as a fraction, so 0.9 = 90%"), and ideally the label gets fixed in the tool wrapper.
- **L181 overclaims:** "to ensure that no sequencing-quality errors are introduced" — trimming reduces the impact of quality errors, it doesn't guarantee none.
- **L130 / L194 `hg19`:** fine if that's what the KneadData reference-data table offers, but hg19 is dated. If a newer build is cached, prefer it; otherwise add a one-line note that this is the bundled KneadData database.
- **L552–L556:** please double-check the three event label strings (`shared_strain`, `other_strain`, `same_clade`) against the actual `SameStr Summarize` output values — they need to match exactly for readers to find them.

## Missing explanation

- **SGB is never defined.** The results section leans entirely on identifiers like `t__SGB4285`, `t__SGB15299`, `t__SGB6362` (L561, L572) but never explains that MetaPhlAn 4 reports species-level genome bins, what the `t__` prefix means, or why a clade has a bin ID rather than a species name. Please add SGB (and "marker gene") to the key-terms box at L38 — this is essential for interpreting the final table.
- **L45, the MVS definition is hard to parse:** "the fraction of shared positions with enough coverage between two samples that contain at least one matching nucleotide" — the relative clause dangles. Suggest: *"Of all positions covered in both samples, the fraction at which the two samples share at least one nucleotide variant."*
- **L338 / L342:** the Flatten → Apply Rules step is by far the most error-prone action in the tutorial, and it's the only one with no snippet and no screenshot. A Rule Builder screenshot would help a lot. A one-sentence aside that this is a workaround for the tool's flat-list input requirement would also tell readers it isn't a conceptual step.
- **L462, SameStr Stats:** the details box discusses `f_covered`, `f_mono` and `f_poly` but shows no example output, unlike the Summarize section which gets three figures. Either add a figure or turn it into a question that has the learner read their own values.
- **The results section is screenshot-only.** The learner has just generated these exact outputs, so the text should direct them to open *their* datasets, with the figures as reference. As written, questions quoting `895` / `2783` / `t__SGB4285` break if the screenshots or reference data ever change.

## Workflow reference

L120 hard-codes a usegalaxy.eu published-workflow ID (`8e13c87e751e88d8`). Since the PR ships `workflows/main_workflow.ga`, please point at the GTN-hosted workflow instead — it appears on the tutorial's Workflows tab, works across servers, and stays in sync with the repo. The `{% snippet faqs/galaxy/workflows_run_trs.md %}` pattern is the usual approach. Also worth giving the `.ga` file a meaningful name rather than `main_workflow.ga`.

## Markdown / GTN style

- **L32: `<!-- This is a comment. -->`** — leftover from the template, please delete.
- **L203: snippet indentation is wrong.** `> {% snippet faqs/galaxy/tools_select_collection.md %}` sits at one `>` level but needs to be indented under list item 1 (`>    `) like the other snippets. As-is it terminates the list, and step 2 at L210 will renumber as "1." The warning box right after it is at `>    >`, so indentation is inconsistent within this one box.
- **L213 / L215: wrong icon.** These are collections but use `{% icon param-file %}` — should be `{% icon param-collection %}`.
- **Outputs are named by internal API name rather than the label the user sees:** `paired_output` / `unmatched_paired` (L207, L213, L215) and `sam_output_file` (L261). Please use the visible output names so learners can find them in the history.
- **Inconsistent input references:** most steps use `(output of **Tool** {% icon tool %})`, but L477–L478 refers to inputs by hands-on-box title ("the collection produced by the *Remove Empty SNV Profile Datasets* step"). Suggest standardizing on the former.
- **L424 and L431:** two separate hands-on boxes for two runs of the same tool. These can be merged into one box with steps 1 and 2.
- **Heading case is mixed** — "SamestrGal Workflow Overview", "Get data", "Pre-processing of reads with KneadData", "Taxon Counts table", "Co-occurrence table". GTN prefers sentence case; pick one and apply throughout.
- **Trailing whitespace** on L181, L379, L472, L552.
- **The import URL list at L85–L90 is out of order:** `28C_R1, 28C_R2, 28A_R1, 28B_R2, 28A_R2, 28B_R1`. Sorting them (`28A_R1, 28A_R2, 28B_R1, …`) would let learners match them against the rename table at L97 — better still, rename the files on Zenodo to `Donor_R1` etc. and drop the rename step entirely.
- **Cross-file note:** L80–L82 points learners to `GTN - Material` → `{{ page.topic_name }}`, but `data-library.yaml` still has the template placeholders `name: New topic` / `description: Topic summary` (and empty `ext: ''`), so that path won't exist. Setting the topic to Microbiome and `ext: fastqsanger.gz` would fix it.

## Writing

Three paragraphs are single blocks of 470–1,035 characters and would read better as bulleted lists, especially since two of them are enumerations:

- **L62** (851 chars) walks the entire pipeline in prose immediately before the workflow *diagram*. A numbered list would mirror the figure.
- **L183** (1,035 chars) repeats "is another of the internal tools" three times. A list with one bullet each for Trimmomatic / TRF / FastQC / host removal would be much clearer.
- **L579** (conclusion, 624 chars) is one paragraph that mostly restates `key_points`.

Smaller ones:

- L375: "Why do we have **less** datasets than before?" → "fewer".
- L449: "SameStr Stats produces one of the final outputs of the workflow, since its results are not required by any subsequent tool" — the "since" reverses cause and effect. Try: *"SameStr Stats is a side branch of the workflow: its output is a final result and is not consumed by any later tool."*
- L108: the `#donor` / `#pre-fmt` / `#post-fmt` tags are added "so they are easy to identify later in the workflow" but are never referenced again — the renaming in step 3 already does that job. Either drop the step or say what the tags are actually for (propagation through collections).

xens25:
Hi @shiltemann , could I get a review on this tutorial? It's part of my thesis and should be ready for my submission. Appreciate any help!

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
+---
+layout: tutorial_hands_on
+title: Using SamestrGal to identify shared strains in FMT-treated rCDI samples
```
shiltemann: I wonder if the title could be adapted to make it clearer for learners not familiar with the tool and acronyms.

Perhaps something closer to how you formulated your learning questions/objectives, e.g. "Detecting shared microbial strains across metagenomic samples"?
(unless this method is very specific to this type of data, or if the tool name is really something so popular people will look for that specifically)

What do you think? Up to you but maybe something to consider.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+> In this tutorial, we will cover:
+>
+> 1. TOC
+> {:toc}
+>
+{: .agenda}
+
+# SamestrGal Workflow Overview
+
+SamestrGal uses metagenomic samples as input and runs a sequence of tools to detect shared strains between them. The reads are first trimmed and quality-filtered with KneadData, then aligned to species-specific marker genes with MetaPhlAn, which also produces a taxonomic profile for each sample. The resulting alignments are then processed by the SameStr tools, which convert them into a species-specific SNV profile for each clade detected in a sample. These SNV profiles are merged across samples for each clade. The Maximum Variant Profile Similarity (MVS) is then calculated between samples for each clade. Thresholds for the overlap and the MVS between samples are also set, and the results are reported in a table across all sample comparisons. Based on these thresholds, the table indicates whether each clade is classified as a shared strain.
```
shiltemann: Is SamestrGal just SameStr for Galaxy? Maybe put that in parentheses because it confused me for a second that this acronym was not introduced yet.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+
+SameStr is a bioinformatic tool that detects shared microbial strains across metagenomic samples by comparing the genetic variation observed at species-specific marker genes. In FMT studies, this allows tracking of strain persistence and engraftment between donor, pre-FMT, and post-FMT samples {% cite samestr_podlesny %}.
+
+> <details-title>Key terms used in this tutorial</details-title>
+>
+> - **rCDI**: recurrent *Clostridium difficile* infection, caused by *C. difficile* spores that survive antibiotic treatment and later reactivate.
```
shiltemann: Just so you are aware, you can also define these abbreviations in metadata, then use braces when you use them to link to their definition (if that is useful for you).

see here for more info https://galaxyproject.github.io/training-material/topics/contributing/tutorials/create-new-tutorial-content/tutorial.html#abbreviations

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
shiltemann: Tip: when writing tool names, you can use e.g. `{% tool KneadData %}` for a bit nicer formatting.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>    https://zenodo.org/records/20745835/files/28B_R2.fastq.gz
+>    https://zenodo.org/records/20745835/files/28A_R2.fastq.gz
+>    https://zenodo.org/records/20745835/files/28B_R1.fastq.gz
+>    ```
+>
+>    {% snippet faqs/galaxy/datasets_import_via_link.md %}
+>
+>    {% snippet faqs/galaxy/datasets_import_from_data_library.md %}
+>
+> 3. Rename the datasets to show the role of each sample in FMT:
```
shiltemann, line 127: If we rename the files in Zenodo, people wont have to do this somewhat tedious renaming step in Galaxy.

Or skip the renaming and rely on the tags to know which is Pre/Donor/Post, which would also illustrate the usefulness of the tags.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>
+{: .hands_on}
+
+# Running SamestrGal as a Complete Workflow
+
+This section shows how to run the entire SamestrGal analysis in a single step, using the published workflow. The following section then goes through each tool individually to explain what it does.
+
+> <hands-on-title> Run SamestrGal </hands-on-title>
+>
+> 1. Import the [SamestrGal workflow](https://usegalaxy.eu/published/workflow?id=8e13c87e751e88d8) into your account
```
shiltemann: by linking to a workflow on Galaxy EU it becomes harder to import for people on other Galaxies. 

Better to link to the workflow here in the GTN, we have a snippet to do this nicely:

```
{% snippet faqs/galaxy/workflows_run_trs.md path="topics/microbiome/tutorials/samestr-strain-detection/workflows/samestrain-strain-detection.ga" title="Samestr Strain Detection Workflow" %}
```

shiltemann:
@xens25 Thanks a lot for your contribution, I am happy to have a look at this :)

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>
+>    {% snippet faqs/galaxy/workflows_import.md %}
+>
+> 2. Build a paired collection from the three samples uploaded in the previous step. Each sample's forward and reverse reads will be grouped together
```
shiltemann: I would put this data collection creation step in the previous hands on box, where we also do the rest of the data preparation steps.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>
+>    {% snippet faqs/galaxy/datasets_add_tag.md %}
+>
+{: .hands_on}
+
+# Running SamestrGal as a Complete Workflow
+
+This section shows how to run the entire SamestrGal analysis in a single step, using the published workflow. The following section then goes through each tool individually to explain what it does.
+
+> <hands-on-title> Run SamestrGal </hands-on-title>
```
shiltemann: The way it is presented now, I think most people will run the full workflow, and then repeat everything again step-by-step. Resulting in double jobs, which can be a problem during e.g. training courses.

I would suggest you put the workflow section at the end. Once people have run through the entire workflow step by step, you can explain how they can do it faster next time or on their own data by using the workflow.

You can always already link to this workflow section here by saying something like "Want to just run the workflow? jump here!"

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+> 1. {% tool [KneadData](toolshed.g2.bx.psu.edu/repos/iuc/kneaddata/kneaddata/0.12.1+galaxy1) %} with the following parameters:
+>    - *"Read type"*: `Paired reads`
+>        - *"Save all unmatched reads"*: `Yes`
+>        - {% icon param-collection %} *"Paired reads collection"*: the paired collection created earlier
+>    - In *"Alignment Tool"*:
+>        - *"Select alignment tool for host removal"*: `Bowtie2 (default)`
+>            - *"Select reference genome database"*: `hg19`
+>    - In *"Quality Control with FastQC"*:
+>        - *"FastQC quality reports"*: `Do not generate FastQC reports`
+>    - In *"Trimmomatic"*:
```
shiltemann, line 176: In general, you can leave out parameters that are left to their default settings (unless you explicitly want to draw attention to them for didactic reasons). Please check all tools for settings that can be omitted.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>        - *"Run Trimmomatic quality/adapter trimming"*: `Enable Trimmomatic trimming`
+>            - *"Available sequencers"*: `NexteraPE`
+>    - In *"Tandem Repeat Finder (TRF)"*:
+>        - *"Specify if you want to include the TRF step in the workflow."*: `Skip TRF Step`
+>
+> {% snippet faqs/galaxy/tools_select_collection.md %}
```
shiltemann: the indentation here needs to match the surrounding, or it will break the numbering.

But in this case I think this can be removed. The way the tool works this parameter already expects a collection so there are no buttons to switch, and this information is not relevant in this step.

But definitely add it at the first time it is needed, which I believe is the samtools view step.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>    >
+>    > Enabling TRF adds two extra output collections containing only the reads affected by repeat removal. These can be useful for checking how many reads remain after TRF, but are not meant to be carried to the next step. `paired_output` and `unmatched_paired` are still the correct outputs to use in the next step.
+>    {: .warning}
+>
+> 2. {% tool [Merge collections](__MERGE_COLLECTION__) %} with the following parameters:
+>    - In *"Input Collections"*:
+>        - {% icon param-repeat %} *"Insert Input Collections"*
+>            - {% icon param-file %} *"Input Collection"*: `paired_output` (output of **KneadData** {% icon tool %})
+>        - {% icon param-repeat %} *"Insert Input Collections"*
+>            - {% icon param-file %} *"Input Collection"*: `unmatched_paired` (output of **KneadData** {% icon tool %})
```
shiltemann: In this step I got outputs named

- Unmatched trimmed reads
- Trimmed paired reads
- Unmatched reads
- Paired output reads

But it is unclear which output you mean with `unmatched_paired`. Maybe update the names in this part to match exactly with Galaxy output names. 

And I assume the trimmed outputs should be used here?

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+
+
+SameStr Convert takes the aligned reads and the taxonomic profile from MetaPhlAn and produces one SNV profile per detected clade, describing the nucleotides observed at each covered position of that clade's marker genes.
+
+
+> <hands-on-title> Generate Per-Clade SNV Profiles </hands-on-title>
+>
+> 1. {% tool [SameStr Convert](toolshed.g2.bx.psu.edu/repos/iuc/samestr/samestr_convert/1.2025.111+galaxy0) %} with the following parameters:
+>    - {% icon param-collection %} *"Aligned reads"*: the BAM collection produced by **Samtools view** {% icon tool %}
+>    - {% icon param-collection %} *"Taxonomic profile"*: the taxonomic profile collection produced by **MetaPhlAn** {% icon tool %}
```
shiltemann: I would be as explicit as possible in hands on boxes and match the names of the outputs in Galaxy exactly

```suggestion
>    - {% icon param-collection %} *"Taxonomic profile"*: `Predicted taxon relative abundances` collection produced by **MetaPhlAn** {% icon tool %}
```

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>    > - *"Minimum alignment length"* is the shortest amount of aligned base pairs a read can have and still be kept. Shorter alignments are considered too unreliable to use.
+>    > - *"Minimum base quality"* is the lowest sequencing quality score a nucleotide call can have and still be counted. A score of 20 corresponds to a 99% probability that the base was called correctly.
+>    > - *"Minimum alignment quality"* is the lowest mapping quality score a read can have and still be kept. Setting it to 0 means no read is excluded on this basis alone.
+>    > - *"Minimum vertical coverage"* is the minimum number of reads that must cover a position for that position to be included in the SNV profile.
+>    >
+>    {: .comment}
+>
+{: .hands_on}
+
+> <details-title> How is the quality score calculated? </details-title>
```
shiltemann: this box seems a bit out of place here?

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>    >
+>    > - *"Percent identity"* is the minimum fraction of bases in a read that must match the reference marker sequence. Reads with more mismatches than this are discarded.
+>    > - *"Minimum alignment length"* is the shortest amount of aligned base pairs a read can have and still be kept. Shorter alignments are considered too unreliable to use.
+>    > - *"Minimum base quality"* is the lowest sequencing quality score a nucleotide call can have and still be counted. A score of 20 corresponds to a 99% probability that the base was called correctly.
+>    > - *"Minimum alignment quality"* is the lowest mapping quality score a read can have and still be kept. Setting it to 0 means no read is excluded on this basis alone.
+>    > - *"Minimum vertical coverage"* is the minimum number of reads that must cover a position for that position to be included in the SNV profile.
+>    >
+>    {: .comment}
+>
+{: .hands_on}
```
shiltemann, line 297: It might be worth briefly discussion the outputs after this step. Even if people cant view the binary outputs, just briefly mention what the SNV profile outputs are/contain.

And maybe worth having them look at the alignment stats output, do a bit of a QC check there?

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+> <hands-on-title> Fix the Collection Identifiers </hands-on-title>
+>
+> 1. {% tool [Apply rules](__APPLY_RULES__) %} with the following parameters:
+>    - {% icon param-collection %} *"Input Collection"*: the flattened SNV profile collection produced by **Flatten collection** {% icon tool %}
+>    - In the Rule Builder:
+>        - *"Add Column"* > *"Using a Regular Expression"*
+>        - Select *"Create column from expression replacement"*
+>            - *"Regular Expression"*: `^[^_]+_(.*)$`
+>            - *"Replacement Expression"*: `\1`
+>        - Set *"List Identifier(s)"* to column B
```
shiltemann: The Apply rules tool is a complicated tool and not so widely known, so be as explicit as possible.

Setting list Identifier to column B involves more steps, e.g.
- clicking on "rules"
- click on "Add/Modify column definitions"
- ...

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,596 @@
...
+>
+{: .hands_on}
+
+
+> <comment-title> Expected output </comment-title>
```
shiltemann, line 361: very nice that you added this :)

shiltemann (commented):
Thanks a lot @xens25! Very nice tutorial and workflow :)

I am testing the tutorial now, I have not made it to the end yet, so may add a few more comments later, but this way you can already get started.

xens25:
@shiltemann Thank you for all your feedback. I realized I had accidentally uploaded an older version of the tutorial, which is why some of your comments referenced content that was already outdated. I've now pushed the corrected version with your feedback applied. Let me know if there's anything else to change

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+>
+>    > <warning-title> If TRF is enabled </warning-title>
+>    >
+>    > Enabling TRF adds two extra output collections containing only the reads affected by repeat removal. These can be useful for checking how many reads remain after TRF, but are not meant to be carried to the next step. `Paired output reads` and `Unmatched reads` are still the correct outputs to use in the next step.
+>    {: .warning}
+>
+> 2. {% tool [Merge collections](__MERGE_COLLECTION__) %} with the following parameters:
+>    - In *"Input Collections"*:
+>        - {% icon param-repeat %} *"Insert Input Collections"*
+>            - {% icon param-collection %} *"Input Collection"*: `Paired output reads` (output of **KneadData** {% icon tool %})
```
shiltemann, line 190: Why aren't we using the "Trimmed paired reads" and "Unmatched trimmed reads" datasets here? Why did we perform the trimming if not to use them? Maybe add some explanation about that because it feels counterintuitive.

(these files are not the ones you refer to in your warning box above right? because we skipped TRF and still get these trimmed outputs)

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+>        - {% icon param-repeat %} *"Insert Input Collections"*
+>            - {% icon param-collection %} *"Input Collection"*: `Unmatched reads` (output of **KneadData** {% icon tool %})
+>    - In *"Advanced Options"*:
+>        - *"How should conflicts (or potential conflicts) be handled?"*: `Keep first instance`
+>
+>    > <comment-title> Why merge these two collections? </comment-title>
+>    >
+>    > KneadData outputs two separate collections of reads: `Paired output reads`, the quality-filtered and host-decontaminated reads that kept their mate pair, and `Unmatched reads`, the orphaned reads. Orphaned reads are those that lost their paired read during quality filtering but still contain information themselves. Merging both collections ensures that all of the surviving microbial sequence data is used in the strain-level analysis.
```
shiltemann: I get 4 output collections, "Paired output reads", "Trimmed paired reads", and "Unmatched reads", "Unmatched trimmed reads"

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+>        4. Fill in *"Replacement Expression"*: `\1`
+>        5. Click *"Apply"*
+>        6. Go to *"Add/Modify Column Definitions"*
+>        7. Click *"List Identifier(s)"* and select `B`
+>
+{: .hands_on}
+
+> <warning-title> Incorrect Identifiers </warning-title>
+>
+> The regex `^[^_]+_(.*)$` strips everything before the first underscore, which only works because samples were renamed to `Pre-FMT`, `Donor`, and `Post-FMT` earlier. Make sure the rename step was completed before running this tool.
```
shiltemann: wouldnt it still have worked without the rename since the original files also didnt have an underscore (28A, 28B, 28C)?

(please do keep this box telling people how they should or shouldn't name their inputs in general e.g. for their own data, I guess no underscores is the main point here?)

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+>    >
+>    > SameStr Convert organizes its output by sample, with each sample containing one npz file per detected clade. SameStr Merge instead needs a single flat list of npz files, since it groups them by clade internally based on their filenames. Flatten collection removes the per-sample nesting so all npz files can be passed to SameStr Merge together.
+>    {: .comment}
+>
+{: .hands_on}
+
+
+The {% tool [Flatten collection](__FLATTEN__) %} tool joins each identifier with an underscore, which adds the sample name as an extra prefix, producing `sample_clade.sample` (e.g. `Pre-FMT_t__SGB14809.Pre-FMT`). This breaks the `clade.sample` format that SameStr Merge expects, so the following step removes this prefix and restores the original identifiers. The step is added to the workflow as a workaround for Apply Rules' flat-list input requirement.
+
+![Rule Builder configured to restore each clade's original identifier](../../images/samestr-rule-builder.gif "Apply Rules Rule Builder with the regular expression that strips the sample-name prefix added by Flatten collection.")
```
shiltemann: I would move this gif to the hands-on box below

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+> 1. {% tool [Apply rules](__APPLY_RULES__) %} with the following parameters:
+>    - {% icon param-collection %} *"Input Collection"*: the flattened SNV profile collection produced by **Flatten collection** {% icon tool %}
+>    - In the Rule Builder:
+>        1. Click *"Add Column"* > *"Using a Regular Expression"*
+>        2. In *"From Column"*, select *"Create column from expression replacement"*
+>        3. Fill in *"Regular Expression"*: `^[^_]+_(.*)$`
+>        4. Fill in *"Replacement Expression"*: `\1`
+>        5. Click *"Apply"*
+>        6. Go to *"Add/Modify Column Definitions"*
+>        7. Click *"List Identifier(s)"* and select `B`
```
shiltemann: Just some tweaks to match the interface more precisely:

```suggestion
>    - Under *"Rules"* click "*{% icon galaxy-wf-edit %} Edit*" to open the **Rule Builder**
>        1. Click *"{% icon plus %} Column"* -> *"Using a Regular Expression"*
>           - In *"From Column"*, choose `A` from the dropdown
>           - Select *"Create column from expression replacement"*
>           - Fill in *"Regular Expression"*: `^[^_]+_(.*)$`
>           - Fill in *"Replacement Expression"*: `\1`
>           - Click *"Apply"*
>        2. Click *"{% icon plus%} Rules"* -> *"Add/Modify Column Definitions"*
>           - Click *"List Identifier(s)"* and select `B`
>           - Click *"Apply"*
>        3. Click *"Save"*

```

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+### SameStr Merge
+
+{% tool [SameStr Merge](toolshed.g2.bx.psu.edu/repos/iuc/samestr/samestr_merge/1.2025.111+galaxy0) %} combines the per-sample SNV profiles for the same clade into a single multi-sample profile. For each clade, it produces one merged npz profile combining the data from every sample that contains it, along with a sample name file listing which samples are included.
+
+> <hands-on-title> Merge SNV Profiles Across Samples </hands-on-title>
+>
+> 1. {% tool [SameStr Merge](toolshed.g2.bx.psu.edu/repos/iuc/samestr/samestr_merge/1.2025.111+galaxy0) %} with the following parameters:
+>    - {% icon param-collection %} *"Input SNV profiles"*: the SNV profile collection produced by **Apply rules** {% icon tool %}
+>    - *"SameStr database"*: `mpa_vJan25_CHOCOPhlAnSGB_202503-11062025`
+>
```
shiltemann, line 376: this would be a good place to add the FAQ about switching to the collection input

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+
+{% tool [SameStr Convert](toolshed.g2.bx.psu.edu/repos/iuc/samestr/samestr_convert/1.2025.111+galaxy0) %} takes the aligned reads and the taxonomic profile from MetaPhlAn and produces one SNV profile per detected clade, describing the nucleotides observed at each covered position of that clade's marker genes.
+
+
+> <hands-on-title> Generate Per-Clade SNV Profiles </hands-on-title>
+>
+> 1. {% tool [SameStr Convert](toolshed.g2.bx.psu.edu/repos/iuc/samestr/samestr_convert/1.2025.111+galaxy0) %} with the following parameters:
+>    - {% icon param-collection %} *"Aligned reads"*: `filtered alignments` collection produced by **Samtools view** {% icon tool %}
+>    - {% icon param-collection %} *"Taxonomic profile"*: `Predicted taxon relative abundances` collection produced by **MetaPhlAn** {% icon tool %}
+>    - *"SameStr database"*: `mpa_vJan25_CHOCOPhlAnSGB_202503-11062025`
```
shiltemann: just to match exactly what is in the dropdown: 

```suggestion
>    - *"SameStr database"*: `MetaPhlAn clade-specific marker genes (mpa_vJan25_CHOCOPhlAnSGB_202503-11062025)`
```

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+>    - In the Rule Builder:
+>        1. Click *"Add Column"* > *"Using a Regular Expression"*
+>        2. In *"From Column"*, select *"Create column from expression replacement"*
+>        3. Fill in *"Regular Expression"*: `^[^_]+_(.*)$`
+>        4. Fill in *"Replacement Expression"*: `\1`
+>        5. Click *"Apply"*
+>        6. Go to *"Add/Modify Column Definitions"*
+>        7. Click *"List Identifier(s)"* and select `B`
+>
+{: .hands_on}
```
shiltemann, line 353: could be nice to add our faq on regular expressions to this box

`{% snippet faqs/galaxy/analysis_regular_expressions.md %}`

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+
+
+# Running SameStr as a complete workflow
+
+Now that you have run through each step individually, this section shows how to repeat the same analysis in a single step using the published workflow.
+
+> <hands-on-title> Run SameStr </hands-on-title>
+>
+> 1. Run the **SameStr** {% icon workflow %} workflow using the parameter values described in the previous section, tool by tool
+>    - {% icon param-collection %} *"Raw Reads"*: the paired collection built in the *Get data* step
```
shiltemann, line 600: When I opened this workflow in Galaxy, the default option for Metaphlan was not the Jan25 one, so maybe add those database parameters here as well to make sure they match

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+>    > - *"Select SameStr database"* must always match the profiler selected in *"Select MetaPhlAn Database"*. Mixing databases from different profilers will cause errors.
+>    > - *"Minimum overlap for comparison"* (5000) and *"Minimum similarity for shared strains"* (0.999) are the two thresholds that ultimately decide whether a clade is classified as a shared strain.
+>    >
+>    > The remaining alignment and filtering parameters are explained individually, tool by tool, in the previous section.
+>    {: .details}
+>
+>
+>    > <warning-title> Changing Parameters </warning-title>
+>    >
+>    > The tutorial dataset was reduced to the minimum size that still produces results with the parameter values listed above. Changing these parameters may therefore cause the workflow to produce no output. To experiment with different parameter values, download the complete samples from the ENA under accession PRJEB39023 and repeat the Data Upload step with the full dataset.
```
shiltemann: can you give a link to ENA here?

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
```diff
@@ -0,0 +1,619 @@
...
+>    >
+>    > - *"Minimum overlap for comparison"* (5000) requires at least this many shared covered positions between two samples before their similarity is even considered, so the comparison isn't based on too little data to be reliable.
+>    > - *"Minimum similarity for shared strains"* (0.999) is the MVS threshold a sample pair must have to be classified as sharing a strain.
+>    >
+>    {: .comment}
+>
+{: .hands_on}
+
+
+## Interpreting the results from SameStr Summarize
```
shiltemann, line 535: the images in this section have the 28A/B/C identifiers in them, shouldnt that be Pre-FMT/Donor/Post-FMT?

shiltemann (commented):
thanks again @xens25! I finished testing the tutorial, just a few more comments here, but then I think it is good to merge :)

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
xens25: I will add a better explanation about the KneadData's results

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
xens25: Yes, that's correct. I modified the warning box so that it focuses on the general idea of not including underscores.

topics/microbiome/tutorials/samestr-strain-detection/tutorial.md
xens25: In theory yes, I just didn't do the renaming step. I will add it as a comment box

shiltemann (approved):

shiltemann:
thanks a lot @xens25!

PR #7018, opened 2026-07-28
Category

Marie59:
Hi all, 

Update of the subtopics here and some modification of tags
@annefou I was wondering maybe for some of your climate trainings could we add the tag atmosphere ? 

Plus, with Anne we were wondering if we could rename the topic Climate into something like Earth data sciences (does it look good to you too Anne) ?

### DIFF
--- a/topics/climate/metadata.yaml
+++ b/topics/climate/metadata.yaml
@@ -2,7 +2,7 @@
 name: climate
 type: use
 topic_type: field
-title: Climate
+title: Earth Data science
 summary: "Learn to access, analyze, and visualize Earth data using Galaxy."
 requirements:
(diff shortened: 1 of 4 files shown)


shiltemann:
@Marie59 do you care if the URL still has `/topics/climate/` in it?

If you would like to change this also we will have to rename the folder and setup redirects. If you are happy with just the title of the topic then this is good as is

Marie59:
I think this is good enough like that ! Will see if someone ask for a change in the url, but for me I don't think it's necessary to shift everything

shiltemann (approved):

shiltemann:
ok great, then we can merge this :)

PR #7029, opened 2026-08-11
Add priority to introduction to Digital Humanities tutorial

Sch-Da:
Adding priority to enable the introduction to show first in the list of tutorials


shiltemann:
(just sneaking in a bugfix for priorities not working as expected when there are no subtopics, still something weird happening but works for now)

shiltemann (approved):

Sch-Da:
That worked, thanks so much @shiltemann!

PR #7032, opened 2026-08-13
Change topic-level prerequisite

shiltemann:
have the introduction topic as a whole as a prerequisite, not the specific genomics intro tutorial

thanks @Sch-Da for pointing this out


Sch-Da:
that was quick, thanks so much!

PR #7033, opened 2026-08-13
Removing work around for requirements

Sch-Da:
Removed requirements section from IIIF tutorial since this no longer applies due to https://github.com/galaxyproject/training-material/pull/7032 thanks to @shiltemann


shiltemann (approved):

shiltemann:
thanks!

PR #7050, opened 2026-08-25
Add 'new_fave_extract' to the plugin list

Sch-Da:
Adding new fave's short tool ID to enable using it in GTN: https://github.com/galaxyproject/training-material/pull/7049


Sch-Da:
as agreed on with @shiltemann this is not needed now. Once the tools are on IUC, we will update WF and tutorial but leave it in alpha state in the tutorial unitl then.

PR #7056, opened 2026-08-31
added dianas workshop at ibsi

teresa-m:
Diana will give a conference workshop, and I added the event page from the hub as an external event.
@dianichj: do you like it like this?


dianichj:
Thanks so very much, Teresa! It looks great!

shiltemann (approved):

shiltemann:
thanks @teresa-m!

PR #7069, opened 2026-09-03
link a SC tutorial from the proteomics topic to single cell topic

shiltemann:
As requested by @nomadscientist ages ago in https://github.com/galaxyproject/training-material/issues/5762 I've linked a Single Cell tutorial from the Proteomics topic into the Single Cell topic.

@pavanvidem @dianichj @MarisaJL I didn't know what the best place was to put this, so I put it in the multiomics subtopic, but please let me know if you think somewhere else is better


pavanvidem:
The subtopic strictly does not contain multiomics but non-scRNA tutorials. Until we have real multiomics tutorials, that single-cell proteomics tutorial should fit there :+1:

pavanvidem (approved):

pavanvidem:
thanks! :pray:

PR #7078, opened 2026-09-07
Update GRANTS.yaml with new project entries

Sch-Da:
Added new entries for STARDAST project with funding details and added ai4social cordis website link as requested by @bgruening 

@teresa-m could you please double check that this works for you - thank you :)

I took all info from here:
https://cordis.europa.eu/project/id/101268773

### DIFF
--- a/GRANTS.yaml
+++ b/GRANTS.yaml
@@ -40,7 +40,7 @@ ai4social:
     github: false
     joined: 2026-03
 #   avatar: "/training-material/topics/ai4life/images/AI4Life-logo_giraffe-solid.png" No avatar yet
-#   url: https://ai4life.eurobioimaging.eu/  No website yet
+    url: https://doi.org/10.3030/101292886
     funding_id: "101292886"
     funding_database: cordis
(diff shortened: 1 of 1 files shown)


GRANTS.yaml
```diff
@@ -473,6 +473,18 @@ oscars:
     github: false
     avatar: "/training-material/shared/images/logos/OSCARS-logo-EUflag.png"
 
+stardast:
+    name: STARDAST
+    github: false
+    joined: 2026-09
+#   avatar: "/training-material/topics/ai4life/images/AI4Life-logo_giraffe-solid.png" No avatar yet
```
shiltemann, line 480: if there is no logo yet, you could consider for now using the Horizon Europe image that is also on the URL you provide below? might be a bit nicer than the default (but totally up to you)

Image

shiltemann (approved):

GRANTS.yaml
Sch-Da: I would leave that for @teresa-m to decide.

PR #7082, opened 2026-09-08
SAM3 tutorial: remove duplicate parameters

bernt-matthias:
Remove a duplicate filter

When doing the tutorial I was wondering if we have tools for filtering by p-value? Do people really filter manually?


shiltemann (approved):

shiltemann:
thanks @bernt-matthias :+1:

PR #7084, opened 2026-09-08
Update organization names for clarity and consistency

Sch-Da:
I hereby suggest adding popular German shortcuts in front of organisation names to show those instead of the currently used less common English translations on the HUB.

For example:
deNBI
deKCD
and MWK are more common than their English translations and will more likely be searched under the German acronyms.


shiltemann (approved):

shiltemann:
good idea, thanks @Sch-Da :)

PR #7093, opened 2026-09-11
Fix the PMID regex in GTN:004

### LINKED ISSUE #7092: Galaxy training-material: GTN:004's PMID regex has a doubled slash, so the check has never fired
## Summary

`bin/lint.rb::check_pmids` matches:

```ruby
# https://www.ncbi.nlm.nih.gov/pubmed/24678044
find_matching_texts(contents,
                    %r{(\[[^\]]*\]\(https?://www.ncbi.nlm.nih.gov/pubmed//[0-9]*\))})
```

The pattern requires `pubmed//` — two slashes — while the documenting comment on the
line directly above shows the real URL shape with one. Real PubMed links never match,
so GTN:004 has never fired for a PMID link. Its companion `check_dois` is unaffected.

## Measurement

Across the **547** `tutorial.md` files reachable from the topic API
(689 discovered across all 35 topics; the rest are slides-only or renamed):

| | count |
|---|---|
| links the shipped regex matches | **0** |
| links `.../pubmed/<id>` — the shape the comment documents | 47 |
| links on the modern host `pubmed.ncbi.nlm.nih.gov/<id>/` | 3 |
| tutorials with at least one missed link | 23 |

Most affected: `proteomics/protein-id-oms` (11), `proteomics/metaproteomics` (5), `proteomics/protein-id-sg-ps` (4), `proteomics/labelfree-vs-labelled` (3), `transcriptomics/rna-seq-genes-to-pathways` (3).

## Proposed fix

`check_pmids.patch` drops the duplicated slash, requires at least one digit, allows
the trailing slash PubMed itself emits, and additionally matches the modern
`pubmed.ncbi.nlm.nih.gov` host the check predates.

`test_check_pmids.rb` (16 assertions, all passing on ruby 4.0.6) pins the behaviour:
the shipped pattern misses all five real-world link forms; the proposed one matches
all five; and it ignores DOI links (handled by `check_dois`), the identifier-less
PubMed search page, non-PubMed NCBI links, bare URLs without link text, and existing
`{% cite %}` shortcodes. One assertion checks that capture group 1 is the whole
markdown link, since `ReviewDogEmitter` slices on `selected.begin(0)`/`end(0)`.

TilakSharma07:
Fixes #7092.

The `check_pmids` pattern in `bin/lint.rb` requires `pubmed//` — a
doubled slash. Real PubMed links have one, as the comment directly above
the pattern shows (`pubmed/24678044`), and as `check_dois` a few lines
up does correctly.

The check has therefore never fired. Across 23 tutorials it saw 47 PMID
links and flagged none.

The patched pattern was tested against legacy host, modern host, plain
HTTP and trailing-slash variants, plus URLs it must not match: 7/7.

### DIFF
--- a/bin/lint.rb
+++ b/bin/lint.rb
@@ -276,8 +276,11 @@ module Gtn
     # Companion function to Gtn::Linter.check_dois
     def self.check_pmids(contents)
       # https://www.ncbi.nlm.nih.gov/pubmed/24678044
-      find_matching_texts(contents,
-                          %r{(\[[^\]]*\]\(https?://www.ncbi.nlm.nih.gov/pubmed//[0-9]*\))}).map do |idx, _text, selected|
+      # https://pubmed.ncbi.nlm.nih.gov/24678044/
+      find_matching_texts(
+        contents,
+        %r{(\[[^\]]*\]\(https?://(?:www\.ncbi\.nlm\.nih\.gov/pubmed|pubmed\.ncbi\.nlm\.nih\.gov)/[0-9]+/?\))}
+      ).map do |idx, _text, selected|
         ReviewDogEmitter.warning(
           path: @path,
           idx: idx,


hexylena (approved):
This looks OK. Thank you @TilakSharma07

hexylena:
`test_check_pmids.rb` is mentioned in #7092, that could be included in this PR as well.

PR #7098, opened 2026-09-15
Update learning path image

paulzierep:
Use the mags image
@shiltemann somehow serve-quick did not show the new icon, but I think the logic should work


shiltemann (approved):

shiltemann:
yeah, some of these things are cached and only update when you completely clean and rebuild, not when you serve with --incremental, but this looks good to me. Thanks!

PR #7108, opened 2026-09-21
update DH landing page to add further community resources

Sch-Da:
Added links to the Social Sciences and Humanities Special Interest Group and SSH Subdomain for more information.

@shiltemann I hope this is the right place to do this. Otherwise, please redirect me to the correct file - thanks so much!


shiltemann:
@Sch-Da that is the correct file, but if you would like to draw a bit more attention to it, you can can also do something like the single cell topic and define it in special CTA (call to action) metadata:

definition:
https://github.com/galaxyproject/training-material/blob/main/topics/single-cell/metadata.yaml#L14

How it looks:
image

Sch-Da:
Wow, that is next level - thanks for sharing!
I will try to adapt that for the DH community as well.

Sch-Da:
I hope I added all details now correctly :)

shiltemann (approved):

shiltemann:
yep, looks great, thanks!

Sch-Da:
thanks a lot for suggesting, checking and merging!

PR #7109, opened 2026-09-21
correct typos and propose an enhencement

yvanlebras:
Discover a typo


yvanlebras:
Trying to finish the tutorial, I was not able to search for a feature... maybe I made something wrong...
image

shiltemann (approved):

shiltemann:
thanks @yvanlebras!

@abretaud @hexylena do you know why Yvan couldn't search in Jbrowse?

abretaud:
Hi!
I think that's because you need to check the "Index this track" checkbox when adding the gff track, otherwise it doesn't get indexed and a search like you did returns nothing. Can you try?

(We should update these tutorials to use jbrowse2 instead)

yvanlebras:
THANK YOU @abretaud ! 

Unfortunately, this seems to not resolve my issue, but maybe it is "just" I am not using JBrowse as I must to do ?

Here some screenshot, in case it helps

I pushed to ON the "index" parameter for both tracks:

image

image

image

And on JBrowse to search I tried on the left panel to type what I am seraching on the "search" text parameter

image


 and also on "viesw/serach features"

image

abretaud:
hmm you probably want to use the search box at the top of the window:

image

PR #7113, opened 2026-09-23
Add Eclipse tutorial input data to Zenodo, and regenerate data library file

shiltemann:
@Sch-Da I added the input datasets for this tutorial to Zenodo, and let our system auto-generate the data library file to the new format.

I've left the links in the tutorial pointing to DASCH.swiss in case that is a data source you want to show off, but now there is a backup in Zenodo (and hopefully soon also in the Galaxy Data Library)


Sch-Da (approved):
Thanks a lot for incorporating the files in the data library - that is great.
Yes, the tutorial specifically wants to show off Dasch.swiss as a data source, but I agree that for usage on .EU an incorporation as a data library would be very helpful - thank you. Looks good from my side

PR #7115, opened 2026-09-23
Update Galaxy server compatibility checking

shiltemann:
We were checking every Galaxy that ever existed for tutorial compatibility, but many of these are no longer up or maintained, and this step was adding 6 minutes to our deploy time.

So switching to simply having a manual list of Galaxy servers to check. Anybody running a Galaxy server who would like to be listed under "Available on these Galaxies" is very welcome to add their server to the list.

Will simplify things a bit, speed them up a lot, and allow us to set more generous timeouts etc (e.g. Galaxy Europe had been timing out recently and therefore not appearing in the dropdown (thanks @Sch-Da for reporting))


hexylena (approved):

bgruening (approved):
Do we know why EU timed out? Do we have too many tools?

shiltemann:
@bgruening no I can't tell, and testing locally didn't run into it, may have just been a transient issue. I will put some retries into the script now that we only have a small number to check.

